PR #5, opened 2019-05-27
Use pub(crate), not include!(), to access private fields

FauxFaux:
This PR removes the surprising use of `include!()`, which is documented to be there in order to access private fields. This seems to be better served by marking the fields as `pub(crate)`?

I care about this mostly because the use of `include!()` breaks autocomplete in my IDE, and generally my mental model of what's going on with the code.

I have additionally deleted `Cargo.lock`. This is a library crate, and [committing `Cargo.lock` is discouraged for library crates](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html), as it does not actually get used in "production".


algesten:
Good idea! This library started before `pub(crate)` was a thing, and it's indeed much better this way. Thanks!

PR #8, opened 2019-09-08
flush during send_prelude

FauxFaux:
I'm seeing requests fail with `BadStatus`. A lot. I'm using the Agent API.

I believe it's only after periods of inactivity. I can reproduce it on multiple machines, but am yet to reproduce it with sufficient logging to be certain as to what is happening.

My best guess is that:

 * some time after the first request, but before the second, the server is closing the connection (very reasonable)
 * the second request comes along, through `send_prelude`, which is filling up the TLS buffer, but not actually trying to use the socket, so not seeing an (IO) error
 * `do_from_read` is doing the first actual IO operation on the connection
 * this is getting the `UnexpectedEOF` (which `read_next_line` calls `ConnectionAborted`)
 * ..which is reported as a `BadStatus`

The code already copes with the underlying socket dying... so long as `send_prelude` notices. I suspect that it is not. We can fix this by trying to force some IO inside `send_prelude`, by flushing here.

--

The server this happens most with is https://api.twitter.com, which seems happy for you to reuse a connection ten times, or to leave it data-idle (maybe there are tcp keepalives?) for 25 minutes sometimes, but sometimes will just kill it after five minutes. I'm unclear on how it is killing it, maybe it's confusing `rustls`?

I'm on modern amd64 ubuntu, stable/nightly rust from today, home/datacentre internet.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -292,8 +292,9 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> IoResult<()> {
     // finish
     write!(prelude, "\r\n")?;
 
-    // write all to the wire
     stream.write_all(&prelude[..])?;
+    // write all to the wire
+    stream.flush()?;
 
     Ok(())
 }


FauxFaux:
I waited like three hours, you utter sock wrangling piece of software.

With this patch:

```patch
diff --git a/src/response.rs b/src/response.rs
index c93d756..8f6f11c 100644
--- a/src/response.rs
+++ b/src/response.rs
@@ -412,7 +412,10 @@ impl Response {
     fn do_from_read(mut reader: impl Read) -> Result<Response, Error> {
         //
         // HTTP/1.1 200 OK\r\n
-        let status_line = read_next_line(&mut reader).map_err(|_| Error::BadStatus)?;
+        let status_line = read_next_line(&mut reader).map_err(|e| {
+            log::warn!("bad status: {:?}", e);
+            Error::BadStatus
+        })?;
 
         let (index, status) = parse_status_line(status_line.as_str())?;

```

..I just saw:
` WARN  ureq::response > bad status: Custom { kind: ConnectionAborted, error: "CloseNotify alert received" }`

Which is not a denial of my suggestion; the connection is being closed (at the TCP level) and we're not noticing until we try and read. Maybe this is a `rustls` bug?

FauxFaux:
I thought this was familiar. I've just found the time someone (*cough* it was me) fixed this in a different http client.

I think I'm slowly losing it. It's the socks. I swear I'm still sane, it's the socks.

https://github.com/jayjamesjay/http_req/pull/3/commits/6887f12b373cde9a6a7f040b17a1fe51fe8f06ea#diff-5bcc8caa34cc400be17a2bd1a01a41ebR72

FauxFaux:
This doesn't work. See #10.

PR #9, opened 2019-09-09
Map CloseNotify to EOF

FauxFaux:
rustls informs us about a clean EOF by causing an error. Handle this case.

The test does fail for me without the fix, but it's not guaranteed to be a good test long-term, maybe something else should be considered.


lolgesten:
I'm happy with this for now. We can always review how to identify the stream close if/when it changes.

lolgesten:
Thank you very much!!!

PR #22, opened 2019-10-20
Fix compilation errors with the cookies feature

orf:
Sorry! I thought it was a bit suspicious that the tests passed first time.

I've made the tests run without `default_features`, but they only pass if both `tls` and `json` are enabled. So I've made them always run.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -1,5 +1,5 @@
 use std::io::{Result as IoResult, Write};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, MutexGuard};
 
 use base64;
 #[cfg(feature = "cookie")]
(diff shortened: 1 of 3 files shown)


src/unit.rs
```diff
@@ -44,19 +44,11 @@ impl Unit {
...
         let query_string = combine_query(&url, &req.query, mix_queries);
 
-        let cookie_headers: Vec<_> = {
-            let state = req.agent.lock().unwrap();
-            match state.as_ref().map(|state| &state.jar) {
-                None => vec![],
-                Some(jar) => match_cookies(jar, &hostname, url.path(), is_secure),
-            }
-        };
+        let state = req.agent.lock().unwrap();
```
algesten: Instead of locking the mutex here, I would prefer if you pass the state into `extract_cookies()` and keep the lock local to there. It would alter semantics more than you have, but I don't think we need state to be locked until end of this function.

orf: Hey! Sorry for the lateness, but I've implemented this 👍

PR #25, opened 2020-01-05
Bump base64 to 0.11; drops `byteorder` dependency

Shnatsel:
This drops ~220 unsafe expressions from the dependency tree. 

Tests still pass. There is one doc test failing but it fails both before and after this change.

After this the scariest thing remaining is `smallvec` that's pulled in by `unicode-normalization`, but that only uses a small subset of smallvec API, so it should be possible to replace with 100% safe code in less than a day of work.


lolgesten:
Thanks! This is fixed in #27

PR #31, opened 2020-01-17
Remove last `unsafe` usage and forbid further usages

EkardNT:
Prompted by [Smoke-testing Rust HTTP clients](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6), I looked into whether the `unsafe` code directly present in ureq could be removed and I think it can. The only uses of `unsafe` in ureq today are present because ureq wants to "unwrap" a `ChunkDecoder<Stream>` back into a plain `Stream`, but there's no way of safely doing so. To remove this unsafe usage, I've submitted https://github.com/frewsxcv/rust-chunked-transfer/pull/3 to the chunked_transfer crate, and this PR which makes use of the new ChunkDecoder::unwrap method to remove the need for unsafe. I also added `#![forbid(unsafe_code)]` to prevent further unsafes from sneeking in.

This cannot be merged until the linked rust-chunked-transfer PR is merged and a new version published, as ureq will need to be updated to depend on the new version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ cookies = ["cookie"]
 
 [dependencies]
 base64 = "0.11"
-chunked_transfer = "1"
+chunked_transfer = { version = "1", path = "rust-chunked-transfer" }
 cookie = { version = "0.12", features = ["percent-encode"], optional = true}
 lazy_static = "1"
 qstring = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ cookies = ["cookie"]
 
 [dependencies]
 base64 = "0.11"
-chunked_transfer = { version = "1", path = "rust-chunked-transfer" }
+chunked_transfer = "1"
 cookie = { version = "0.12", features = ["percent-encode"], optional = true}
 lazy_static = "1"
 qstring = "0.7"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![forbid(unsafe_code)]
 #![warn(clippy::all)]
 //! ureq is a minimal request library.
 //!
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::io::{Read, Result as IoResult};
 
-use crate::stream::Stream;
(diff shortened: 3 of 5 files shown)


EkardNT:
Test is failing because the unwrap() method is not present, which is expected until the linked PR is merged.
```
error[E0599]: no method named `unwrap` found for type `chunked_transfer::decoder::Decoder<R>` in the current scope
##[error]  --> src/stream.rs:80:23
   |
80 |         chunk_decoder.unwrap().into()
   |                       ^^^^^^ method not found in `chunked_transfer::decoder::Decoder<R>`
```

lolgesten:
Thank you! This is so great and improves the code a ton!

EkardNT:
I thought more about this and realized that it is possible to achieve `unsafe` removal without depending on a new `into_inner` method in chunked-transfer by using `Rc<RefCell<Stream>>`. Code which does this is in my [remove-unsafe-rc](https://github.com/EkardNT/ureq/tree/remove-unsafe-rc) branch. 

However, while it works, its not as nice a solution as depending on `into_inner`. Objectively, it requires an additional heap allocation for the `Rc`. Subjectively, the code is a bit harder to understand and is brittle against future bugs due to the runtime checking done by `RefCell`. So I'd prefer to keep this `into_inner` approach if possible, and reserve the `Rc<RefCell<>>` only for the unlikely case that its difficult to upgrade the chunked-transfer dependency.

lolgesten:
@EkardNT interesting! i much prefer the `into_inner()` as well. The connection pooling function is another matter… I only use ureq without keep-alive or any such – so it's very untested territory and I know it has issues.

frewsxcv:
> Test is failing because the unwrap() method is not present, which is expected until the linked PR is merged.

just published a new version of `chunked-transfer` that has this https://github.com/frewsxcv/rust-chunked-transfer/pull/3

algesten:
Hm. Seems it conflicts with current master.

algesten:
@EkardNT there is some small amount of tidy up to do. If I check out your branch, I get a path dependency to `chunked_transfer` in Cargo.toml. I don't get why that difference isn't showing up in the github UI though 🤔 

The commit message also talks about `.unwrap`() which was renamed to `into_inner()`

algesten:
Now in master. Thanks!

Issue #30, opened 2020-01-15
Connection closed too early on some websites


Shnatsel:
Some websites transmit data really slowly, but ureq drops the connection almost immediately after establishing it, without actually downloading the content, and without reporting an error either.

Example of where this happens is `7911game.com` (warning: malicious website, ships some kind of VBScript so I assume it exploits Internet Explorer). `curl` takes a long time to download it and loads it gradually, as does https://github.com/jayjamesjay/http_req

Shnatsel:
I've tried setting really high timeouts for both connection and read (5_000_000) but the issue persists.

lolgesten:
5_000_000 is a high timeout indeed. #32 might do something to address this, but it seems strange an infinite timeout would be any better than 5_000_000.

jsha:
I've tried and failed to reproduce this. As of right now, on my Ubuntu 20.04 system, `7911game.com` produces connection timeouts with both ureq and curl:

```
$ time cargo run http://7911game.com
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/curl3 'http://7911game.com'`
Error: Oops("Connection Failed: Connection timed out (os error 110)")

real    2m10.222s
user    0m0.036s
sys     0m0.018s
```

```
$ time curl http://7911game.com -i -vvv
*   Trying 39.96.202.49:80...
* TCP_NODELAY set
* connect to 39.96.202.49 port 80 failed: Connection timed out
* Failed to connect to 7911game.com port 80: Connection timed out
* Closing connection 0
curl: (28) Failed to connect to 7911game.com port 80: Connection timed out

real    2m11.390s
user    0m0.000s
sys     0m0.016s
```

I've also adapted one of the unittests from #67 to dribble out data every 10ms for 300 seconds, and found that ureq successfully fetches the whole thing.

@Shnatsel, does this still reproduce for you? Do you have a more self-contained test case?

Shnatsel:
I suppose that website has fixed their weird behavior, whatever it was.

I've discovered this issue by downloading frontpages of the [top million websites](https://blog.majestic.com/development/majestic-million-csv-daily/) with a bunch of different HTTP clients and comparing the results. A more thorough writeup can be found [here](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6).

I suppose that's the easiest way to discover cases when it happens, and it might also uncover other issues.

jsha:
[...]

PR #32, opened 2020-01-17
Disable the native system timeout for socket

RazvanRotari:
Fix for https://github.com/algesten/ureq/issues/30

Based on the document of TcdStream to disable timeout we need to pass None to TcpSocket::set_read_timeout/TcpSocket::set_write_timeout

https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_read_timeout


lolgesten:
Thanks much!

RazvanRotari:
I have a test case for this, but don't know how to automate it.  I just compared the output from curl to the one from ureq for the url in the bug.

Issue #19, opened 2019-10-19
Please clarify the license


akhilles:
The Cargo manifest suggests that this project is dual licensed under MIT OR Apache-2.0. However, there are no LICENSE files and the README suggests it's only licensed under MIT.

If maximum compatibility with the Rust ecosystem is desired, I would follow the recommendations at https://rust-lang-nursery.github.io/api-guidelines/necessities.html#crate-and-its-dependencies-have-a-permissive-license-c-permissive.

algesten:
Fixing. Thanks!

PR #34, opened 2020-01-18
Add missing license files

waldyrious:
Follow-up of commit ee40b2cff, which addressed #19. That change included relative links in the README to files that don't exist in the repository.

I've also submitted PR https://github.com/rust-lang/api-guidelines/pull/214, adding the necessary clarifications to the guidelines that @akhilles mentioned in #19, so that others aren't misled to the same overlook when following them.


algesten:
Thanks!

PR #36, opened 2020-01-26
Add README badges for crates.io and documentation

MichaelAquilina:
Add convenient shortcuts to commonly used links in rust projects.

This is particularly useful for the documentation badge as the README doesn't provide the same level of detail that the documentation on docs.rs


algesten:
Thanks

PR #38, opened 2020-02-05
Set a default content type for JSON requests

robyoung:
When sending a JSON request a Content-Type of `application/json` is usually wanted. This is often set as a default for JSON methods by HTTP clients so can be confusing when it is not set. However, we do not want to prevent the user from setting their own Content-Type.


lolgesten:
This is out in 0.11.4

PR #41, opened 2020-03-03
Bump rustls + co

nickbabcock:
Hopefully this isn't as urgent as an upgrade, as rustls 0.17 didn't switch major ring versions -- but I figured I'd send a PR to keep on top of dependencies anyways.


algesten:
Thanks

PR #42, opened 2020-03-13
proxy support

rustysec:
added support for unauthenticated proxies as well as basic auth support.


rustysec:
I mostly tested this with `tinyproxy` against http and https websites, everything seems to be working well.

algesten:
This looks excellent!

PR #44, opened 2020-04-03
Improve chunked transfer upload performance

nviennot:
This implementation improves over chunked_transfer's Encoder + io::copy with the following performance optimizations:

1) It avoid copying memory
2) chunked_transfer's Encoder issues 4 separate write() per chunk. This is costly overhead. Instead, we do a single write() per chunk

The measured benefit on a Linux machine is a 50% reduction in CPU usage on a https connection.

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,6 +1,5 @@
 use crate::stream::Stream;
-use chunked_transfer;
-use std::io::{copy, empty, Cursor, Read, Result as IoResult};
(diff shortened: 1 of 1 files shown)


nviennot:
It's true, but that dependency is pretty light, and doesn't have dependencies, so I wouldn't worry about it.

Performance wise, on the download side, rustls has a pretty big issue: https://github.com/ctz/rustls/pull/354 -- once fixed, ureq will be blazing fast :) :)

PR #45, opened 2020-04-04
Use platform certificates. Useful in Entreprise configurations

nviennot:


Cargo.toml
```diff
@@ -17,7 +17,7 @@ all-features = true
 default = ["tls", "cookies"]
 json = ["serde_json"]
 charset = ["encoding"]
-tls = ["rustls", "webpki", "webpki-roots"]
+tls = ["rustls", "rustls-native-certs", "webpki"]
```
algesten, line 20: I personally think it's better security wise to not trust the host's root certs. So many "enterprise" have installed MITM attack proxies.

But there are certainly situations where this could be useful too.

Could you explore making this a feature that is off by default? `native-certs`. Since turning off dependencies depending on features, it's probably best to add both webpki-roots and rustls-native-certs if the feature is on.

nviennot: I made the feature optional, see the new commit.

> So many "enterprise" have installed MITM attack proxies
Just MITM proxies. The enterprise is not attacking its employees. I mean, what's the threat model? The entreprise has root access on employees machines, employees don't.

algesten: > Just MITM proxies. The enterprise is not attacking its employees. I mean, what's the threat model? The entreprise has root access on employees machines, employees don't.

You're right, I should use less divise language. I just feel strongly against this practice.

algesten:
PR looking good!

PR #47, opened 2020-04-10
Check status code that has no response body

kkazuo:
This patch fix receiving response with HTTP status code 204 or 304.
These status codes mean that there is no response body.


algesten:
Sweet! Thanks!!!

PR #48, opened 2020-04-10
Add support for application/x-www-form-urlencoded

### LINKED ISSUE #13: Please implement forms with application/x-www-form-urlencoded
While the library does many things without the dependency baggage, this one feature is crucial for many uses of post. Not supporting it precludes usage of this library. 

Please consider expediting.

bbx0:
This adds a new method `send_form` with support for `application/x-www-form-urlencoded` content to fix #13. The method takes a sequence of (key, value) pairs.

```rust
//using a slice
let _ = ureq::post("/my_page").send_form(&[("foo", "bar"),("foo2", "bar2")]);
//or a vec
let _ = ureq::post("/my_page").send_form(vec![("foo", "bar"),("foo2", "bar2")]);
```

The `QString` serialization via `percent_encoding` is not fully compliant ([1] and [2]), so the `form_urlencoded` serialization of `url` is used, which is already a dependency. 

`charset` is not supported as only `utf-8` appears to be conforming ([3]) with this `Content-Type`.

This pull request does not add `Serde` support but if someone else want's to take a look `serde_urlencoded` ([4]) would probably be the way to go.

Fixes #13 

[1]: https://www.w3.org/TR/html4/interact/forms.html#h-17.13.4.1
[2]: https://github.com/algesten/qstring/issues/3
[3]: https://url.spec.whatwg.org/#application/x-www-form-urlencoded
[4]: https://docs.rs/serde_urlencoded/0.6.1/serde_urlencoded/fn.to_string.html

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,9 +1,10 @@
+use std::borrow::Borrow;
 use std::io::Read;
 use std::sync::{Arc, Mutex};
 
 use lazy_static::lazy_static;
 use qstring::QString;
-use url::Url;
+use url::{form_urlencoded, Url};
 
 use crate::agent::{self, Agent, AgentState};
 use crate::body::Payload;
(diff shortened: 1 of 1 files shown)


src/request.rs
```diff
@@ -181,6 +182,37 @@ impl Request {
...
+    ///     .send_form(&[("foo", "bar"),("foo2", "bar2")]);
+    /// println!("{:?}", r);
+    /// }
+    /// ```
+    pub fn send_form<I, K, V>(&mut self, pairs: I) -> Response
+    where
+        I: IntoIterator,
+        I::Item: Borrow<(K, V)>,
+        K: AsRef<str>,
+        V: AsRef<str>,
```
algesten, line 205: Nice thanks!

This breaks the pattern in the current API where we use `&str` and not `AsRef<str>`.

```rust
pub fn query(&mut self, param: &str, value: &str) -> &mut Request {
```

Maybe we should stick to the variant the rest of the API uses in this PR and then make a separate PR switching the entire API over to `AsRef<str>`? What do you think?

bbx0: Sure, sounds good! I changed the signature to be more consistent with the rest of the API.

```rust
pub fn send_form(&mut self, data: &[(&str, &str)]) -> Response {
```

Advancing to `AsRef` in a separate PR sounds very reasonable to me.

algesten:
Looks great! Thanks!

PR #49, opened 2020-04-11
Add SOCKS5 feature.

sklv:
Adds support for SOCKS5 proxying via the socks crate.

Despite a few attempts I was not able to get the existing Proxy functionality working. I think it's intended to be an HTTP Connect proxy. If it worked before, I think it should still work, but I've not been able to test it.

I also marked the Proxy connect function as (crate) as it returns a string which is only useful to ureq internally. I don't think it makes sense to expose it publically.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -272,6 +272,17 @@ fn connect_socks5(
         ));
     }
 
+    // Since Socks5Stream doesn't support set_read_timeout, a suboptimal one is implemented via
+    // thread::spawn.
+    // # Happy Path
+    // 1) thread spawns 2) get_socks5_stream returns ok 3) tx sends result ok
+    // 4) slave_signal signals done and cvar notifies master_signal 5) cvar.wait_timeout receives the done signal
+    // 6) rx receives the socks5 stream and the function exists
+    // # Sad path
+    // 1) get_socks5_stream hangs 2)slave_signal does not send done notification 3) cvar.wait_timeout times out
+    // 3) an exception is thrown.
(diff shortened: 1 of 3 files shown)


src/stream.rs
```diff
@@ -202,27 +229,128 @@ pub(crate) fn connect_host(unit: &Unit, hostname: &str, port: u16) -> Result<Tcp
...
+        thread::spawn(move || {
+            let (lock, cvar) = &*slave_signal;
+            if tx
+                .send(get_socks5_stream(&proxy, &proxy_addr, &host_addr))
+                .is_ok()
+            {
+                let mut done = lock.lock().unwrap();
+                *done = true;
+                cvar.notify_one();
+            }
```
algesten, line 289: Ouch, I understand why you're doing this, but it feels wrong to launch a thread. Maybe make a code comment about:

1. Why the need for a thread.
2. A 'TODO' about approaching the upstream crate so that `Socks5Stream::connect...` can take a timeout.
3. Why `tx.send(…).is_ok()` being false is an expected code path.

sklv: Thanks for the quick response.
Please have a look at the 3 most recent commits and let me know if you'd like any further changes.

sklv: BTW, I completely agree about it being wrong to launch a thread, and trying to fix it upstream is my intention. I have briefly looked at Socks5Stream and I can see it using a BufReader internally, so it wasn't immediately obvious to me how to change that hence my choice to implement something basic initially.

PR #56, opened 2020-05-02
Make Response::into_json deserialize into a serde DeserializeOwned

paolobarbolini:
This removes the necessity to take the result of Response::into_json and having to convert it into a struct by using serde_json::from_value

This adds no new dependencies since serde_json already depends on serde. Users of ureq will have to include `serde_derive` either by importing it directly or by using serde with the `derive` feature, unless they want to manually implement `Deserialize` on their structs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ all-features = true
 
 [features]
 default = ["tls", "cookies"]
-json = ["serde_json"]
+json = ["serde", "serde_json"]
 charset = ["encoding"]
 tls = ["rustls", "webpki", "webpki-roots"]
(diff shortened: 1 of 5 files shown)


oshanz:
thanks @paolobarbolini I was looking for this :slightly_smiling_face:

algesten:
@paolobarbolini thanks!

This change makes a lot of sense, and it's the way I should have done it. However. Here we are and this would be a breaking API change.

Can we add this under another name and keep the old one (with the tests)?

The name is up for bike shedding. `into_deserialize()`, `into_json_value()`, `into_json_deserialize()`

paolobarbolini:
Fixed @algesten

Issue #37, opened 2020-02-04
Providing a client certificate


robyoung:
Are there any plans to support providing a client certificate or custom root certificates?

lolgesten:
@robyoung I'm going to clarify the future of ureq by updating the README soon. But in a nutshell.

I'll keep maintaining ureq as is cause there seems to be interest in a _blocking_ simple http client. With maintaining I mean keep dependencies somewhat fresh and address any bad bugs. I will however not personally implement new features in ureq, but I do welcome PR with open arms.

The code base is extremely simple, one might even call naive. It's a good project to hack on as first learning experience in Rust. I will uphold some base line of code hygiene, but won't block a PR due to something being a bit inelegant.

robyoung:
That is really good to know and thank you for your quick response.

robyoung:
Implemented in #58

PR #58, opened 2020-05-14
Allow TLS client config to be overridden

robyoung:
See: [ClientConfig](https://docs.rs/rustls/latest/rustls/struct.ClientConfig.html)

Addresses: #37

I am adding this so that I can provide a client certificate. However providing a custom `ClientConfig` allows for changing a lot more, such as custom trust anchors.

I have done the integration testing with https://badssl.com to avoid having to spin up a TLS terminating server locally, hope that's ok.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,5 +1,7 @@
 use std::io::Read;
 use std::sync::{Arc, Mutex};
+#[cfg(feature = "tls")]
+use std::fmt;
 
 use lazy_static::lazy_static;
 use qstring::QString;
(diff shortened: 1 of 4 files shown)


davidiw:
@algesten anyway we could get this into crates.io? We're likely going to have to patch this in in the meantime.

algesten:
@davidiw hey! I've been busy with some work stuff. I'll check the PR now.

algesten:
Pushed in 1.1.0

davidiw:
Thank you so much!

PR #67, opened 2020-06-15
Add overall timeout for requests.

### LINKED ISSUE #28: No way to set timeout for the entire request, which allows DoS attacks
ureq currently does not allow specifying a timeout for the entire request (i.e. until the request body is finished transferring), which means an established connection will keep going indefinitely if the remote host keeps replying.

This is fine for the simple use cases, like a user downloading something interactively, but enables denial-of-service attacks in automated scenarios: if the remote host keeps transferring data at a really low speed, e.g. several bytes per second, the connection will be open indefinitely. This makes it easy for an attacker who can submit URLs to the server to cause denial of service through some kind of resource exhaustion - running out of RAM, networking ports, etc.

jsha:
This deprecates timeout_read() and timeout_write() in favor of
timeout(). The new timeout method on Request takes a Duration instead
of a number of milliseconds, and is measured against overall request
time, not per-read time.

Once a request is started, the timeout is turned into a deadline
specific to that call. The deadline is used in conjunction with the
new DeadlineStream class, which sets a timeout on each read according
to the remaining time for the request. Once the request is done,
the DeadlineStream is unwrapped via `.into::<Stream>()` to become
an undecorated Stream again for return to the pool. Timeouts on the
stream are unset at this point.

Still to be done:
 - Add a setting on Agent for default timeout.
 - Change header-writing code to apply overall deadline rather than
   per-write timeout.

Fixes #28.

Previously this PR said:

Open questions:
  - Should the overall timeout functionality take precedence over the
    separate connect / read / write timeouts? Should those be deprecated?
  - Adding deadline handling to ReclaimingRead felt like a weird place
    to do it, but it seemed like the only place to get ahold of a Stream
    before handing it off the the user via `into_reader()`. Is there a
    better place?
 - Would it be better to accept a Duration as the timeout instead of millis?


lolgesten:
Looking good!

> Should the overall timeout functionality take precedence over the
separate connect / read / write timeouts? Should those be deprecated?

I'm up for suggestions really. On the one hand, overall timeout is always a better idea, on the other, I'm not super keen on a breaking change. I guess the problem is that when setting overall timeout it doesn't make much sense to set the others?

> Adding deadline handling to ReclaimingRead felt like a weird place
to do it, but it seemed like the only place to get ahold of a Stream
before handing it off the the user via into_reader(). Is there a
better place?

If ReclaimingRead starts to become a bit of a multi-tool, maybe we should just rename it?

> Would it be better to accept a Duration as the timeout instead of millis?

When ureq was first made, Duration had really terrible ergonomics (it got better around rust 1.33). I still feel that milliseconds typically is a good unit from a human perspective interacting with an API. However Duration is the rust default – so I arrived at a compromise in my new project:

https://docs.rs/hreq/0.3.0/hreq/trait.RequestBuilderExt.html#tymethod.timeout

jsha:
> I guess the problem is that when setting overall timeout it doesn't make much sense to set the others?

I think the main reason to deprecate is that maintaining multiple paths of timeout code is complex. Also, though, the `read_timeout` and `write_timeout` settings probably don't do what most people think they do, and are worth deprecating for that reason alone. I agree breaking changes are undesirable. I propose to mark them with `#[deprecated]` and not remove them for quite a while, ideally when that can be batched up with other needed breaking changes.

It's worth noting that there is a potential reason for keeping the connect_timeout along with an overall timeout: Someone might want to give a long overall timeout, on the order of minutes, to allow downloading large files, but abort after only a few seconds of waiting to connect, since that suggests the host is unavailable.

> If ReclaimingRead starts to become a bit of a multi-tool, maybe we should just rename it?

Sounds like a plan! BTW I'm still trying to wrap my head around what ReclaimingRead/PoolReturnRead do. Am I correct in understanding that they use the raw stream pointer to share the stream, and their combined logic ensures only one of the two is using it at a time? Would a [`Cell`](https://doc.rust-lang.org/std/cell/) be a good way to implement the same functionality?

lolgesten:
I agree with your plan for timeouts. To sum it up:

* Deprecate read_timeout
* Deprecate write_timeout
* Keep connect_timeout
* Make overall timeout include connect_timeout, but note the reason you might want to set connect_timeout low

> ReclaimingRead/PoolReturnRead do. Am I correct in understanding that they use the raw stream pointer to share the stream, and their combined logic ensures only one of the two is using it at a time? Would a Cell be a good way to implement the same functionality?

Yes yes and yes :)

The only reason for this contortion was that `chunked-transfer` didn't let us "get back" the wrapped stream. But the upstream crate has now implemented `into_inner()` which means we can get rid of this unsafe pointer trickery. The better solution was worked on here https://github.com/algesten/ureq/pull/31 – but that PR hasn't seen any love in a while.

jsha:
Thanks for the great feedback. I've made these changes:

     - Deprecate read_timeout and write_timeout.
     - Add test case for timeout while reading headers.
     - Implement From<DeadlineStream> for Stream, to let PoolReturnRead
       extract the origin stream again and unset timeouts on the socket.
     - Take timeout as an Option<Duration> instead of a u64.

This is now ready for review.

Incidentally, while I was working on this, I felt it was inefficient to be setting a fresh timeout on every read() call. I was curious how other languages and tools handle this. Here are my rough notes:

python3 urllib3: only a per-read timeout, not an overall timeout
python3 requests: only a per-read timeout, not an overall timeout
  (https://requests.readthedocs.io/en/master/user/quickstart/#timeouts)
node http package: only a per-read timeout, not an overall timeout
 (https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback)
ruby: has a read_timeout, open_timeout (i.e. connect), no overall timeout
  (https://ruby-doc.org/stdlib-2.7.1/libdoc/net/http/rdoc/Net/HTTP.html)
wget: has --dns-timeout, --connect-timeout, --read-timeout, plus --timeout which
  sets all three to the same thing
  https://www.gnu.org/software/wget/manual/wget.html
perl LWP::UserAgent: supports a per-read timeout
Go: has an overall timeout that includes reading from body (https://golang.org/pkg/net/http/#Client)
curl: has an overall timeout.
  https://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT_MS.html
  Uses select in a loop, then reads when data is available.

So, I was surprised to see that among the systems I checked, almost none of them implement this functionality of an overall timeout. Go does implement it, but uses goroutines and async I/O to achieve it, which are not available to us in this library. Curl uses select, which ureq could in theory use without going the full async route, but it seems Rust does not expose a select API for TcpStreams.

A couple of possible refinements for the future:
 - Add a BufReader at some layer above the DeadlineStream to reduce the number of read() calls.
 - Instead of setting the socket read timeout to the remaining time, set it to, e.g. 1 second, unless the deadline is less than 1 second away, in which case set it to the remaining time. This would allow setting it once and not again until the deadline has almost arrived. This might require some internal retrying of WouldBlock errors.

src/stream.rs
```diff
@@ -35,6 +35,59 @@ pub enum Stream {
...
+        // Since we are turning this back into a regular, non-deadline Stream,
+        // remove any timeouts we set.
+        let stream = deadline_stream.stream;
+        if let Some(socket) = stream.socket() {
+            socket.set_read_timeout(None).unwrap();
+            socket.set_write_timeout(None).unwrap();
+        }
+        stream
+    }
+}
```
algesten, line 65: The rust idiomatic way for returning a wrapped value is often called `into_inner`. It's not based on traits or anything, so it isn't enforced, it's just very common to see both in the std lib and in other crates.

[`BufReader`](https://doc.rust-lang.org/std/io/struct.BufReader.html#method.into_inner) is a relevant example.

Since `DeadlineStream` is internal to ureq, I don't really care if you change or keep it like this. Just worth noting.

src/stream.rs
```diff
@@ -332,15 +389,23 @@ pub(crate) fn connect_host(unit: &Unit, hostname: &str, port: u16) -> Result<Tcp
 
     // rust's absurd api returns Err if we set 0.
     // Setting it to None will disable the native system timeout
-    if unit.timeout_read > 0 {
+    if let Some(deadline) = deadline {
+        stream
+            .set_read_timeout(Some(deadline - Instant::now()))
+            .ok();
+    } else if unit.timeout_read > 0 {
```
algesten, line 396: I think it might be worth noting in the doc for the deprecated `read_timeout` that if one sets both that and `timeout`, then timeout takes precedence.

src/stream.rs
```diff
@@ -332,15 +389,23 @@ pub(crate) fn connect_host(unit: &Unit, hostname: &str, port: u16) -> Result<Tcp
...
     } else {
         stream.set_read_timeout(None).ok();
     }
 
-    if unit.timeout_write > 0 {
+    if let Some(deadline) = deadline {
+        stream
+            .set_write_timeout(Some(deadline - Instant::now()))
+            .ok();
+    } else if unit.timeout_write > 0 {
```
algesten, line 408: Ditto doc for `write_timeout`.

src/test/agent_test.rs
```diff
@@ -123,3 +121,120 @@ fn connection_reuse() {
     }
     assert_eq!(resp.status(), 200);
 }
+
```
algesten, line 124: Your test setup is just great (so simple and to the point!), I think you should break out your tests into a separate test file (`src/test/timeout.rs`?)

src/test/agent_test.rs
```diff
@@ -123,3 +121,120 @@ fn connection_reuse() {
...
+    let dribble_handler = |mut stream: TcpStream| {
+        read_headers(&stream);
+        if let Err(e) = dribble_body_respond(&mut stream) {
+            eprintln!("sending dribble repsonse: {}", e);
+        }
+    };
+    thread::spawn(move || {
+        for stream in listener.incoming() {
+            thread::spawn(move || dribble_handler(stream.unwrap()));
+        }
```
algesten, line 171: I guess threads are pretty cheap and we don't really care that this thread will hang around for the entire test suite, right?

Would a tweak be to not have a for loop, and just accept one single connection, then exit?

src/response.rs
```diff
@@ -580,13 +587,13 @@ fn read_next_line<R: Read>(reader: &mut R) -> IoResult<String> {
 }
 
 /// Limits a `Read` to a content size (as set by a "Content-Length" header).
-struct LimitedRead<R> {
+struct LimitedRead<R: Read> {
```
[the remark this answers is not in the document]
jsha: Good point. Mainly I had found it confusing that you could create a LimitedRead that wraps something `!Read`, but then it just wouldn't do anything. But we can achieve that constraint by limiting `new`, since the fields of this struct are private. I'm guessing a type constraint on the struct itself would be important for structs that have public fields and can be constructed without going through `new`.

src/stream.rs
jsha: Good point. I like that idiom, as opposed to the more generic Into. Currently, though, PoolReturnRead has this type constraint:

```
impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
```

Introduced in #31 / #68.

I think if we want to refactor this, it should be a separate PR.

src/test/agent_test.rs
jsha: Normally I like to clean up threads when tests are done, but I don't yet know how to do this in Rust in a clean and simple way. I think I'd need to set up a condition variable and a select of some sort between listener.incoming() and the condition variable. But listener.incoming() doesn't seem to have affordances for that. I'm interested in learning how to do that, but for now I decided to just leak the thread until the tests are done.

I considered the one-accept-and-then-done approach but concluded it would be less realistic. For instance, if something triggered a retry during the tests, we would get a failure, but for the wrong reason.

Ultimately I'd like to evolve these test servers into something that can be used for more of the test cases, so developing a clean shutdown mechanism will be worthwhile.

src/stream.rs
algesten: Sounds good!

src/test/agent_test.rs
algesten: > I think I'd need to set up a condition variable and a select of some sort between listener.incoming() and the condition variable.

If you were outside rust, how does one normally do this? `listener.incoming()` is blocking, so I assume it's using some underlying syscall that also blocks. Are there variants of that syscall with a timeout like for read/write?

> I considered the one-accept-and-then-done approach but concluded it would be less realistic. For instance, if something triggered a retry during the tests, we would get a failure, but for the wrong reason.

Fair enough!

> Ultimately I'd like to evolve these test servers into something that can be used for more of the test cases, so developing a clean shutdown mechanism will be worthwhile.

You got plans! 😎 

My attention is mostly on hreq now, and incidentally just this week I've been hacking on a server side to it. Today I wrote the very rows where I also [need a shutdown mechanism](https://github.com/algesten/hreq/blob/server/src/server/mod.rs#L89-L91), though it might be simpler because of async.

algesten:
Just checking before I merge, 

Do you want to squash the commits manually, or should I do it with the github "Squash and merge", or do you prefer to keep the commit history intact?

jsha:
> Do you want to squash the commits manually, or should I do it with the github "Squash and merge", or do you prefer to keep the commit history intact?

If you would do Squash and Merge in GitHub, that's my preferred workflow. I've updated the PR description to be a suitable squash commit description (you can omit the part that says "Previously"; I left that in so the conversation would make sense).

> If you were outside rust, how does one normally do this? listener.incoming() is blocking, so I assume it's using some underlying syscall that also blocks. Are there variants of that syscall with a timeout like for read/write?

I had thought that Go handled this by selecting across a channel of incoming sockets and a channel indicating "done" status, but it looks like it actually just closes the socket on which Accept is being called: https://github.com/golang/go/blob/master/src/net/http/server.go#L2738-L2745. I'll give that a try, and may have a followup PR ready sooner than I thought.

By the way, one thing I wanted to flag as I was doing a last check-through: Request's timeout method takes a Duration, not an Option<Duration>. My thinking: Since Requests start with no timeout, there's almost never a need to unset the timeout on a request. Just build it without one in the first place. This could make things more awkward if you decide to add a default timeout in the future, since it means people can't explicitly unset it - they'd just have to set a ludicrously high timeout instead. I think this is an acceptable API tradeoff.

algesten:
> Just build it without one in the first place. This could make things more awkward if you decide to add a default timeout in the future, since it means people can't explicitly unset it - they'd just have to set a ludicrously high timeout instead. I think this is an acceptable API tradeoff.

I'm ok with that trade-off. I don't think a user would ever expect a default timeout.

Shnatsel:
> Should the overall timeout functionality take precedence over the separate connect / read / write timeouts? Should those be deprecated?

I'd say both are useful and presence of one should not deprecate the other.

Read timeout and timeout for the entire request serve different purposes. Read timeout is for interrupting connections where the remote host has suddenly stopped responding, before the TCP timeout of ~5 minutes expires. This is a universally useful thing to have, and it's reasonable to have a default timeout of this kind. The same goes for connection timeout, although connection and read timeouts can be collated to simplify the API.

Timeout for the entire request on the other hand is for more specialized use cases - e.g. fetching data from URLs obtained from an untrusted source, where the remote host may be malicious and keep the connection open indefinitely to cause DoS or resource exhaustion.

algesten:
@Shnatsel The overall timeout works by manipulating the connect/read/write timeouts though. The relationship between overall timeout and the non-deprecated connect timeout is that connect can be set shorter than the overall. Would that be the relationship you'd expect between overall vs read and overall vs write too? 

I.e. You can only set read/write _shorter_ than overall if you have set an overall.

Shnatsel:
> I.e. You can only set read/write shorter than overall if you have set an overall.

Yes, that's the behavior I'd expect. However, there are uses for having a read timeout without an overall timeout (e.g. downloading a large file and not knowing in advance how long it will take), so I don't think connect or read timeouts should be deprecated.

jsha:
@Shnatsel makes a good point. I'll send a PR un-deprecating read/write timeout for now; we can always reconsider later if the maintenance burden winds up being significant.

PR #68, opened 2020-06-16
Remove last `unsafe` usage and forbid further usages (builds on #31)

jsha:
This builds on @EkardNT's excellent work on #31. All I've done is merged the latest code and bumped the version dependency. @algesten, [your comment](https://github.com/algesten/ureq/pull/31#issuecomment-619492748) mentions that some tidying is needed because a commit message "talks about .unwrap() which was renamed to into_inner()." Would you like me to try and rebase @EkardNT's commits into a single commit that talks about the correct method?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ socks-proxy = ["socks"]
 
 [dependencies]
 base64 = "0.12"
-chunked_transfer = "1"
+chunked_transfer = "1.2.0"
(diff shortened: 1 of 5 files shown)


algesten:
@jsha thanks for doing this! I think we can leave the commits as is. Let's merge!

algesten:
Today I learn: there are situations where git rebase will conflict, but git merge is ok. I typically want rebase, but the github UI showed me this:

image

I rebased manually locally and pushed to master.

PR #81, opened 2020-06-22
Limit max idle connections.

### LINKED ISSUE #77: Limit maximum size of connection pool
Right now, connections stay in the pool indefinitely. If someone makes requests to a wide variety of hosts, that can quickly fill up the pool. Each entry in the pool uses an FD, and eventually a program will hit ulimit_nofile and stop working.

We should set a max size for the pool, and when a new connection needs to be added but the pool is full, remove one of the existing connections. A couple of options for how to remove:

 - We could scan the pool for connections where the server has already closed them. Such connections are taking up resources needlessly. But if the connection pool is full in steady state, this could be a lot of scanning.
 - We could use an LRU cache.
 - We could remove an entry at random.

jsha:
Adds limit of 100 connections to ConnectionPool. This is implemented
with a VecDeque acting as an LRU. When the pool becomes full, the oldest
stream is popped off the back from the VecDeque and also removed from
the map of PoolKeys to Streams.

Fixes #77


jsha:
Thought about testing on this and realized we can't test effectively without a mock Resolver: #82. I've done some manual testing and this works as expected. I think we should proceed with this change and tag a release, and pursue more robust testing for the next release.

jsha:
Update: In a smoke test of the top 1M domains, I was able to produce a panic due to:

"invariant failed: key existed in lru but not in recycle"

I'll be debugging further, but in the meantime this shouldn't be merged.

jsha:
Found the bug: When .add() was called on a key that already existed in the recycle map, it would overwrite the existing entry, and would add a new entry to the lru, resulting in two entries in lru that correspond to just a single entry in the recycle map. Added a test to catch that case. Ready for review.

PR #85, opened 2020-06-23
Add smoke test in examples.

jsha:
This loads a list of top domain names (e.g. from
https://tranco-list.eu/) and tries to fetch them all, in parallel. This
can be used to exercise ureq and find panics.

### DIFF
--- a/examples/smoke-test/main.rs
+++ b/examples/smoke-test/main.rs
@@ -1,4 +1,3 @@
-#![feature(duration_constants)]
 use chrono::Local;
 use rayon::prelude::*;
 use rayon_core;
@@ -39,10 +38,9 @@ impl fmt::Display for Oops {
     }
 }
 
-type Bytes = Vec<u8>;
 type Result<T> = result::Result<T, Oops>;
 
-fn get(agent: &ureq::Agent, url: &String) -> Result<Bytes> {
+fn get(agent: &ureq::Agent, url: &String) -> Result<Vec<u8>> {
     let response = agent
         .get(url)
         .timeout_connect(5_000)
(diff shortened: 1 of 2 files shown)


examples/smoke-test/main.rs
```diff
@@ -0,0 +1,112 @@
...
+            env::current_exe()?
+        );
+        return Ok(());
+    }
+    let file = std::fs::File::open(args.skip(1).next().unwrap())?;
+    let bufreader = BufReader::new(file);
+    let hostnames = bufreader
+        .lines()
+        .map(|line| Ok(line?.rsplit(",").next().unwrap().to_string()))
+        .collect::<io::Result<Vec<String>>>()?;
```
lolgesten, line 102: Here you don't need to `collect`, cause `for var in iter { ... }` works over iterator. But I guess you'd push the error handling into the for loop instead, so maybe it's cleaner this way.

examples/smoke-test/main.rs
```diff
@@ -0,0 +1,112 @@
...
+impl fmt::Display for Oops {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+type Bytes = Vec<u8>;
+type Result<T> = result::Result<T, Oops>;
+
+fn get(agent: &ureq::Agent, url: &String) -> Result<Bytes> {
```
lolgesten, line 45: For someone just scanning the code, `Bytes` often means this crate https://docs.rs/bytes/0.5.5/bytes/#bytes

Doesn't matter much though.

Issue #43, opened 2020-03-26
No automatic chunked Transfer Encoding when using Request.send


Deluvi:
On `ureq` version `0.12.0`:

When using the [`Request.send`](https://docs.rs/ureq/0.12.0/ureq/struct.Request.html#method.send) method, no headers are set to indicate some body content to the server. It causes some servers to completely ignore the body (like `tiny_http`).
The `Request.send` method should automatically enable the `chunked` encoding.

A workaround for this issue is to set the Transfer-Encoding header prior sending the request to enable the `chunked` transfer:
```
request.set("Transfer-Encoding", "chunked");
let response = request.send(body_reader);
```

I am doing this issue more as a warning for people like me that will stumble upon this gotcha. I have noticed that you are actually working a brand new version on `ureq` that probably fixed that issue already so it's maybe not a good idea to spend time fixing this issue. Maybe it should just be mentioned in the documentation of `Request.send` that the user may want to set the `chunked` transfer.

algesten:
You're right! It could be good to set this header if we haven't set a content-length already. At this point it would be a (subtle) breaking change, so not entirely how to handle it.

Maybe submit a PR and we do it when we bump to 0.13.0?

I'm not rewriting `ureq`, it will stay the same (sync forever), and I will keep maintaining it until there's 0 interest in it. My rewrite efforts are in [`hreq`](https://github.com/algesten/hreq).

Deluvi:
Alright, thank you for your answer! I'll submit a PR when I have some time: I have a pretty good idea where to do the edit.

jsha:
[Here's the relevant section of RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.3):

>    The presence of a message body in a request is signaled by a
   Content-Length or Transfer-Encoding header field.  Request message
   framing is independent of method semantics, even if the method does
   not define any use for a message body.

So I think tiny_http is right to ignore message bodies if neither Content-Length nor Transfer-Encoding is provided.

Right now there are two ways to provide a message body in ureq: `.send_string()` and `.send()`. If you want to send arbitrary binary content (i.e. not UTF-8), you need to use `.send()`. The user of `ureq` can currently choose between Content-Length (if they know the size of the data they'll send) or Transfer-Encoding (if they don't). It would be great to add some support to make it easier to do the right thing, but keep in mind making `.send()` always set Transfer-Encoding would effectively force chunked encoding for anyone who wants to send message bodies containing arbitrary bytes.

[...]

PR #73, opened 2020-06-20
Update documentation.

jsha:
Synchronize goals section between README and rustdoc, and add two goals
(blocking API; no unsafe) that were mentioned elsewhere in the README.

Add error handling to examples for the module and for the Response
object.

And a section on synthetic errors to the top-level module documentation.

Related to #55

PR #86, opened 2020-06-23
Set chunked Transfer-Encoding when using `send`

Deluvi:
Related issue: #43 

Previously, using `.send()` on `Request` would require to set either the Transfer-Encoding or the Content-Length header.

In an effort to provide better ergonomics for this library and to avoid making users fall in a not-so obvious pitfall, the library should build a valid Request without asking the user to mess around with the headers.

This PR attempts to fix this issue: if the user use `.send()` to provide an unknown sized reader, the chunked Transfer-Encoding will be used. Of course, there are prior checks to ensure we do not override the user wish, like if the user already set a Content-Length or a Transfer-Encoding.

This PR also fix an edge case where the Content-Length would not be automatically set if the Transfer-Encoding is set but not chunked.

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -42,11 +42,21 @@ impl Default for Payload {
     }
 }
 
+/// The size of the body.
+///
+/// *Internal API*
+#[derive(Debug)]
+pub(crate) enum BodySize {
+    Empty,
+    Unknown,
+    Known(u64),
+}
+
 /// Payloads are turned into this type where we can hold both a size and the reader.
 ///
 /// *Internal API*
 pub(crate) struct SizedReader {
-    pub size: Option<usize>,
+    pub size: BodySize,
     pub reader: Box<dyn Read + 'static>,
 }
 
@@ -57,7 +67,7 @@ impl ::std::fmt::Debug for SizedReader {
 }
 
 impl SizedReader {
-    fn new(size: Option<usize>, reader: Box<dyn Read + 'static>) -> Self {
+    fn new(size: BodySize, reader: Box<dyn Read + 'static>) -> Self {
         SizedReader { size, reader }
     }
 }
@@ -65,7 +75,7 @@ impl SizedReader {
 impl Payload {
     pub fn into_read(self) -> SizedReader {
         match self {
-            Payload::Empty => SizedReader::new(Some(0), Box::new(empty())),
+            Payload::Empty => SizedReader::new(BodySize::Empty, Box::new(empty())),
             Payload::Text(text, _charset) => {
                 #[cfg(feature = "charset")]
                 let bytes = {
@@ -78,20 +88,20 @@ impl Payload {
                 let bytes = text.into_bytes();
                 let len = bytes.len();
                 let cursor = Cursor::new(bytes);
-                SizedReader::new(Some(len), Box::new(cursor))
+                SizedReader::new(BodySize::Known(len as u64), Box::new(cursor))
             }
             #[cfg(feature = "json")]
             Payload::JSON(v) => {
                 let bytes = serde_json::to_vec(&v).expect("Bad JSON in payload");
                 let len = bytes.len();
                 let cursor = Cursor::new(bytes);
-                SizedReader::new(Some(len), Box::new(cursor))
+                SizedReader::new(BodySize::Known(len as u64), Box::new(cursor))
             }
-            Payload::Reader(read) => SizedReader::new(None, read),
+            Payload::Reader(read) => SizedReader::new(BodySize::Unknown, read),
             Payload::Bytes(bytes) => {
                 let len = bytes.len();
                 let cursor = Cursor::new(bytes);
-                SizedReader::new(Some(len), Box::new(cursor))
+                SizedReader::new(BodySize::Known(len as u64), Box::new(cursor))
             }
         }
     }
--- a/src/body.rs
+++ b/src/body.rs
@@ -65,7 +65,7 @@ impl SizedReader {
 impl Payload {
     pub fn into_read(self) -> SizedReader {
         match self {
-            Payload::Empty => SizedReader::new(None, Box::new(empty())),
+            Payload::Empty => SizedReader::new(Some(0), Box::new(empty())),
             Payload::Text(text, _charset) => {
                 #[cfg(feature = "charset")]
                 let bytes = {
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -47,12 +47,12 @@ impl Unit {
     pub(crate) fn new(req: &Request, url: &Url, mix_queries: bool, body: &SizedReader) -> Self {
         //
 
-        let is_chunked = req
+        let (is_transfer_encoding_set, mut is_chunked) = req
             .header("transfer-encoding")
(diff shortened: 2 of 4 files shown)


Deluvi:
I have three remarks about this PR:

The Content-Length header is not set if the size of the body is 0 to not break the previous behavior. The HTTP standard does ask for a Content-Length header in some HTTP methods requests such as POST, PUT and PATCH. This might be better addressed on another PR.

[According to some RFC](https://tools.ietf.org/html/rfc7230#section-3.3.1):
> A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field.

It does not make sense in the case where the Transfer-Encoding is something else than `chunked`. [According to the Mozilla wiki](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#Directives), this restriction seems to only apply on the `chunked` encoding. On this PR, I took Mozilla side and automatically set the Content-Length if the size is known and if the Transfer-Encoding is not `chunked`.

Finally, I do not check if the user has set both the `chunked` Transfer-Encoding and the Content-Length headers, which is not allowed in the HTTP1.1 standard.

src/body.rs
```diff
@@ -65,7 +65,7 @@ impl SizedReader {
 impl Payload {
     pub fn into_read(self) -> SizedReader {
         match self {
-            Payload::Empty => SizedReader::new(None, Box::new(empty())),
+            Payload::Empty => SizedReader::new(Some(0), Box::new(empty())),
```
algesten, line 68: Is this change needed for the overall logic of this PR?

Deluvi: I needed a way to differentiate `Payload::Reader` and `Payload::Empty`, `None` allowed me to express that the `SizedReader` has an unknown size, as opposed to `Some(0)` which express an empty Reader.

algesten:
I've been thinking about this. There are basically three ways the user can provide a body:

1. A call without a body, GET, HEAD, etc.
2. Send a body with unknown size (results in chunked)
3. Send a body with known size (sets content-length)

The code did already not capture this intention because it is using an `Option<usize>` of value `None` covered both case 1 and 2. Changing case 1 to be represented as `Some(0)` feels wrong – it means we can't distinguish that case from say sending a 0 byte file. It's no worse than today's code, but it's a sideways move.

One intention of this PR, and a potential follow-up PR is to bring us closer to spec. In that spirit, regardless of how that next PR will look, I think we should introduce an enum that preserves the user intent so we can properly deal with the logic later.

Here's a suggestion, totally up for bike shedding:

```rust
pub enum BodySize {
  NoBody,
  Unknown,
  Known(u64),
}
```

I deliberately chose `u64` over `usize`.

jsha:
@Deluvi said:
> > A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field.
> 
> It does not make sense in the case where the Transfer-Encoding is something else than chunked. According to the Mozilla wiki, this restriction seems to only apply on the chunked encoding. On this PR, I took Mozilla side and automatically set the Content-Length if the size is known and if the Transfer-Encoding is not chunked.

A little earlier [in that same section](https://tools.ietf.org/html/rfc7230#section-3.3.1), RFC 7230 says:

> If any transfer coding
   other than chunked is applied to a request payload body, the sender
   MUST apply chunked as the final transfer coding to ensure that the
   message is properly framed. 

So it does make sense to the Content-Length MUST NOT coexist with Transfer-Encoding, because if there is any Transfer-Encoding header, the last element of the last Transfer-Encoding header MUST be "chunked."

I think the right behavior for `.send()` here is:

 - If "Transfer-Encoding" and "Content-Length" were both set by the user, error.
 - If "Content-Length" was set by the user, copy the bytes without using chunked encoding (this is current behavior; I documented it wrong in #73).
 - If "Transfer-Encoding" was set by the user, and the last element of the last Transfer-Encoding header was "chunked," use chunked encoding and don't add any headers.
 - Otherwise, use chunked encoding and add a "Transfer-Encoding: chunked" header.

@algesten said:
>     pub enum BodySize {
>       NoBody,
>       Unknown,
>       Known(u64),
>     }
> I deliberately chose u64 over usize.

This seems like a good idea. Out of curiosity, why u64 over usize?

Also, to make it explicit since it took me a minute to figure out: the reason to make this distinction is so that on NoBody requests (GET etc) we want to send neither Content-Encoding nor Transfer-Encoding.

algesten:
@jsha 

> This seems like a good idea. Out of curiosity, why u64 over usize?

Since usize is either 32 or 64 bits depending on arch, it makes a lot of sense that for example `Vec::len()` returns an usize. It's highly unlikely that this would ever be a problem, but the size of the body would be arbitrarily limited to 4GB if ureq is compiled on a 32 bit arch.

> the reason to make this distinction is because on NoBody requests (GET etc) we want to send neither Content-Encoding nor Transfer-Encoding.

Exactly.

Deluvi:
Thank you you two for your feedback, very insightful !

I was on the edge to write something like `BodySize` when I was writing this PR. I tried to keep it simple so I didn't add it, but since you are suggesting it, I will add it gladly ! :smile: 

> A little earlier in that same section, RFC 7230 says...

Nice catch @jsha :+1:  ! I missed that part of the RFC, but now it does make sense !

In your description of `.send()`, what should we do if Transfer-Encoding is set but does not contain `chunked` ? If we stick to the spec, we should either automatically add `chunked` or error.

algesten:
> In your description of .send(), what should we do if Transfer-Encoding is set but does not contain chunked ? If we stick to the spec, we should either automatically add chunked or error.

I think we can just error at this point – if someone ever complains about it we can fix later.

src/unit.rs
```diff
@@ -47,12 +47,19 @@ impl Unit {
...
+                let is_transfer_encoding_set = !enc.is_empty();
+                let last_encoding = enc.split(',').last();
+                let is_chunked = last_encoding
+                    .map(|last_enc| last_enc.trim() == "chunked")
+                    .unwrap_or(false);
+                (is_transfer_encoding_set, is_chunked)
+            })
             // otherwise, no chunking.
-            .unwrap_or(false);
+            .unwrap_or((false, false));
```
algesten, line 62: I think this block could be more clear like this:

```rust
        let transfer_enc_header = req.header("transfer-encoding");

        let is_transfer_encoding_set =
            transfer_enc_header.is_some() && transfer_enc_header != Some("");

        let mut is_chunked =
            transfer_enc_header.and_then(|enc| enc.split(',').last()) == Some("chunked");
```

src/unit.rs
```diff
@@ -64,8 +71,21 @@ impl Unit {
...
+                    BodySize::Known(size) => {
+                        extra.push(Header::new("Content-Length", &format!("{}", size)))
+                    }
+                    BodySize::Unknown => {
+                        if !is_transfer_encoding_set {
+                            extra.push(Header::new("Transfer-Encoding", "chunked"));
+                            is_chunked = true;
+                        }
+                    }
+                    BodySize::Empty => {}
```
algesten, line 88: This looks so nice now! ✨

algesten:
Let's land this

PR #87, opened 2020-06-24
Remove some panics in deadline checking.

jsha:
time_until_deadline had a time of check to time of use problem - the
deadline could pass between a call to checked_duration_since and the
evaluation of `deadline - now` (which panics if the result would be
negative). Resolve that by flipping the order of
checked_duration_since's arguments and using the result rather than
ignoring it.

Also there were three places that called deadline - now(), which could
panic. Replace those with time_until_deadline().

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -81,12 +81,12 @@ impl Read for DeadlineStream {
 // then. Otherwise return a TimedOut error.
 fn time_until_deadline(deadline: Instant) -> IoResult<Duration> {
     let now = Instant::now();
-    match now.checked_duration_since(deadline) {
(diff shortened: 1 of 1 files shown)


src/stream.rs
```diff
@@ -81,12 +81,12 @@ impl Read for DeadlineStream {
 // then. Otherwise return a TimedOut error.
 fn time_until_deadline(deadline: Instant) -> IoResult<Duration> {
     let now = Instant::now();
-    match now.checked_duration_since(deadline) {
-        Some(_) => Err(IoError::new(
+    match deadline.checked_duration_since(now) {
```
algesten, line 84: Good catch!

PR #83, opened 2020-06-22
1.3.0

jsha:
To be merged after #81.

PR #88, opened 2020-06-24
Fix reference to Instant in socks feature.

jsha:
The reference to time::Instant under feature = socks-proxy was
incorrectly scoped, and should have been just Instant. This breaks the
doc build and any builds that use feature = socks-proxy.


jsha:
Once this is merged, I think we're ready for 1.3.0. Want to merge #83 after this one, then tag and push to crates.io?

PR #89, opened 2020-06-24
Add a GitHub action for building docs.

jsha:
This helps ensure we don't break that doc build.


algesten:
@jsha because of #90, I think we need to amend the features to be in sync.

jsha:
Good point! Sync'ed up the features.

PR #98, opened 2020-06-28
Factor out TestServer.

jsha:
This creates a struct that encapsulates the test server setup, and adds
a shutdown method to clean up the thread when done.

### DIFF
--- a/src/test/agent_test.rs
+++ b/src/test/agent_test.rs
@@ -1,6 +1,7 @@
 use crate::test;
-use std::io::{BufRead, BufReader, Read, Write};
-use std::thread;
+use crate::test::testserver::*;
+use std::io::{self, Write};
+use std::net::TcpStream;
 use std::time::Duration;
 
 use super::super::*;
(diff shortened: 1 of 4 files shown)


src/test/agent_test.rs
```diff
@@ -1,6 +1,7 @@
 use crate::test;
-use std::io::{BufRead, BufReader, Read, Write};
-use std::thread;
+use crate::test::testserver::*;
```
algesten, line 2: I typically prefer to enumerate the imports, or it's hard to see where things come from.

src/test/testserver.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Read a stream until reaching a blank line, in order to consume
+// request headers.
+pub fn read_headers(stream: &TcpStream) {
+    for line in BufReader::new(stream).lines() {
+        let line = match line {
+            Ok(x) => x,
+            Err(_) => return,
+        };
```
algesten, line 19: This could be written more succinctly as:

```rust
if line.is_err() { 
    return;
}
```

Or if we want to log or do something with just the error (or value):

```rust
if let Err(e) = line {
   ...
}
```

src/test/testserver.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+impl TestServer {
+    pub fn new(handler: fn(TcpStream) -> io::Result<()>) -> Self {
+        let listener = TcpListener::bind("localhost:0").unwrap();
+        let port = listener.local_addr().unwrap().port();
+        let done = Arc::new(AtomicBool::new(false));
+        let done_clone = done.clone();
+        let handle = thread::spawn(move || {
```
algesten, line 32: If you don't want the join handle, you can skip declaring it. Sometimes rust insist on a value being used, like `Result`. You can trigger the behaviour yourself for your own types. This is the rust source:

```rust
#[must_use = "this `Result` may be an `Err` variant, which should be handled"]
pub enum Result<T, E> {
    Ok(T),
    Err(E),
}
```

For those types you can ignore the return with:

```rust
let _ = thread::spawn(..`
```

src/test/testserver.rs
```diff
@@ -0,0 +1,54 @@
...
+            for stream in listener.incoming() {
+                thread::spawn(move || handler(stream.unwrap()));
+                if done.load(Ordering::Relaxed) {
+                    break;
+                }
+            }
+            println!("tesserver on {} exiting", port);
```
algesten, line 39: typo

src/test/testserver.rs
jsha: Good point! I just made this an unwrap, since we don't expect errors during the tests and want to panic if we get them.

algesten:
Merge away!

PR #99, opened 2020-06-30
Validate header field names.

### LINKED ISSUE #96: header: disallow white space between field name and colon
https://tools.ietf.org/html/rfc7230#section-3.2.4
>    No whitespace is allowed between the header field-name and colon.  In
   the past, differences in the handling of such whitespace have led to
   security vulnerabilities in request routing and response handling.  A
   server MUST reject any received request message that contains
   whitespace between a header field-name and colon with a response code
   of 400 (Bad Request).  A proxy MUST remove any such whitespace from a
   response message before forwarding the message downstream.

jsha:
This rejects spaces after field names as well as various other invalid
characters.

Fixes #96

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -103,13 +103,10 @@ fn valid_name(name: &str) -> bool {
 #[inline]
 fn is_tchar(b: u8) -> bool {
     match b {
-        b'!' | b'#' | b'$' | b'%' | b'&' | b'\'' | b'*' | b'+' | b'-' | b'.' | b'^' | b'_'
-        | b'`' | b'|' | b'~' | b'A' | b'B' | b'C' | b'D' | b'E' | b'F' | b'G' | b'H' | b'I'
-        | b'J' | b'K' | b'L' | b'M' | b'N' | b'O' | b'P' | b'Q' | b'R' | b'S' | b'T' | b'U'
-        | b'V' | b'W' | b'X' | b'Y' | b'Z' | b'a' | b'b' | b'c' | b'd' | b'e' | b'f' | b'g'
-        | b'h' | b'i' | b'j' | b'k' | b'l' | b'm' | b'n' | b'o' | b'p' | b'q' | b'r' | b's'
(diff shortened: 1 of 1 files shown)


src/header.rs
```diff
@@ -82,6 +88,32 @@ pub fn add_header(headers: &mut Vec<Header>, header: Header) {
...
+#[inline]
+fn is_tchar(b: u8) -> bool {
+    match b {
+        b'!' | b'#' | b'$' | b'%' | b'&' | b'\'' | b'*' | b'+' | b'-' | b'.' | b'^' | b'_'
+        | b'`' | b'|' | b'~' | b'A' | b'B' | b'C' | b'D' | b'E' | b'F' | b'G' | b'H' | b'I'
+        | b'J' | b'K' | b'L' | b'M' | b'N' | b'O' | b'P' | b'Q' | b'R' | b'S' | b'T' | b'U'
+        | b'V' | b'W' | b'X' | b'Y' | b'Z' | b'a' | b'b' | b'c' | b'd' | b'e' | b'f' | b'g'
+        | b'h' | b'i' | b'j' | b'k' | b'l' | b'm' | b'n' | b'o' | b'p' | b'q' | b'r' | b's'
+        | b't' | b'u' | b'v' | b'w' | b'x' | b'y' | b'z' | b'0' | b'1' | b'2' | b'3' | b'4'
+        | b'5' | b'6' | b'7' | b'8' | b'9' => true,
```
lolgesten, line 112: Maybe you prefer this for clarity (or speed?), but just pointing out that there are these in the std lib:

ALPHA => https://doc.rust-lang.org/beta/std/primitive.u8.html#method.is_ascii_alphabetic
DIGIT => https://doc.rust-lang.org/beta/std/primitive.u8.html#method.is_ascii_digit

or even the combo:

https://doc.rust-lang.org/beta/std/primitive.u8.html#method.is_ascii_alphanumeric

jsha: Ah, thanks for the reminder! I will admit, there was part of me that figured this might be a little speedier, but it's premature optimization; I prefer to use the clearer function and possibly optimize later if it turns out to be a bottleneck.

PR #104, opened 2020-07-01
Request: rename path field to URL.

jsha:
This more accurately reflects its use.


src/request.rs
```diff
@@ -523,7 +523,7 @@ impl Request {
     ///     .build();
     /// assert_eq!(req1.get_host().unwrap(), "cool.server");
     ///
-    /// let req2 = ureq::post("/some/path")
+    /// let req2 = ureq::post("/some/url")
```
algesten, line 526:
```suggestion
    /// let req2 = ureq::post("/some/path")
```

PR #116, opened 2020-07-13
Added method to set the TLS connection builder

edevil:
This allows configuring the accepted certs, client authentication,
etc, when using the native TLS crate.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -51,7 +51,7 @@ pub struct Request {
     #[cfg(feature = "tls")]
     pub(crate) tls_config: Option<TLSClientConfig>,
     #[cfg(all(feature = "native-tls", not(feature = "tls")))]
-    pub(crate) tls_builder: Option<TLSBuilder>,
+    pub(crate) tls_connector: Option<TLSConnector>,
 }
 
 impl ::std::fmt::Debug for Request {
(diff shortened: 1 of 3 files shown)


algesten:
@jsha my only concern would be to keep the functionality equivalent to rustls. I'm wondering whether this PR should use the built `TLSConnector` rather than `TLSBuilder`? Both implement `Clone`. I guess on some level the builder is more akin to `rustls::ClientConfig`. You got any opinion?

edevil:
I'd be happy to change the API if needed.

edevil:
Ping

algesten:
Hey! I'm off the grid this week. I'll be back on monday. Sorry!

jsha:
I'm happy to keep this rolling while you're off the grid. I agree with the thought that we should semantically parallel the existing code which uses `ClientConfig`. I'm not super familiar with native-tls, but I think you're right that TLSConnector seems closer than TLSBuilder. After all, why build a new object for each request if you don't have to? Are there any borrowing issues if using a single TLSConnector for multiple concurrent requests?

@edevil I'm guessing you're more familiar with native-tls. Do you have thoughts on the above questions? Also, I'm curious to hear what your use case is that requires native-tls over rustls. Do you need older TLS versions or ciphersuites, or locally-installed roots, or something else?

edevil:
For me it makes sense to build a new connector on every request, since that operation can fail. Previously we were also creating a new TLS connector for every request, using `new`, I'm guessing for the same reason.

The reason I'm using native TLS instead of rusttls is because webpki::DNSNameRef does not support connecting to URLs like:

https://[2606:4700:4700::1111]/

Webpki doesn't (yet) support validating a certificate for an IP address yet https://github.com/briansmith/webpki/issues/54

algesten:
@jsha 

> I'm happy to keep this rolling while you're off the grid.

Thanks, I'm back!

@edevil 

> For me it makes sense to build a new connector on every request, since that operation can fail.

My overriding concern is API congruency. ureq primary TLS impl is, for better or worse, rustls. Unless there are some additional arguments I haven't thought of, I think we should amend the PR to take `TLSConnector`.

> For me it makes sense to build a new connector on every request, since that operation can fail.

Presumably it can only fail if it's misconfigured somehow (i.e. incorrect cert name or similar). Which means that once it's built, it can't fail to clone it.

> Previously we were also creating a new TLS connector for every request, using `new`, I'm guessing for the same reason.

This is an internal detail, which we could happily amend to clone some default instance instead. We should think about the surface API first.

algesten:
> This is an internal detail, which we could happily amend to clone some default instance instead. We should think about the surface API first.

To be clear: We don't need to change anything here right now.

edevil:
I have amended the PR to take `TLSConnector`.

Issue #117, opened 2020-07-13
Add support for sending the request body via a Write impentation


aj-bagwell:
Some crates for exampe `tar` take a `Write` object to send their data to instead of creating a `Read` implemntation where data can be pulled from. It would be nice if there was a way to `write` the body of a request.

jsha:
Welcome to the project, and thanks for filing the issue. My first instinct is to try and avoid adding new APIs for a use case that isn't too onerous to solve in some other way. For this use case, does `pipe` do what you need to be able to work with the current `send` method? https://docs.rs/pipe/0.3.0/pipe/

Also, are there other crates that provide output via a `Write` rather than a `Read` and that people are likely to want to use as a request body?

aj-bagwell:
Thank you for your fast feedback and for giving me the push I needed to actually try `pipe`, I had written `pipe` off as I did not want to deal with the hasle of synconising mutiple threads as that was precicsely the kind of thing I was trying to avoid by using ureq instead of a full blown async client.

Using `pipe` is less ugy than I thought it would be, but getting good error reporting out of it is a pain as the errors are reported out of order and on a different thread, e.g. a typo in the host name turns up on the write thread as `pipe reader has been dropped` rather than the helpful `Dns Failed` message that could have been returned to the user before writing even started.

There are loads of crates that provide output via `Write` and not `Read` e.g. some more examples [zstd compression](https://docs.rs/zstd/0.5.3+zstd.1.4.5/zstd/stream/write/struct.Encoder.html), [JPEG encoding](https://docs.rs/image/0.23.7/image/jpeg/struct.JPEGEncoder.html), [CSV](https://docs.rs/csv/1.1.3/csv/struct.Writer.html) and many other interesting file formats. Many formats are just nicer to code to a `Write` interface than `Read` as read forces you to explicitly keep track framing in a resumeable manor instead of just writing it.

They may not be needed for the classic REST use cases but they are realy useful when trying to upload the result of some process to a server, especially if the result is too big to fit in memory. e.g. adding a screenshot to a Jira ticket, a process to extract a report from an DB and upload it to S3, or my case of uploading log files from an IoT device.

I very much apreciate the goal of keeping ureq simple but thought that this may be something that other people might find useful.

algesten:
We recommend pipe for this use case. https://crates.io/crates/pipe

Re-open a new issue if there are more issues with that.

aj-bagwell:
[...]

PR #118, opened 2020-07-13
Add Write support for sending request bodies

aj-bagwell:
Some crates for exampe `tar` take a `Write` object to send their data to instead of creating a `Read` implemntation where data can be polled from. This adds an `into_write` method to `Request` which returns a Writer that can be used to write the body of the request.

Implements feature request algesten/ureq#117


aj-bagwell:
I tried to make this as uninvasive as posible, but it might be cleaner to make the `Write` version the "main" implementation then use `std::io::copy` to build the read versions ontop.

algesten:
@aj-bagwell Thanks for submitting this. I'm on the fence about whether to merge. The surface API looks good and it's not an unreasonable feature to have. For cases where back pressure of the upstream server should slow down the sending side, it would be really great.

But it's somewhat of a big code change… hence I feel uneasy. I will think about it a bit more.

algesten:

PR #121, opened 2020-07-26
Added the possibility to disable certification checks

Roba1993:
Hi @algesten,

I want to replace my actual reqwest implementation for one of my libraries.
The library is used to call some rest information from a local smart home device with TSL encryption but no registered certification.

The reqwest crate allows to disable the certification checks for such cases.
I have implemented the same behavior here as a pull request together with a check to demonstrate it.

Maybe you are interested to merge this addition into ureq?

Thanks,
Robert

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ lazy_static = "1"
 qstring = "0.7"
 url = "2"
 socks = { version = "0.3.2", optional = true }
-rustls = { version = "0.18", optional = true, features = [] }
+rustls = { version = "0.18", optional = true, features = ["dangerous_configuration"] }
 webpki = { version = "0.21", optional = true }
 webpki-roots = { version = "0.20", optional = true }
 rustls-native-certs = { version = "0.4", optional = true }
--- a/src/request.rs
+++ b/src/request.rs
@@ -599,6 +599,34 @@ impl Request {
         self
     }
 
+    /// Creates new TLS client config, with disabled certification validation
+    ///
+    /// # Warning
+    ///
+    /// You should think very carefully before using this method. If
+    /// invalid certificates are trusted, *any* certificate for *any* site
+    /// will be trusted for use. This includes expired certificates. This
+    /// introduces significant vulnerabilities, and should only be used
+    /// as a last resort.
+    #[cfg(feature = "tls")]
+    pub fn danger_accept_invalid_certs(&mut self) -> &mut Request {
+        // New TLS client config, because the old one is not editable (stucked in Arc)
+        let mut tls_config = rustls::ClientConfig::new();
+
+        // Take a certificate verifier which is not checking the certificate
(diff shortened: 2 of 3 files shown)


jsha:
Thanks for the pull request! Right now ureq exposes options that allow you to set the rustls config yourself: `request.set_tls_config`. It sees that for risky features like this, it's acceptable to have clients build their own rustls config along with `set_certificate_verifier`. Are there reasons why that doesn't work for you? One advantage of that approach is that rather than funneling everyone straight into the "no verification" path, they get the opportunity to verify differently. For instance someone might have a problem with hostnames and might just want to verify an alternate hostname. But supporting that full complexity in ureq would get complex. :-)

Roba1993:
@jsha The main problem is, the implementation is quiete messy if you need to implement it yourself over `set_tls_config` because you need in addition the `rustls` and `wepkit` dependencies and make the implementation correctly.
Reqwest offers the same functionality, as you can see [here](https://docs.rs/reqwest/0.10.7/reqwest/struct.ClientBuilder.html#method.danger_accept_invalid_certs).

If you not agree I can also close the PR.

jsha:
>  the implementation is quiete messy if you need to implement it yourself over set_tls_config because you need in addition the rustls and wepkit dependencies and make the implementation correctly.

This is a good point, though in general I'd prefer asking the upstream rustls about making this use case simpler rather than implementing a simplification layer in downstream libraries.

Here's what I propose to do: Let's not merge this for now, but I'll take this as a vote for this sort of feature. If a lot of people turn out to need some common form of override for certificate verification, we can design a feature around their shared needs.

One other thing: Rather than disable certificate verification, maybe you'd like to call `config.root_store.add_server_trust_anchors(...)` on the root that signed your local device's TLS certificate. For many devices that will be a self-signed certificate, so you'd just put the specific certificate from your device in there (or make it a config parameter in case you want to reuse the code with a different device).

algesten:
@jsha 

>  Let's not merge this for now, but I'll take this as a vote for this sort of feature. If a lot of people turn out to need some common form of override for certificate verification.

I agree. Let's keep this PR around for people to weigh in, in case this is something more people want.

@Roba1993 

We're not merging this PR right now, but the impl is also not entirely complete. Since we support `native-tls`, if we are adding this function we would also need to do so for both TLS variants.

Roba1993:
Okay good for me, thanks

algesten:

CapitanShinChan:
Quick question here: Is there a simple way to disable the certificate validation? Or should I apply this patch to the crate in order to do so?

Thanks!

Roba1993:
I have no clue on disabling the certification check in a different way.

algesten:
@CapitanShinChan 

Have a look at AgebtBuilder. You can set a Rustls ClientConfig. There you have some DangerousClientConfig that lets you handle certificate validation.

PR #123, opened 2020-07-30
Make Response::to_json preserve io::Error of ErrorKind::TimedOut

algesten:


algesten:
@jsha Does this look reasonable to you? serde_json doesn't let us get the actual underlying struct, so to "preserve" the io::Error, I clone to be as good as possible.

I'll add a test case if you think it looks ok.

src/request.rs
```diff
@@ -613,7 +613,10 @@ impl Request {
     ///     .set_tls_connector(tls_connector.clone());
     /// ```
     #[cfg(all(feature = "native-tls", not(feature = "tls")))]
-    pub fn set_tls_connector(&mut self, tls_connector: Arc<native_tls::TlsConnector>) -> &mut Request {
+    pub fn set_tls_connector(
```
jsha, line 616: This looks like an unrelated formatting change. Usually I try to put formatting changes in their own commit so they're easier to understand in a `git blame`.

algesten:
@jsha Making the test case I realised we needed to cover the `WouldBlock` scenario too. I added some more commits around that. Notably I ended up blanket cloning all `io::Error`. Looks ok?

algesten:
Ok! I'll make another PR to land first for that.

algesten:
@jsha I believe this one is ready to go in now (take 2).

PR #130, opened 2020-08-05
Make Cookie header RFC6265 compliant

### LINKED ISSUE #129: Cookie implementation uses one Header per Cookie instead of one Header for all Cookies
RFC 6265 (https://tools.ietf.org/html/rfc6265#section-4.2) defines setting cookies in a request as:

4.2.  Cookie

4.2.1.  Syntax

   The user agent sends stored cookies to the origin server in the
   Cookie header.  If the server conforms to the requirements in
   Section 4.1 (and the user agent conforms to the requirements in
   Section 5), the user agent will send a Cookie header that conforms to
   the following grammar:

   cookie-header = "Cookie:" OWS cookie-string OWS
   cookie-string = cookie-pair *( ";" SP cookie-pair )

Note that this is defined as one Cookie header separated by ';'.

cazgp:
Fixes #129

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -61,7 +61,7 @@ impl Unit {
 
         let query_string = combine_query(&url, &req.query, mix_queries);
 
-        let cookie_headers: Vec<_> = extract_cookies(&req.agent, &url);
+        let cookie_header: Option<Header> = extract_cookies(&req.agent, &url);
 
         let extra_headers = {
             let mut extra = vec![];
@@ -86,7 +86,7 @@ impl Unit {
         let headers: Vec<_> = req
             .headers
             .iter()
-            .chain(cookie_headers.iter())
+            .chain(cookie_header.iter())
             .chain(extra_headers.iter())
             .cloned()
             .collect();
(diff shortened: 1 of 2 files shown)


src/unit.rs
```diff
@@ -281,9 +282,12 @@ fn match_cookies(jar: &CookieJar, domain: &str, path: &str, is_secure: bool) ->
...
             let value = c.value().to_string();
             let nameval = Cookie::new(name, value).encoded().to_string();
-            Header::new("Cookie", &nameval)
+            nameval
         })
-        .collect()
+        .collect::<Vec<_>>()
+        .join(";");
+
+    Header::new("Cookie", &cookie)
```
cazgp, line 290: I don't *think* that setting an empty cookie header is a bad thing but it may also not be desired?

jsha: I agree, we should avoid sending the cookie header if there are no cookies. How about making this function return an `Option` too?

cazgp: Done :)

cazgp:
@jsha all done and two tests added!

PR #131, opened 2020-08-06
Normalize timeout ErrorKind::WouldBlock to be TimedOut

algesten:
`set_read_timeout` and `set_write_timeout` can cause `ErrorKind::WouldBlock` on unix-y platforms.

This PR normalize those cases to `ErrorKind::TimedOut`. This will make it simpler higher up in the stack to deal with timeouts.

This PR should land before #123


algesten:
@jsha This is what we talked about I think.

PR #132, opened 2020-08-09
Switch to Result-based API.

### LINKED ISSUE #128: Proposal: Result-based API
Right now, ureq handles errors by `.synthetic_error()` (for protocol-level errors, like connection refused), and `.error()`, (for protocol-level errors *and* HTTP-semantic errors like 404 and 500).

It would be very convenient to use Results for errors, in particular because it allows using the `?` sugar to trivially propagate errors.

I think we can offer a new API without breaking backwards compatibility (for now). We should define a new method on Request like so:

```
fn do(&self) -> Result<Response, Error>
```

By default, this method would return Error only for protocol-level errors; for 4xx and 5xx, the user would be responsible for checking the status code and reading the body. However, for convenience we could add another method on Request:

```
fn error_on_status(&mut self) -> &mut Request
```

When the user sets this option for a request, 4xx and 5xx errors have their bodies read automatically and get turned into `Error`s.

One other change needs to happen as part of this: send_bytes, send_string, send_form, and send_json currently return Response. To provide equivalents in the Result-based API, I propose we define `body_bytes`, `body_string`, `body_form`, and `body_json`, which set the body of a Request object to one of those things, so the body gets sent as part of `.do()`. Example:

```
  let resp = ureq::post("https://myapi.example.com/ingest")
      .set("X-My-Header", "Secret")
      .error_on_status()
      .body_json(json!({
          "name": "martin",
          "rust": true
      }))
      .do()?;
  println!("{}", resp.into_string()?);
```

Related: #94 #125

Thoughts? @BurntSushi @algesten

jsha:
Gets rid of synthetic_error, and makes the various send_* methods return `Result<Response, Error>`.
Introduces a new error type "HTTP", which represents an error due to status codes 4xx or 5xx.
The HTTP error type contains a boxed Response, so users can read the actual response if they want.
Adds an `error_for_status` setting to disable the functionality of treating 4xx and 5xx as errors.
Adds .unwrap() to a lot of tests.

Fixes #128.


jsha:
Alright, I think this is ready to take a look at. If this is the path we take, I don't want to merge it right away. I'd like to cut a release with the fixes we've got since 1.3.0, and solicit feedback from users.

One thing I ran into implementing examples error_for_status: Because the only ways to read the body consume the Response, it's not possible to consume the body unconditionally and then check for errors afterwards, which would be necessary to return connections to the pool. I.e., I wanted to write this:

```
fn main() -> Result<(), ureq::error::Error> {
  let resp = ureq::get("http://example.com/").call()?;
  let body = resp.into_string()?;
  resp.error_for_status()?;
}
```

But `resp` is moved in the `let body` line.

If we're looking at a 2.0 we might want to consider changing the semantics of `.into_*` so they take the stream out of the Response rather than taking the whole Response.

src/lib.rs
```diff
@@ -235,8 +219,8 @@ mod tests {
     #[test]
     #[cfg(any(feature = "tls", feature = "native-tls"))]
     fn connect_https_invalid_name() {
-        let resp = get("https://example.com{REQUEST_URI}/").call();
-        assert_eq!(400, resp.status());
-        assert!(resp.synthetic());
+        let result = get("https://example.com{REQUEST_URI}/").call();
+        let err = result.err().unwrap();
```
algesten, line 223: This can be done as `result.unwrap_err()`

src/request.rs
```diff
@@ -22,6 +22,8 @@ use crate::Response;
 #[cfg(feature = "json")]
 use super::SerdeValue;
 
+type Result<T> = std::result::Result<T, Error>;
```
algesten, line 25: Maybe this type should be public since we return it everywhere?

src/response.rs
```diff
@@ -176,48 +165,31 @@ impl Response {
...
-    pub fn synthetic(&self) -> bool {
-        self.error.is_some()
-    }
-
-    /// Get the actual underlying error when the response is
-    /// ["synthetic"](struct.Response.html#method.synthetic).
-    pub fn synthetic_error(&self) -> &Option<Error> {
-        &self.error
+    pub fn error_for_status(self) -> Result<Response, Error> {
+        if self.status > 400 {
```
algesten, line 188: Should probably be `>=`

jsha:
I pushed another commit (0cd8a2f) with a possible direction for making 4xx and 5xx return Errors by default, with the ability to turn that off at the Request level.

I made the HTTP error type include a full Response object, so clients can handle different types of errors differently; alternately if a client wants to process errors at the same level as success codes, they can turn off return_error_for_status.

I also wound up making 3xx return an Error when redirects(0) is set. This seemed more consistent, and clients can retrieve the actual redirect response via the boxed Response just as they can for 4xx and 5xx.

Curious what you think!

jyn514:
>  Thanks for the report, it's really useful to hear from folks with this problem. Would you mind checking out #132 and building against it to see if it would solve your problem? We're making changes to the API for a 2.0.0 release to solve this and other problems, so we'd like to make sure it provides a solution before finalizing the API.

This worked really well for me. The switch from the old to new API was pretty simple:
```diff
diff --git a/src/check.rs b/src/check.rs
index a1e0f04..4012dd7 100644
--- a/src/check.rs
+++ b/src/check.rs
@@ -90,9 +90,8 @@ fn check_http_url(url: &Url, ctx: &CheckContext) -> Result<(), CheckError> {
         }
     }
 
-    let resp = ureq::get(url.as_str()).call();
-    match resp.synthetic_error() {
-        None => {
+    match ureq::get(url.as_str()).call() {
+        Ok(resp) => {
             if resp.ok() {
                 Ok(())
             } else {
@@ -102,6 +101,6 @@ fn check_http_url(url: &Url, ctx: &CheckContext) -> Result<(), CheckError> {
                 )))
             }
         }
-        Some(e) => Err(CheckError::Http(HttpError::Fetch(url.clone(), (*e).clone()))),
+        Err(e) => Err(CheckError::Http(HttpError::Fetch(url.clone(), e)))
     }
 }
```
and unlike the old one, it actually compiles 😆. Would love to have this in a published release!

src/error.rs
```diff
@@ -112,7 +56,7 @@ impl From<io::Error> for Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.body_text())
+        write!(f, "{:?}", self)
```
jyn514, line 59: Hmm, are you sure this shouldn't instead print the old `body_text()` for synthetic errors?

src/lib.rs
```diff
@@ -24,13 +24,11 @@
 //!           "rust": true
 //!       }));
 //!
-//!   // .ok() tells if response is 200-299.
-//!   if resp.ok() {
-//!     println!("success: {}", resp.into_string()?);
+//!   if resp.is_ok() {
+//!     println!("success: {}", resp.unwrap().into_string()?);
```
jyn514, line 28:
```suggestion
//!   if let Ok(resp) = resp {
//!     println!("success: {}", resp.into_string()?);
```

src/response.rs
```diff
@@ -85,15 +76,13 @@ impl Response {
...
     ///
     /// assert_eq!(resp.status(), 401);
     /// ```
-    pub fn new(status: u16, status_text: &str, body: &str) -> Self {
+    pub fn new(status: u16, status_text: &str, body: &str) -> Result<Response, Error> {
         let r = format!("HTTP/1.1 {} {}\r\n\r\n{}\n", status, status_text, body);
-        (r.as_ref() as &str)
-            .parse::<Response>()
-            .unwrap_or_else(|e| e.into())
+        (r.as_ref() as &str).parse::<Response>()
```
jyn514, line 85: You don't need the type annotations here I think.
```suggestion
        (r.as_ref() as &str).parse()
```

src/stream.rs
```diff
@@ -416,6 +416,7 @@ pub(crate) fn connect_host(unit: &Unit, hostname: &str, port: u16) -> Result<Tcp
             None => None,
         };
 
+        println!("connecting to {}", &sock_addr);
```
jyn514, line 419: This looks like debugging - maybe use `log::debug!` instead? (or just remove it altogether if you don't want to introduce a log dependency.)

jyn514 (commented):
I found myself with too much time on my hands; feel free to ignore any or all of these comments 😆

src/stream.rs
jsha: We actually just added `log` as a dependency!

src/error.rs
jsha: Good call!

algesten:
Merge away!

src/response.rs
algesten: Ah yes, let's try without!

PR #133, opened 2020-08-09
Allow saving multiple idle streams per host.

### LINKED ISSUE #110: Allow saving multiple streams with the same address to the pool
It would be useful to be able to save multiple streams to the same address in the connection pool - for example, when doing chunked downloading via Range header, multiple connections can be used and reused.

cfal:
Fixes #110 

Please take a look, thanks!

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -9,8 +9,8 @@ use crate::Proxy;
 use url::Url;
 
 pub const DEFAULT_HOST: &str = "localhost";
-const MAX_IDLE_CONNECTIONS: usize = 100;
-const MAX_IDLE_CONNECTIONS_PER_HOST: usize = 1;
+const DEFAULT_MAX_IDLE_CONNECTIONS: usize = 100;
+const DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST: usize = 1;
 
 /// Holder of recycled connections.
 ///
@@ -28,12 +28,62 @@ pub(crate) struct ConnectionPool {
     // recently used Streams are added to the back of the queue;
     // old streams are removed from the front.
     lru: VecDeque<PoolKey>,
+    max_idle_connections: usize,
+    max_idle_connections_per_host: usize,
 }
 
 impl ConnectionPool {
     pub fn new() -> Self {
         ConnectionPool {
+            max_idle_connections: DEFAULT_MAX_IDLE_CONNECTIONS,
+            max_idle_connections_per_host: DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST,
             ..Default::default()
+        }
+    }
+
+    pub fn set_max_idle_connections(&mut self, max_connections: usize) {
+        if self.max_idle_connections == max_connections {
+            return;
+        }
+        self.max_idle_connections = max_connections;
+
+        if max_connections == 0 {
+            // Clear the connection pool, caching is disabled.
+            self.lru.clear();
+            self.recycle.clear();
+            return;
+        }
+
+        // Remove any extra connections if the number was decreased.
+        while self.lru.len() > max_connections {
+            self.remove_oldest();
+        }
+    }
+
+    pub fn set_max_idle_connections_per_host(&mut self, max_connections: usize) {
+        if self.max_idle_connections_per_host == max_connections {
+            return;
+        }
+        self.max_idle_connections_per_host = max_connections;
+
+        if max_connections == 0 {
+            // Clear the connection pool, caching is disabled.
+            self.lru.clear();
+            self.recycle.clear();
+            return;
+        }
+
+        // Remove any extra streams if the number was decreased.
+        for (key, val) in self.recycle.iter_mut() {
+            while val.len() > max_connections {
+                val.pop_front();
+                let index = self
+                    .lru
+                    .iter()
+                    .position(|x| x == key)
+                    .expect("PoolKey not found in lru");
+                self.lru.remove(index);
+            }
         }
     }
 
@@ -80,11 +130,15 @@ impl ConnectionPool {
     }
 
     fn add(&mut self, key: PoolKey, stream: Stream) {
+        if self.max_idle_connections == 0 || self.max_idle_connections_per_host == 0 {
+            return;
+        }
+
         match self.recycle.entry(key.clone()) {
             Entry::Occupied(mut occupied_entry) => {
                 let streams = occupied_entry.get_mut();
                 streams.push_back(stream);
-                if streams.len() > MAX_IDLE_CONNECTIONS_PER_HOST {
+                if streams.len() > self.max_idle_connections_per_host {
                     streams.pop_front();
                     self.remove_from_lru(&key);
                 }
@@ -96,7 +150,7 @@ impl ConnectionPool {
             }
         }
         self.lru.push_back(key);
-        if self.lru.len() > MAX_IDLE_CONNECTIONS {
+        if self.lru.len() > self.max_idle_connections {
             self.remove_oldest()
         }
     }
@@ -175,7 +229,7 @@ fn pool_connections_limit() {
     // filling and draining it. The pool should evict earlier connections
     // when the connection limit is reached.
     let mut pool = ConnectionPool::new();
-    let hostnames = (0..MAX_IDLE_CONNECTIONS * 2).map(|i| format!("{}.example", i));
+    let hostnames = (0..DEFAULT_MAX_IDLE_CONNECTIONS * 2).map(|i| format!("{}.example", i));
     let poolkeys = hostnames.map(|hostname| PoolKey {
         scheme: "https".to_string(),
         hostname,
@@ -185,9 +239,9 @@ fn pool_connections_limit() {
     for key in poolkeys.clone() {
         pool.add(key, Stream::Cursor(std::io::Cursor::new(vec![])));
     }
-    assert_eq!(pool.len(), MAX_IDLE_CONNECTIONS);
+    assert_eq!(pool.len(), DEFAULT_MAX_IDLE_CONNECTIONS);
 
-    for key in poolkeys.skip(MAX_IDLE_CONNECTIONS) {
+    for key in poolkeys.skip(DEFAULT_MAX_IDLE_CONNECTIONS) {
         let result = pool.remove(&key);
         assert!(result.is_some(), "expected key was not in pool");
     }
@@ -207,15 +261,15 @@ fn pool_per_host_connections_limit() {
         proxy: None,
     };
 
-    for _ in 0..MAX_IDLE_CONNECTIONS_PER_HOST * 2 {
+    for _ in 0..pool.max_idle_connections_per_host * 2 {
         pool.add(
             poolkey.clone(),
             Stream::Cursor(std::io::Cursor::new(vec![])),
         );
     }
-    assert_eq!(pool.len(), MAX_IDLE_CONNECTIONS_PER_HOST);
+    assert_eq!(pool.len(), DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST);
 
-    for _ in 0..MAX_IDLE_CONNECTIONS_PER_HOST {
+    for _ in 0..DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST {
         let result = pool.remove(&poolkey);
         assert!(result.is_some(), "expected key was not in pool");
     }
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -277,6 +277,50 @@ fn pool_per_host_connections_limit() {
 }
 
 #[test]
+fn pool_update_connection_limit() {
+    let mut pool = ConnectionPool::new();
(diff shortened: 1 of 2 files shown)


src/pool.rs
```diff
@@ -136,38 +214,43 @@ fn pool_size_limit() {
...
+        let result = pool.remove(key);
         assert!(result.is_some(), "expected key was not in pool");
     }
 }
 
 #[test]
-fn pool_duplicates_limit() {
-    // Test inserting duplicates into the pool, and subsequently
+fn pool_connections_limit() {
+    // Test inserting multiple connections into the pool with the same key,
```
algesten, line 226: strange english here

src/pool.rs
```diff
@@ -20,52 +22,129 @@ const MAX_IDLE_CONNECTIONS: usize = 100;
...
+                let streams = occupied_entry.get_mut();
+                let stream = streams.pop_front();
+                assert!(
+                    stream.is_some(),
+                    "key existed in recycle but no streams available"
+                );
+                if streams.len() == 0 {
+                    occupied_entry.remove_entry();
+                    self.lru.retain(|x| x != &key);
+                }
```
algesten, line 107: Wonder if the `key` should be shifted to the back of `lru` here? If the `stream` we use failed, it won't be reinserted in the pool, and I wonder if the order should be changed anyway?

cfal: Sorry, I don't quite understand! `remove_entry` removes the key/now empty VecDeque from `self.recycle` when the last stream has been removed, and then `retain` removes `key` from `lru` as well. Do you mean keep the key in `lru` even when there are no streams of that key remaining in `recycle`? If the used `stream` fails, then I guess for the same `key` to be reinserted, it would mean the user created a new connection, or another in-use connection was returned. In both cases, `add` also moves the key to the back of `lru` at the moment.

cfal:
Thanks for the quick reviews!

> Awesome change, thanks for making it! I have one piece of feedback on the API - sorry I didn't provide this earlier when you filed the issue. Rather than max_idle_hosts, I think it makes sense to configure max_idle_connections (and keep max_idle_connections_per_host). You can express the same range of policies - e.g. 100 connections to a single host, or 100 connections to different hosts. But since a connection is the fundamental unit of resource usage (file descriptors, memory, TCP sockets), I think that's what people care about the most, so it's nice to be able to configure that directly (and allow tweaking idle_connections_per_host for more sophisticated needs).

That makes sense to me, but would require a bit more book-keeping.
I guess something like this would work where we go back to tracking connections instead of hosts in `lru`:

- on adding a connection to the pool, insert the pool key to lru. If the length of the lru queue is > than the connection limit, pop the first key, lookup the key in self.recycle and get the queue of streams, and pop the first stream in that queue.
- on try_get_connection(), lookup the key in self.recycle, pop the first stream to be returned. iterate through self.lru and find the first matching pool key, and remove it.

How does that sound?

jsha:
> That makes sense to me, but would require a bit more book-keeping. We could change self.recycle from a HashMap to a Vec, and iterate through looking for a matching stream on try_get_connection().

I think it should be possible to do it without iterating over a whole Vec. I'm thinking of Go's net/http here:

https://github.com/golang/go/blob/ba9e10889976025ee1d027db6b1cad383ec56de8/src/net/http/transport.go#L2811-L2848

Its LRU consists of both a list, and a map from addresses of specific `persistConn`'s to their position in the list, for quick removal (a persistConn represents a stream plus some extra stuff). Using raw addresses isn't an option for us, but we could do something similar by wrapping Stream in some type that has an ID field, and using the ID to find the elements in the LRU that need removing when we pull an entry from the `recycle` map.

This could get a bit complex and I'm not sure I'm explaining it very well right now - feel free to ask followup questions!

cfal:
> > That makes sense to me, but would require a bit more book-keeping. We could change self.recycle from a HashMap to a Vec, and iterate through looking for a matching stream on try_get_connection().
> 
> I think it should be possible to do it without iterating over a whole Vec. I'm thinking of Go's net/http here:
> 
> https://github.com/golang/go/blob/ba9e10889976025ee1d027db6b1cad383ec56de8/src/net/http/transport.go#L2811-L2848
> 
> Its LRU consists of both a list, and a map from addresses of specific `persistConn`'s to their position in the list, for quick removal (a persistConn represents a stream plus some extra stuff). Using raw addresses isn't an option for us, but we could do something similar by wrapping Stream in some type that has an ID field, and using the ID to find the elements in the LRU that need removing when we pull an entry from the `recycle` map.
> 
> This could get a bit complex and I'm not sure I'm explaining it very well right now - feel free to ask followup questions!

With the approach above, when a `try_get_connection()` happens, wouldn't you still have to iterate through the LRU list in order to find the PoolKey (or some PoolKey+ID) to remove?

src/pool.rs
algesten: What I mean is in the case where `stream.len() > 0`. I.e. we have two connections for some host, we take one out and one remains – maybe we should shift the position of the remaining one?

jsha:
I spent a bunch of time tonight thinking about how to do this. Traditionally an LRU is implemented as a linked list + a hash map, with cross references. That gives you O(1) lookup by name, O(1) removal by name, and O(1) removal by oldest.

Unfortunately, std::collections::LinkedList doesn't have the primitives we need: We need to be able to store a reference to an item in the list. I think LinkedList's Cursor might give us what we need, but it is nightly-only right now.

There are a couple of crates that implement LRU's, but (a) they don't support multiple elements for the same key, which we need, and (b) they use a heaping spoonful of `unsafe`. The latter might be okay, but if we can get away without it, more's the better.

I'd still like to get the API right, even if it comes at the cost of a little performance today. Probably we can come up with some improvement down the road. Also, I just realized - we already take an O(n) performance hit iterating through the LRU for an item to remove anyhow. So I definitely don't want to ask you to make the performance even better than it is just in order to do this feature!

Here's a proposal:

```
type AgedStream = (Instant, Stream); // Associated with the time of last use
...
recycle: HashMap<PoolKey, VecDeque<AgedStream>>,
lru: VecDeque<(PoolKey, Instant)>,
```

The VecDeque in `recycle` would be maintained in sorted order, oldest to freshest.

To fetch a stream for reuse, we look in the HashMap. If there's an entry, we pop off the freshest stream. We then iterate through VecDeque to find an instance of the same PoolKey with the same Instant, and delete it.

To remove the oldest stream, we pop off the front of `lru`, then look up the corresponding entry in the HashMap and remove it. It should always be the frontmost element of the VecDeque in the HashMap.

I think this approach winds up equivalent in performance to what we currently have. The most expensive operation is reusing a stream that is about halfway through the LRU on its way to expiring off. The more heavily our workload is slanted towards reusing recent streams, the more efficient it is.

What do you think?

src/pool.rs
cfal: Ah, got it. That makes sense. Perhaps it should be shifted to the end of the `lru` since the host becomes the most recently used.

cfal:
Thanks for taking the time.
 
> Here's a proposal:
> 
> ```
> type AgedStream = (Instant, Stream); // Associated with the time of last use
> ...
> recycle: HashMap<PoolKey, VecDeque<AgedStream>>,
> lru: VecDeque<(PoolKey, Instant)>,
> ```
> 
> The VecDeque in `recycle` would be maintained in sorted order, oldest to freshest.
> 
> To fetch a stream for reuse, we look in the HashMap. If there's an entry, we pop off the freshest stream. We then iterate through VecDeque to find an instance of the same PoolKey with the same Instant, and delete it.

Do we need to check or include `Instant`? The `VecDeque` being iterated, I assume is `lru`. If it's the freshest stream, wouldn't it always correspond to the last matching `PoolKey` in `lru`?

> To remove the oldest stream, we pop off the front of `lru`, then look up the corresponding entry in the HashMap and remove it. It should always be the frontmost element of the VecDeque in the HashMap.
> I think this approach winds up equivalent in performance to what we currently have. The most expensive operation is reusing a stream that is about halfway through the LRU on its way to expiring off. The more heavily our workload is slanted towards reusing recent streams, the more efficient it is.

I'm uncertain if reusing the most recent stream is better than reusing the oldest stream, or an approach somewhere in the middle (eg. set some "max idle" time for a stream to be re-usable, and use the oldest one that hasn't reached that time, and drop the older ones). But that might be more complex than necessary.

jsha:
> Do we need to check or include Instant? The VecDeque being iterated, I assume is lru. If it's the freshest stream, wouldn't it always correspond to the last matching PoolKey in lru?

I think you're right.

We can phrase part of this as an invariant: For every entry in `recycle`, if it has N elements, there are also N elements of the same PoolKey in LRU.

And part of it as semantics: For any given PoolKey `p`, the Nth entry in lru matching `p` represents the Nth entry in `recycle[p]`'s VecDeque.

Also: The Nth entry in lru always represents an older stream than the N+1th entry.

> I'm uncertain if reusing the most recent stream is better than reusing the oldest stream, or an approach somewhere in the middle

Good question.

Pros for freshest: Most likely to still be alive; if we open a lot of connections to one host during a spike and then settle down to less frequent invocations, allows old connections to idle out.

Pros for oldest: Servers commonly kill idle connections; by preferring to reuse idle connections we might maximize the number of connections we keep alive.

Con for oldest: We're more likely to send a request on a connection that is dead but where the OS doesn't yet know it's dead (e.g. a firewall idled it out with sending an RST).

I don't have a conclusive answer, but I lean towards freshest. It looks like that's what Go's `net/http` does too: https://github.com/golang/go/blob/go1.15/src/net/http/transport.go#L976-L1004

src/pool.rs
algesten: Yeah. I think so.

src/pool.rs
cfal: Updated.

src/pool.rs
cfal: Please take another look. Because of the changes jsha suggested, I had to update this logic.

cfal:
Thanks for all the feedback!

I've updated the commit now to once again have a max connections limit and to take the freshest connection. I made a decision about removing the oldest matching key in the LRU when a connection gets used, let me know if that makes sense.

cfal:
@algesten @jsha gentle ping

algesten:
@cfal sorry I wasn't on top of this. You and @jsha did most of the braining on this PR, I think we're good to merge, but if possible I'd like @jsha to give the final thumbs up.

cfal:
@jsha do you mind taking another look? :)

jsha (approved):
Sorry I left you hanging for so long, @cfal! I hadn't seen your messages until just now. Normally I rely on GitHub's email notifications to see when there's an update, but for some reason none of the replies after my last comment came through to email, even though other notifications from this repo did come through, and GitHub says I'm subscribed to this PR. :thinking: 

This PR is looking really nice. There are some things I'd like to tweak, which I'm happy to do as a follow-on PR, or which you're welcome to do if you'd like.

When add() removes a stream from the pool, it's pulling the freshest stream from recycle (the one at the back), so it should remove the freshest matching entry from the LRU. Conversely, when remove_oldest removes a stream from the pool, it's removing the oldest one, so it should remove the oldest matching entry from the LRU. Right now both of those cases call remove_from_lru, which removes the oldest.

We have new invariants (there are always the same number of entries for a given PoolKey in the LRU and in recycle's VeqDeque; the Streams in recycle's VeqDeque are ordered from least recent to most recent use). We should document those. Also the remove_from_lru function can be run while some of the invariants don't hold; we should document that.

In remove_oldest, we now have an implicit assumption that max_idle_* is > 0, because remove_oldest won't be called otherwise. If that assumption doesn't hold, we'll hit the panic!() that occurs when there's nothing to pop. We should make the assumptions explicit with an assert!() at the top of remove_oldest.

We've got some deep nesting starting to happen, which hurts readability. In particular, while I like the match ... entry() style, it does seem to lead to deep nesting. I'd like to play around with different ways to reduce how deep in if statements and matches people need to read.

Similarly, the way you've replaced some asserts with expects when we need to use the value is a sweet improvement - but rustfmt makes it awkward by breaking each (short) method call onto its own line. I think we can keep things short by doing:

```
let index = self.lru.iter().position(|k| k == key);
let index = index.expect("invariant failed: key existed in recycle but not lru");
```

cfal:
Thanks for getting back to me and sorry for my own delayed response! And also thanks for fixing the `set_max_idle_connections(0)` bug :)

> When add() removes a stream from the pool, it's pulling the freshest stream from recycle (the one at the back), so it should remove the freshest matching entry from the LRU. Conversely, when remove_oldest removes a stream from the pool, it's removing the oldest one, so it should remove the oldest matching entry from the LRU. Right now both of those cases call remove_from_lru, which removes the oldest.

It's been a little while since I wrote this code, but this made sense to me at the time. Because `add()` was called, this means that a stream with that key was in demand and could now be considered "most recently used".  It made sense to remove the oldest entry, because it could mean that the user would be in need of more streams with this key, so we don't want to evict other matching streams from the cache early.

I'm also OK with your interpretation though. Cheers!

PR #137, opened 2020-08-31
Clean up unused code and long imports.

jsha:
This removes some commented out methods, and also changes instances of
::std::foo to use a more idiomatic import path.


algesten (approved):
Neat! ✨

PR #138, opened 2020-08-31
Put the crate version in the User-Agent header.

jsha:
https://tools.ietf.org/html/rfc7231#section-5.5.3
Each product identifier consists of a name and optional version.

     product         = token ["/" product-version]
     product-version = token

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -369,7 +369,7 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> IoResult<()> {
         }
     }
     if !header::has_header(&unit.headers, "user-agent") {
-        write!(prelude, "User-Agent: ureq/{}\r\n", env!("CARGO_PKG_VERSION"))?;
+        write!(prelude, "User-Agent: rust-ureq/{}\r\n", env!("CARGO_PKG_VERSION"))?;
     }
     if !header::has_header(&unit.headers, "accept") {
         write!(prelude, "Accept: */*\r\n")?;
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -369,7 +369,7 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> IoResult<()> {
         }
     }
     if !header::has_header(&unit.headers, "user-agent") {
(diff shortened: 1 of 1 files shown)


algesten (approved):
If we're changing this, maybe do it on the form `rust/ureq/1.2.3`. Just to make it clear that we're not talking about this [c lib](https://github.com/solusipse/ureq)

It's also the form I chose for [hreq](https://github.com/algesten/hreq/blob/master/src/lib.rs#L379) – for reasons that now elude me, but I'm quite certain I researched it.

jsha:
I think the extra / in the "product" name is not technically valid. `token` is defined at https://tools.ietf.org/html/rfc7230#section-3.2.6:

```
     token          = 1*tchar
     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```

How about rust-ureq?

src/unit.rs
```diff
@@ -369,7 +369,7 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> IoResult<()> {
         }
     }
     if !header::has_header(&unit.headers, "user-agent") {
-        write!(prelude, "User-Agent: ureq\r\n")?;
+        write!(prelude, "User-Agent: rust-ureq/{}\r\n", env!("CARGO_PKG_VERSION"))?;
```
algesten, line 372:
```suggestion
        write!(prelude, "User-Agent: ureq/{}\r\n", env!("CARGO_PKG_VERSION"))?;
```

algesten: I'm in awe of your detail knowledge :)

Let's drop the `rust` bit completely.

https://github.com/algesten/hreq/blob/master/src/lib.rs#L388

algesten (approved):

src/unit.rs
jsha: Thank you! It's not so much detailed knowledge as "that looks funny; I wonder why I've never seen a UA like that?" And then going and Ctrl-F'ing in the spec. :-)

PR #142, opened 2020-09-13
Document and tidy up new Pool invariants.

jsha:
Now that we allow multiple connections for the same PoolKey, update our
invariants accordingly. Also provide a couple of helper functions for
removing the first or last match of an entry in a VecDeque.

This also changes which entry from `lru` gets removed when a stream is
removed from the pool. Previously it was the oldest matching one. Now it's
the newest matching one, which matches the semantics we are applying to
`recycle[K]`.

Followup to #133 

/cc @cfal

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -165,9 +165,7 @@ impl ConnectionPool {
                 }
             }
             Entry::Vacant(vacant_entry) => {
-                let mut new_deque = VecDeque::new();
-                new_deque.push_back(stream);
-                vacant_entry.insert(new_deque);
+                vacant_entry.insert(vec![stream].into());
             }
         }
         self.lru.push_back(key);
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -30,7 +30,7 @@ const DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST: usize = 1;
 /// These invariants hold at the start and end of each method:
 ///  - The length `lru` is equal to the sum of lengths of `recycle`'s VecDeques.
 ///  - Each PoolKey exists the same number of times in `lru` as it has entries in `recycle`.
-///  - If there is an entry in `recycle`, the VecDeque has at least one element.
+///  - If there is an entry in `recycle`, it has at least one element.
 ///  - The length of `lru` is less than or equal to max_idle_connections.
 ///  - The length of recycle[K] is less than or equal to max_idle_connections_per_host.
 ///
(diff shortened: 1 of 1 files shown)


src/pool.rs
```diff
@@ -14,9 +14,25 @@ const DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST: usize = 1;
...
+/// a key K represents the first entry in `recycle[K]`. The second entry in lru
+/// for `K` represents the second entry in `recycle[K]`, and so on. In other
+/// words, `lru` is ordered the same way as the VecDeque entries in `recycle`:
+/// oldest at the front, freshest at the back. This allows keeping track of which
+/// host should have its connection dropped next.
+///
+/// These invariants hold at the start and end of each method:
+///  - The length `lru` is equal to the sum of lengths of `recycle`'s VecDeques.
+///  - Each PoolKey exists the same number of times in `lru` as it has entries in `recycle`.
+///  - If there is an entry in `recycle`, the VecDeque has at least one element.
```
algesten, line 33: > the VecDeque has at least one element

Since both are `VecDeque`, this is a bit ambiguous. Maybe call it `lru` here?

src/pool.rs
```diff
@@ -32,6 +48,20 @@ pub(crate) struct ConnectionPool {
     max_idle_connections_per_host: usize,
 }
 
+fn remove_first_match(list: &mut VecDeque<PoolKey>, key: &PoolKey) -> Option<PoolKey> {
+    match list.iter_mut().position(|x| x == key) {
```
algesten, line 52: This can be an `iter()` since you're not mutating using the iterator.

src/pool.rs
```diff
@@ -32,6 +48,20 @@ pub(crate) struct ConnectionPool {
...
+        Some(i) => list.remove(i),
+        None => None,
+    }
+}
+
+fn remove_last_match(list: &mut VecDeque<PoolKey>, key: &PoolKey) -> Option<PoolKey> {
+    match list.iter_mut().rposition(|x| x == key) {
```
algesten, line 59: ditto

algesten (approved):
Looks good! Tidy! ✨ 

A couple of minor things.

src/pool.rs
jsha: Good call. It turns out this was confusing, since I actually meant to refer to the VecDeque that is a member of `recycle`, not `lru`. Changed to "If there is an entry in `recycle`, it has at least one element".

jsha:
Thanks for the feedback, I've addressed it. I also introduced another small change, simplifying the "insert into vacant entry" case slightly.

algesten (approved):
All looks good to me now!

PR #143, opened 2020-09-13
Release/1.4.1

algesten:


algesten:
@jsha Are we good to publish you think?

jsha:
I think we're close! I'd like to merge the fix to the cookie unittest as a separate change rather than including it under the release-tagging PR. I'll copy it into a PR now.

algesten:
Ok! I'm not in the habit of caring what goes into a PR much since it's not a git function, but a github function. But I can be more strict with it also, no problem.

jsha:
Sounds good. Filed #144. There are a couple of reasons I like to separate topics by PR. The unit of code review on GitHub is the PR, and the code review is a useful source of information. Often if I'm looking at a project's git log, and the commit I'm interested in doesn't have enough context, I'll go look at the conversation on the PR to understand it better. Also, if each PR is an independent unit of work, it's possible to consistently use "Squash and merge" rather than "Create a merge commit." Speaking of which, I haven't previously asked: Do you have a preference between "Squash and merge" and "Create a merge commit"?

On my other projects, I use "Squash and merge." That's so that contributors can push updates to their PRs as a series of small commits, rather than rebasing and force-pushing; but when the PR overall gets merged, it's treated as a unit of work. This also allows the repo to have a mostly straight-through history, which I find easier to understand when tracking down bugs.

algesten:
Makes perfect sense, I'll be stricter with it from now on.

> I haven't previously asked: Do you have a preference between "Squash and merge" and "Create a merge commit"?

I like a linear git history, so I'm in the "rebasing camp". Whether it's squashing or not I think comes down to the quality of the commits in the PR. For big PRs with lots of back/forth and a bunch of "fix after review", I think that when the PR is approved it's better to either rewrite git history by reorganizing the commits and force push before the final merge, or squashing. However, no force pushing while a PR is mid review.

algesten:
Having said that about force pushing. I just allowed myself to do that to get rid of the change you landed in #144

jsha (approved):

PR #144, opened 2020-09-13
Fix cookies_on_redirect unittest.

jsha:
Previously this test relied on a specific ordering of cookies, which was
not guaranteed. This makes the test robust to random ordering.

Also, fix the check for the `second` cookie to actually check the right
cookie.


algesten:
Great!

PR #145, opened 2020-09-14
Use #[cfg(feature = "cookie")] for some functions.

jsha:
Some helper functions introduced as part of the cookies_on_redirect
test are only used by that test, which means they are only used when the
cookie feature is enabled. This causes clippy warnings for some of the
other feature combinations in CI.


algesten:
Good! This prompted me to start this #146

PR #146, opened 2020-09-14
CI: error on dead/unused code for features

algesten:
By using `RUSTFLAGS="-D dead_code -D unused-variables -D unused"`, we tell
the compiler to upgrade warnings for unused code to errors. This combined
with a more elaborate test matrix means we can have the CI catch feature
flag combos that cause warnings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,17 +3,46 @@ on: [push, pull_request]
 name: CI
 
 jobs:
+  doc:
+    name: Docs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Install Rust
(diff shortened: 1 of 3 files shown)


algesten:
@jsha This is ready to merge now. By using a matrix, it tests many combos of features and will help us catch "unused code" warnings. The number of test runs explodes a bit, there is a limit to how many we can have (256), but let's prune later if we hit the limit (the "exclude" directive can be used to remove entire branches).

jsha (approved):
Looks good! The current test count is 52, which is definitely a bit big. Looking at https://github.com/algesten/ureq/actions, it appears we go from 2.5m to run the tests to 4.8m. I think that's still well within acceptable range.

Issue #82, opened 2020-06-22
Add a Resolver interface


jsha:
Right now ureq will always use to_socket_addrs to lookup hostnames. For testing, it would be useful to force all lookups to resolve to localhost and a test server running there. This is particularly useful when a variety of hostnames are needed, or when testing large transfers that would be expensive to send over a real network.

We should add an internal Resolver interface that uses to_socket_addrs by default, but can be overridden during testing to provide mocked-out results.

lolgesten:
@jsha in hreq i got this https://docs.rs/hreq/0.3.0/hreq/trait.RequestBuilderExt.html#tymethod.with_override

Would something like that work here too? it effectively makes it so the entry URI is masked by the override host/port/tls setting. That also works on redirects, so if you follow a redirect chain:

URI1 -> URI2 -> UR1 

also when you get redirected back to URI1, the override kicks in.

jsha:
Yep, the with_override method looks like a great solution! My inclination would be to enforce that it's test-only via `#[cfg(test)]` until someone comes along with a use case that strongly motivates making it public.

lolgesten:
Sure! Let's gate.

rawler:
I have a use-case for this in a production-environment, where I'd like to ensure the "spread" of requests are even between hosts. (Basically round-robin). I'll whip up a PR and let's see what you think.

PR #147, opened 2020-09-15
Add Request::set_addresses to let app override host-resolution

rawler:
This is one simple way to resolve https://github.com/algesten/ureq/issues/82. It has the drawback of not really handling redirection.

Another way to do it would be to implement a "trait Resolver" set in the Agent, and pre-populate it with a default std::net::ToSocketAddrs implementation. Not sure if that is motivated, though. (Not by my use-case)

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -44,6 +44,7 @@ pub struct Request {
     pub(crate) timeout: Option<time::Duration>,
     pub(crate) redirects: u32,
     pub(crate) proxy: Option<crate::proxy::Proxy>,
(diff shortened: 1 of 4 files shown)


rawler:
Hold on. Working on a PR for the alternative approach as well

jsha:
Awesome, thank you for putting together these two PRs! I'll take a look this evening. Off the top of my head I prefer #148 as more flexible (particularly with regards to redirects) but it's great to have two PRs to consider.

jsha:
Closing in favor of #148.

PR #148, opened 2020-09-15
Implement Pluggable Name-resolution

### LINKED ISSUE #82: Add a Resolver interface

rawler:
Alternative to PR#147.

Advantages;
 - API-change local to `Agent`, most users won't notice
 - Handles redirects
Drawbacks:
 - Small (probably neglible) overhead in extra `Mutex`-locking and `Arc`-cloning
 - Cannot do per-request overrides within the same `Agent`-session

Would also resolve #82

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -6,6 +6,7 @@ use std::sync::Mutex;
 use crate::header::{self, Header};
 use crate::pool::ConnectionPool;
 use crate::request::Request;
+use crate::resolve::ArcResolver;
 
 /// Agents keep state between requests.
 ///
@@ -53,6 +54,7 @@ pub(crate) struct AgentState {
     /// Cookies saved between requests.
     #[cfg(feature = "cookie")]
     pub(crate) jar: CookieJar,
+    pub(crate) resolver: ArcResolver,
 }
 
 impl AgentState {
@@ -188,6 +190,29 @@ impl Agent {
         state
             .pool
             .set_max_idle_connections_per_host(max_connections);
+    }
+
+    /// Configures a custom resolver to be used by this agent. By default,
+    /// address-resolution is done by std::net::ToSocketAddrs. This allows you
+    /// to override that resolution with your own alternative. Useful for
+    /// testing and special-cases like DNS-based load balancing.
+    ///
+    /// A `Fn(&str) -> io::Result<Vec<SocketAddr>>` is a valid resolver,
+    /// passing a closure is a simple way to override. Note that you might need
+    /// explicit type `&str` on the closure argument for type inference to
+    /// succeed.
+    /// ```
+    /// use std::net::ToSocketAddrs;
+    ///
+    /// let mut agent = ureq::agent();
+    /// agent.set_resolver(|addr: &str| match addr {
+    ///    "example.com" => Ok(vec![([127,0,0,1], 8096).into()]),
+    ///    addr => addr.to_socket_addrs().map(Iterator::collect),
+    /// });
+    /// ```
+    pub fn set_resolver(&mut self, resolver: impl crate::Resolver + 'static) -> &mut Self {
+        self.state.lock().unwrap().resolver = resolver.into();
+        self
     }
 
     /// Gets a cookie in this agent by name. Cookies are available
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -57,11 +57,7 @@ pub(crate) struct AgentState {
 
 impl AgentState {
     fn new() -> Self {
-        AgentState {
-            pool: ConnectionPool::new(),
-            #[cfg(feature = "cookie")]
-            jar: CookieJar::new(),
-        }
+        Self::default()
     }
     pub fn pool(&mut self) -> &mut ConnectionPool {
         &mut self.pool
--- a/src/resolve.rs
+++ b/src/resolve.rs
@@ -1,0 +1,59 @@
+use std::fmt;
+use std::io::Result as IoResult;
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::sync::Arc;
+
+pub trait Resolver: Send + Sync {
+    fn resolve(&self, netloc: &str) -> IoResult<Vec<SocketAddr>>;
+}
+
+#[derive(Debug)]
+pub(crate) struct StdResolver;
+
+impl Resolver for StdResolver {
+    fn resolve(&self, netloc: &str) -> IoResult<Vec<SocketAddr>> {
+        ToSocketAddrs::to_socket_addrs(netloc).map(|iter| iter.collect())
+    }
+}
+
+impl<F> Resolver for F
+where
+    F: Fn(&str) -> IoResult<Vec<SocketAddr>>,
+    F: Send + Sync,
+{
+    fn resolve(&self, netloc: &str) -> IoResult<Vec<SocketAddr>> {
+        self(netloc)
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct ArcResolver(Arc<dyn Resolver>);
+
+impl<R> From<R> for ArcResolver
+where
+    R: Resolver + 'static,
+{
+    fn from(r: R) -> Self {
+        Self(Arc::new(r))
+    }
+}
+
+impl fmt::Debug for ArcResolver {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ArcResolver(...)")
+    }
+}
+
+impl std::ops::Deref for ArcResolver {
+    type Target = dyn Resolver;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref()
+    }
+}
+
+impl Default for ArcResolver {
+    fn default() -> Self {
+        StdResolver.into()
+    }
+}
--- a/src/test/agent_test.rs
+++ b/src/test/agent_test.rs
@@ -101,6 +101,31 @@ fn connection_reuse() {
     assert_eq!(resp.status(), 200);
 }
 
+#[test]
+fn custom_resolver() {
+    use std::io::Read;
+    use std::net::TcpListener;
+
+    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+
+    let local_addr = listener.local_addr().unwrap();
+
+    let server = std::thread::spawn(move || {
+        let (mut client, _) = listener.accept().unwrap();
+        let mut buf = vec![0u8; 16];
+        let read = client.read(&mut buf).unwrap();
+        buf.truncate(read);
+        buf
+    });
+
+    crate::agent()
+        .set_resolver(move |_: &str| Ok(vec![local_addr]))
+        .get("http://cool.server/")
+        .call();
+
+    assert_eq!(&server.join().unwrap(), b"GET / HTTP/1.1\r\n");
+}
+
 #[cfg(feature = "cookie")]
 #[cfg(test)]
 fn cookie_and_redirect(mut stream: TcpStream) -> io::Result<()> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,6 +125,7 @@ mod header;
 mod pool;
 mod proxy;
 mod request;
+mod resolve;
 mod response;
 mod stream;
 mod unit;
@@ -140,6 +141,7 @@ pub use crate::error::Error;
 pub use crate::header::Header;
 pub use crate::proxy::Proxy;
 pub use crate::request::Request;
+pub use crate::resolve::Resolver;
 pub use crate::response::Response;
 
 // re-export
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -74,10 +74,6 @@ impl Default for ConnectionPool {
 }
 
 impl ConnectionPool {
-    pub fn new() -> Self {
-        Self::default()
-    }
-
     pub fn set_max_idle_connections(&mut self, max_connections: usize) {
         if self.max_idle_connections == max_connections {
             return;
@@ -251,7 +247,7 @@ fn pool_connections_limit() {
     // Test inserting connections with different keys into the pool,
     // filling and draining it. The pool should evict earlier connections
     // when the connection limit is reached.
-    let mut pool = ConnectionPool::new();
+    let mut pool = ConnectionPool::default();
     let hostnames = (0..DEFAULT_MAX_IDLE_CONNECTIONS * 2).map(|i| format!("{}.example", i));
     let poolkeys = hostnames.map(|hostname| PoolKey {
         scheme: "https".to_string(),
@@ -276,7 +272,7 @@ fn pool_per_host_connections_limit() {
     // Test inserting connections with the same key into the pool,
     // filling and draining it. The pool should evict earlier connections
     // when the per-host connection limit is reached.
-    let mut pool = ConnectionPool::new();
+    let mut pool = ConnectionPool::default();
     let poolkey = PoolKey {
         scheme: "https".to_string(),
         hostname: "example.com".to_string(),
@@ -301,7 +297,7 @@ fn pool_per_host_connections_limit() {
 
 #[test]
 fn pool_update_connection_limit() {
-    let mut pool = ConnectionPool::new();
+    let mut pool = ConnectionPool::default();
     pool.set_max_idle_connections(50);
 
     let hostnames = (0..pool.max_idle_connections).map(|i| format!("{}.example", i));
@@ -321,7 +317,7 @@ fn pool_update_connection_limit() {
 
 #[test]
 fn pool_update_per_host_connection_limit() {
-    let mut pool = ConnectionPool::new();
+    let mut pool = ConnectionPool::default();
     pool.set_max_idle_connections(50);
     pool.set_max_idle_connections_per_host(50);
 
@@ -347,7 +343,7 @@ fn pool_update_per_host_connection_limit() {
 fn pool_checks_proxy() {
     // Test inserting different poolkeys with same address but different proxies.
     // Each insertion should result in an additional entry in the pool.
-    let mut pool = ConnectionPool::new();
+    let mut pool = ConnectionPool::default();
     let url = Url::parse("zzz:///example.com").unwrap();
 
     pool.add(
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -4,7 +4,6 @@ use std::io::{
 };
 use std::net::SocketAddr;
 use std::net::TcpStream;
-use std::net::ToSocketAddrs;
 use std::time::Duration;
 use std::time::Instant;
 
@@ -383,14 +382,16 @@ pub(crate) fn connect_host(unit: &Unit, hostname: &str, port: u16) -> Result<Tcp
         unit.deadline
     };
 
-    // TODO: Find a way to apply deadline to DNS lookup.
-    let sock_addrs: Vec<SocketAddr> = match unit.proxy {
+    let netloc = match unit.proxy {
         Some(ref proxy) => format!("{}:{}", proxy.server, proxy.port),
         None => format!("{}:{}", hostname, port),
-    }
-    .to_socket_addrs()
-    .map_err(|e| Error::DnsFailed(format!("{}", e)))?
-    .collect();
+    };
+
+    // TODO: Find a way to apply deadline to DNS lookup.
+    let sock_addrs = unit
+        .resolver()
+        .resolve(&netloc)
+        .map_err(|e| Error::DnsFailed(format!("{}", e)))?;
 
     if sock_addrs.is_empty() {
         return Err(Error::DnsFailed(format!("No ip address for {}", hostname)));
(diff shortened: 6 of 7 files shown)


src/agent.rs
```diff
@@ -40,34 +41,55 @@ pub struct Agent {
     /// Copied into each request of this agent.
     pub(crate) headers: Vec<Header>,
     /// Reused agent state for repeated requests from this agent.
-    pub(crate) state: Arc<Mutex<Option<AgentState>>>,
+    pub(crate) state: Arc<Mutex<AgentState>>,
```
algesten, line 44: I think the resolver might better sit on `Agent` than `AgentState`. That would avoid rewriting this optional.

That would mean it needs to be also `Send + Sync`, we could also require it to extend `Debug`, to continue allow `#[derive(Debug)]`

Since we're not going to deal with `Resolver` that don't have these bounds, I think for simplicity we could just jam all the requirement into the trait itself.

```rust
pub trait Resolver: Debug + Send + Sync + 'static {
  ...
}

pub struct Agent {
   …
  pub(crate) resolver: Arc<dyn Resolver>,
}
```

algesten (changes requested):
I like where this PR is going!

I think it would be worth exploring putting the resolver on `Agent`. Overall I get a feeling the PR currently touches a few too many things (like changing the `Option<AgentState>`). So one hope I have is that my suggested change would mean less implications to the rest of the code.

src/agent.rs
rawler: Hmm, so having the resolver on the Agent, would that imply that the resolver will be copied to the request when making one?

The reason I chose not to, was to avoid extra Arc::cloneing, but perhaps that's a low price to pay.

rawler:
> changing the `Option<AgentState>`

I think this is still worthwhile in and off itself, though? Or are there a reason for why it's Option? (I cannot see when/where it's ever set to None?)

src/agent.rs
algesten: I see what you mean. `Unit` has `AgentState` already 🤔

algesten:
> I think this is still worthwhile in and off itself, though? (Perhaps in a separate PR?) Or are there a reason for why it's Option? (I cannot see when/where it's ever set to None?)

Yes. Might very well be worthwhile. What I meant is that it would be good to reason about that change in a separate PR.

rawler:
Ok, created two separate PR:s for the separate changes. Let's hash those out first, and then maybe return to where the Resolver should reside.

src/resolve.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub trait Resolver: std::fmt::Debug + Send + Sync {
+    fn resolve(&self, netloc: &str) -> std::io::Result<Box<dyn Iterator<Item = SocketAddr>>>;
+}
+
+#[derive(Debug)]
+pub(crate) struct StdResolver;
+
+impl Resolver for StdResolver {
+    fn resolve(&self, netloc: &str) -> std::io::Result<Box<dyn Iterator<Item = SocketAddr>>> {
```
jsha, line 11: Why return an iterator when `ureq` always takes the first item? Also why is the Box required?

jsha (commented):
Some use cases I'd like to be able to support:

 1. Explicit round-robin or other load balancing strategies (your use case).
 2. Unconditionally resolving everything to a specific IP address.
 3. Unconditionally resolving `example.com:1234` to a specific IP address.
 4. Replacing the Rust std resolver with a different one - for instance, one that allows setting request timeouts, or one that uses DoH.

From an API perspective, I think my ideal would be passing an `FnMut + Sync` instead of defining a trait, because it requires less boilerplate and is very convenient for cases (2) and (3):

    agent.set_resolver(|x|
      if (x == "example.com:1234") {
        Some("127.0.0.1:8080".parse())
      } else {
        None // revert to regular DNS
      }
    });

Returning an Option allows the user to say "I don't care about this name; do the normal thing with it."

Requiring that the closure be `Sync` means we can call it concurrently without locking, which is nice for performance in the cases that don't require locks. In use case (1) the user would presumably have the closure call some type that does its own locking and tracks "which backend to call next."

Since we're supporting the round-robin use case, we should also consider what other load balancing strategies we want to make implementable. For instance: How would one implement least-loaded or fastest load balancing? To do those, I think we would need a backchannel for user code to know when a request to given IP address finishes. One way to do that: `Response` could get an `address` field that would return the address that was actually chosen for the request. I don't think we need to implement that in this PR, just presenting it as a case study for "is this extensible enough," and I think it is.

We should clearly define the semantics of whether `ureq` is allowed to internally cache the results of a custom resolver lookup. Since our goal is to support round-robin, the answer should be no. However, I can see us wanting to build an internal cache of results from the default resolver, in case we find system resolvers are inadequate. For instance, curl does this: https://ec.haxx.se/libcurl/libcurl-names#caching. I'd want to investigate more before concluding that was worthwhile, but this suggests to me we should be able to distinguish between when we're using the default resolver and when we're using a custom one.

One last complication: DNS resolution happens *after* attempting to pull a connection from the pool. So if you're making all your requests sequentially from the same Agent, you'll keep reusing the same connection and send them all to one host.

I think you can solve this by constructing one Agent for each of your backend IP addresses, and doing round-robin across those agents.

algesten:
```rust
use std::net::SocketAddr;

pub trait Resolver: Send + Sync {
    fn resolve(&self, host_port: &str) -> Option<SocketAddr>;

    fn reuse_cached(&self, _host_port: &str, resolved: SocketAddr) -> Option<SocketAddr> {
        Some(resolved)
    }
}

impl<F: Send + Sync + 'static> Resolver for F
where
    F: Fn(&str) -> Option<SocketAddr>,
{
    fn resolve(&self, host_port: &str) -> Option<SocketAddr> {
        (self)(host_port)
    }
}


use std::sync::Arc;

fn set_resolver_override(resolver: impl Resolver) {
    let resolver = Arc::new(Box::new(resolver));
}

fn test_set_resolver() {
    set_resolver_override(|host_port: &str| {
        if host_port == "example.com:1234" {
            "127.0.0.1:8080".parse().ok()
          } else {
            None // revert to regular DNS
          }
    })
}
```

@jsha how about something like this?

* It's a trait that allows for more configurability to support the more complicated load balancing. Not saying `reuse_cached` is enough, but we can add all the hooks we need to the trait.
* A default impl for `Fn` giving you a quick override for the simple cases.

One difference is that it's not `FnMut` like you wanted, but there is a contradiction in `FnMut` combined with `Sync`.

algesten:
One box too many. The `Arc` is enough.

```rust
    let resolver: Arc<dyn Resolver> = Arc::new(resolver);
```

jsha:
It's getting late here (PDT) so I'll have to give your proposal a more thorough and bright-eyed look tomorrow, but on first glance, it still feels a bit complicated. The more I think about it, the more I think plain `Fn` works best. The disadvantage is that the resolver can't store mutating state (like which IP to hit next), but since realizing this:

> One last complication: DNS resolution happens after attempting to pull a connection from the pool. So if you're making all your requests sequentially from the same Agent, you'll keep reusing the same connection and send them all to one host.
> I think you can solve this by constructing one Agent for each of your backend IP addresses, and doing round-robin across those agents.

I think mutating state doesn't really help with the round-robin use case anyhow. The bright side is that there's an alternate solution (maintain an array of Agents) that can be implemented with a static override per Agent. I'd be very curious to hear from @rawler if that seems to address their use case.

rawler:
@jsha Maintaining Pool of Agents could work, although making it a tad bit more complicated. (The array of backends change, so that pool would have to be continuously updated). Not a big problem though, and on the plus-side, that pool could be a bit more clever than "Round Robin". We could wrap `ureq::Response`s in a wrapper-struct and return backends to the pool when responses finish, for even better balancing.

Another option I guess could be to simply disable connection reuse (set to 0?). (For our use-case the average request is several seconds, so connection setup isn't really a problem)

A couple of questions regarding the resolver-interface though; I see in the existing code that it does fallback-tests of the resolved addresses, in order? This would not be possible through this API? Also, if an answer of `None` means "fallback", how would the resolver signal "Host Not Found"? Isn't it easier for the Resolver to itself fallback to `ToSocketAddrs`?

jsha:
>  I see in the existing code that it does fallback-tests of the resolved addresses, in order? This would not be possible through this API?

You're right, I misread the code about picking one address versus falling back to other addresses. That said, if the interface provides a way to say "fall back to ToSocketAddr", we can have a list of addresses in the default case, and a single address in the override case. So the question is: do we ever need multiple addresses in the override case? Curl's doesn't have it: https://curl.haxx.se/libcurl/c/CURLOPT_RESOLVE.html. We would need it if we wanted to fully support use case (4) above (allow plugging in a DoH or other implementation), but on revisiting that, I'm thinking a simpler API is worth the tradeoff.

>  Also, if an answer of None means "fallback", how would the resolver signal "Host Not Found"? Isn't it easier for the Resolver to itself fallback to ToSocketAddrs?

Good point about Not Found. I guess this comes back to what use cases we want to support. If we expect the resolver to be doing network requests (fully replacing the resolver), we need to support Host Not Found - we could do that by returning an io::Result. But if we only care about pre-overriding certain known names with certain known IP addresses, we don't need a Host Not Found. I'm currently leaning towards the latter, on the basis that curl's override options don't provide something more sophisticated, and that seems to have worked, but I could definitely be convinced otherwise.

src/resolve.rs
rawler: `ureq` doesn't always take the first item. There is a succession of fallbacks in https://github.com/algesten/ureq/blob/6a88c2c8bfbb0bac97029265a42703a17c76e01c/src/stream.rs#L408.

An alternative to the `Box<dyn Iterator>` would be returning a `Vec<SocketAddr>` directly. If aiming for just simpler resolvers, `Vec<SocketAddr>` might be simpler.

rawler:
For me, the proposed interface `fn resolve(&self, host_port: &str) -> Option<SocketAddr>` is absolutely workable, using the `AgentPool` above, so I have no strong feelings, either or.

> I'm thinking a simpler API is worth the tradeoff

Subjective, of course, but personally I think an API similar to `ToSocketAddr`:  `fn resolve(&self, host_port: &str) -> io::Result<Vec<SocketAddr>>` is about as simple as it gets.
  - The semantics of the API does not depend on an "alternative" fallback
  - The meaning is (at least to me, again subjective) easily understood
  - Since the API supports a full resolver
    - The default could be a simple implementation from `std::net::ToSocketAddr`
    - `stream::connect_host` can be simple and straight-forward - with no branching in code
    - The API has flexibility enough for the default to potentially implement its own caching, as mentioned
  - The `Vec`-allocation is unfortunate, but also already present in `stream::connect_host`

But, again, both options work for my use-case, and I'm happy either/or. :)

As a side-note on the topic of connection-reuse, one option to solve connection-reuse and round/robin would be to base the connection cache on (hostname, socketaddr). That could be useful for existing name-resolution as well, since name-based services often de-register themselves with DNS some time before forcefully disconnecting.

rawler:
Alright, made an optimistic change with some the discussions included.
 - Resolver is now set on `Agent` directly, and Arc-copied to `Request` and `Unit`.
 - Resolver is still it's own trait (I tried changing to `Fn(&str)`, but hit problems, do not remember exactly what ATM). Instead, there is now a blanket implementation of the trait for such `Fn`:s, for similar advantages.
 - `Box<dyn Iterator<Item=SocketAddr>>` is now simplified to `Vec<SocketAddr>`. No lazy-generation, but not likely usage-scenario anyways

It's not the simple `&str -> Option<SocketAddr>`, but not much more complicated.

src/test/agent_test.rs
```diff
@@ -101,6 +101,31 @@ fn connection_reuse() {
...
+    let server = std::thread::spawn(move || {
+        let (mut client, _) = listener.accept().unwrap();
+        let mut buf = vec![0u8; 16];
+        let read = client.read(&mut buf).unwrap();
+        buf.truncate(read);
+        buf
+    });
+
+    crate::agent()
+        .set_resolver(move |_: &str| Ok(vec![local_addr]))
```
jsha, line 122: Why is move needed here?

src/agent.rs
```diff
@@ -194,6 +197,29 @@ impl Agent {
...
+    /// ```
+    /// use std::net::ToSocketAddrs;
+    ///
+    /// let mut agent = ureq::agent();
+    /// agent.set_resolver(|addr: &str| match addr {
+    ///    "example.com" => Ok(vec![([127,0,0,1], 8096).into()]),
+    ///    addr => addr.to_socket_addrs().map(Iterator::collect),
+    /// });
+    /// ```
+    pub fn set_resolver(&mut self, resolver: impl crate::Resolver + 'static) -> &mut Self {
```
jsha, line 218: Why the `'static` lifetime bound? Also why does this need to be a trait object as opposed to just an object that implements the trait?

src/agent.rs
```diff
@@ -41,6 +42,7 @@ pub struct Agent {
     pub(crate) headers: Vec<Header>,
     /// Reused agent state for repeated requests from this agent.
     pub(crate) state: Arc<Mutex<AgentState>>,
+    pub(crate) resolver: ArcResolver,
```
jsha, line 45: We have a few different options for this type:

 1. Resolver
 2. Arc<Resolver>
 3. Arc<dyn Resolver>
 4. ArcResolver (similar to (3))

Would you mind describing the considerations that went into choosing (4)? My default would be (1) but I'm guessing there's something I'm missing.

jsha (commented):
This looks really good! I feel like the tradeoffs made are decent, and I like that it's possible to provide a simple closure as a resolver. I don't feel too strongly about returning None to fallback to the default impl vs calling ToSocketAddrs explicitly, so I'm happy to go with your recommendation here to call ToSocketAddrs explicitly.

src/agent.rs
jsha: Looking at this again, I see that Unit and Request store `Arc<Mutex<AgentState>>` rather than an an `Arc<Agent>`. I think switching Unit and Request to store an `Arc<Agent>` instead would make for a nicer change here, and would probably remove the need for an Arc on the Resolver object. That could be a separate change or part of this one if you prefer.

jsha: FYI I took a stab at this refactoring and realized it's a bit more complicated than I thought, since the `Arc` has to own the `Agent`, but we expect the user code to own the Agent.

I'll try another round with moving the Mutex inside the AgentState.

algesten: >  1.  Resolver
>  2. Arc<Resolver>
>  3. Arc<dyn Resolver>
>  4. ArcResolver (similar to (3))
>
> Would you mind describing the considerations that went into choosing (4)?

@jsha This is how we need to do it in Rust.

1. Since `Resolver` is a trait, the size of the concrete implementation is not known. The only way to handle that unknown would be to make `Agent` generic over resolver. I.e. we would have something like `Agent<R> where R: Resolver`. That in turn would force all enclosing types to also be generic over `R` and it would "bubble" to the API surface making for an awkward API. 

2. This syntax is pre rust edition 2018. It works, but is deprecated and causes a warning. The idea is that the `dyn` keyword makes it clear that we're talking about a trait object.

3. This is the good way. It doesn't make the API generic over the trait, and it does so by creating a vtable and doing dynamic dispatch. It's heap allocated, so the size of the actual concrete type is "hidden" via pointer indirection.

4. This is the [new type idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html). It relies on `Deref` to access the held data. I suspect @rawler used it because it also gives us a chance to implement `Debug` for all the wrapped concrete impls.

@rawler My preference would be 3, cause the `ArcResolver` creates a level of indirection which I don't think helps the reader of the code much. What do you think?

src/agent.rs
algesten: @jsha 

> Why the `'static` lifetime bound?

Since we're taking ownership of the concrete resolver impl, it's necessary in this case. If we turn it around, the types that _don't_ implement `'static` are reference types (they live for the duration of the reference). The only way to accept reference types would be to make `Agent` generic over the `Resolver` so that the entire struct would have the lifetime of the reference.

> Also why does this need to be a trait object as opposed to just an object that implements the trait?

Isn't that the same? These two are pretty much identical:

1. `fn set_resolver<R: Resolver + 'static>(&mut self, resolver: R) -> &mut self`
2. `fn set_resolver(&mut self, resolver: impl Resolver + 'static) -> &mut self`

Compilation wise they both cause a _generic function_, and must do so since the size of `impl Resolver` isn't known. Machine code explains why: the size of the stack frame must be known before jumping to the function ([monomorphization](https://rustc-dev-guide.rust-lang.org/backend/monomorph.html)).

We could have opted for a signature accepting `Arc<dyn Resolver>`, forcing the caller to box up the concrete impl before calling. That would make for more efficient compilation, but less good ergonomics.

algesten:
@rawler I'm liking this iteration a lot. The one thing is I'd prefer to not have the "extra" wrapper type `ArcResolver`.

@jsha @rawler It's a bit unfortunate to return `Vec<SocketAddr>`, cause it means forcing the implementor to do an extra heap allocation. Maybe we should consider doing it like `ToSocketAddr` does? Here's a proof of concept:

```rust
use std::io::Result as IoResult;
use std::net::{SocketAddr, ToSocketAddrs};

/// Resolver impl with output type `Iter` being limited 
/// to `Iterator` trait that produces `SocketAddr`.
pub trait Resolver {
    type Iter: Iterator<Item = SocketAddr>;
    fn resolve(&self, host_port: &str) -> IoResult<Self::Iter>;
}

/// Blanket impl for quick override. Closures are expected to produce `IoResult<Vec<SocketAddr>>`
///
/// Here we can allow ourselves `Vec`, cause it's convenience. If the lib user
/// want to avoid it, they can by using the trait directly.
///
/// ```
/// agent.set_resolver(|host_port: &str| {
///     if host_port == "example.com:1234" {
///         Ok(vec!["127.0.0.1:8080".parse().unwrap()])
///     } else {
///         Ok(vec![])
///     }
/// })
/// ```
/// 
impl<F> Resolver for F
where
    F: Fn(&str) -> IoResult<Vec<SocketAddr>>,
{
    type Iter = std::vec::IntoIter<SocketAddr>;

    fn resolve(&self, host_port: &str) -> IoResult<Self::Iter> {
        let vec = (self)(host_port)?;
        Ok(vec.into_iter())
    }
}

pub struct StdResolver;

impl Resolver for StdResolver {
    type Iter = std::vec::IntoIter<SocketAddr>;
    fn resolve(&self, host_port: &str) -> IoResult<Self::Iter> {
        host_port.to_socket_addrs()
    }
}
```

@rawler as a side note. I fully agree with doing `io::Result` instead of `IoResult`. We should do a separate PR tidying up the entire code base – until then we keep the ugly `IoResult`.

src/test/agent_test.rs
rawler: To move `local_addr` into the closure. The closure need 'static lifetime (see below)

src/agent.rs
rawler: 4 were, as you said, chosen to be able to provide a blanket `Debug`, but mainly to provide a `Default` implementation. Without it, we need to manually implement `Default` for the entire `Request` structure.

I don't mind explicit implementation, would also solve the problem of the `Default` now being an unusable value. I just got the impression that leaning on derived implementation was idiomatic in this codebase.

rawler:
@algesten traits with associated types cannot be boxed. (How would you know what type of Iterator the boxed Resolver returns?)

If we still want Arc<Resolver> over ArcResolver, perhaps I should make a separate PR to implement Request::default first?

algesten:
@rawler 

> traits with associated types cannot be boxed… `Box<dyn Iterator<Item=SocketAddr>>`

Of course! And we've done full circle :) 

Nvm me. Let's do `Vec`.

algesten:
> If we still want Arc<Resolver> over ArcResolver, perhaps I should make a separate PR to implement Request::default first?

I think you motivated `ArcResolver` existence. I will park that line of thought.

rawler:
Cool! So, are we satisfied having the Resolver in Agent, Request and Unit, or should we go back to putting it in AgentState? (Note that the long-lock-problem can be solved by cloning the Arc<Resolver> out from the AgentState on usage, and calling it while not locking AgentState)

algesten:
@rawler 

> Cool! So, are we satisfied having the Resolver in Agent, Request and Unit, or should we go back to putting it in AgentState? (Note that the long-lock-problem can be solved by cloning the Arc out from the AgentState on usage, and calling it while not locking AgentState)

I think we could push ahead and put it inside `AgentState` for now (and clone it out/release mutex when needed as you suggest). Then a later PR could refactor that.

rawler:
Cool! Then just to summarize, I think we're mostly back to original PR :) with
 - slightly simpler API (`Box<dyn Iterator<Item=SocketAddr>>` → `Vec<SocketAddr>`)
 - the blanket implementation for `Fn(&str) -> IoResult<Vec<SocketAddr>>`
 - `ArcResolver` to solve the `Debug` impl-problem required for the `Fn`-impl

algesten (approved):
@rawler thanks for persisting!

I'm happy to merge this. Going to leave it for @jsha to have time to react, but I say we are in a position to merge this weekend.

rawler:
@rawler thanks for persisting!

> Ingen orsak! Thank you for building a valuable library for us :)

jsha (approved):
This also looks good to me. Thanks for walking me through the motivations for all the choices here!

PR #149, opened 2020-09-17
Turn `Option<AgentState>` into `AgentState`

rawler:
`Some(AgentState)` seem to be assumed pretty much everywhere. I could not find any test or piece of code hinting at how `None` should be interpreted, or even see how a state of `None` could even be constructed.

If the `Option` don't serve a purpose (anymore), the code can be both simpler and most likely more performant without it.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -27,9 +27,8 @@ use crate::pool::DEFAULT_HOST;
 /// It's a "unit of work". Maybe a bad name for it?
 ///
 /// *Internal API*
-#[derive(Debug)]
 pub(crate) struct Unit {
-    pub agent: Arc<Mutex<Option<AgentState>>>,
+    pub agent: Arc<Mutex<AgentState>>,
     pub url: Url,
     pub is_chunked: bool,
     pub query_string: String,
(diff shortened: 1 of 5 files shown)


algesten:
> I could not find any test or piece of code hinting at how None should be interpreted, or even see how a state of None could even be constructed.

The methods from the top of the lib:

`ureq::get("https://myserver.com/")` takes the path: `ureq::request("GET, "https://myserver.com/")`. This in turn does a `Agent::new().request(method, path)`

That `Agent::new()` is just `Default::default()` which fills in `None` in the optional.

The idea here is that since these top level calls are effectively one-shot without a retained state between calls, they will just drop any connections and not keep any cookies. The most efficient way is not instantiating an agent in the first place.

Having said that there are (at least?) two reasons we might want this change anyway.

1. It's not particularly expensive and makes for cleaner code. The new behavior is that the connection will at the end of these one-shot calls be returned to the `ConnectionPool` then dropped when the entire thing is deallocated. To retain the current "optimization", we could simply set `max_idle_connections` to `0` for these kind of "one shot" agents.
2. Cookie retention on redirects. Currently, the top level calls, maybe surprisingly, do not retain cookies when following redirects. It probably is more correct to have a fully functioning implicit `Agent` for this purpose.

@jsha Do you see any holes in my thinking?

algesten:
And thanks for taking time to do this PR!

jsha:
@algesten Yep, your analysis makes sense! I also think it makes sense for `ureq::get()` to use a "real" agent that's capable of recycling connections. For instance, consider the case of redirects: The current behavior is that a new connection must be established for each in a chain of redirects. But with this change, the connection can be reused so long as the redirects are on the same host.

This would also make the `ureq::get()` case exercise more of the same code paths as `Agent::get()`, which is good for ensuring that bugs don't get missed because only one code path was getting common use.

src/unit.rs
```diff
@@ -369,7 +363,11 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> IoResult<()> {
         }
     }
     if !header::has_header(&unit.headers, "user-agent") {
-        write!(prelude, "User-Agent: ureq/{}\r\n", env!("CARGO_PKG_VERSION"))?;
+        write!(
```
jsha, line 366: This looks like a formatting-only change. I'd recommend reverting this so we don't have a spurious diff to these lines in the log.

algesten:
> But with this change, the connection can be reused so long as the redirects are on the same host.

@jsha Good point! That means also we should just leave `max_idle_connections` on default value.

@rawler If you remove that code formatting change from the PR, we're ready to merge this.

rawler:
@algesten :+1: I have dropped the formatting-change

Out of curiosity, what's the project-ambition with regards to code formatting? Separate PR:s occasionally with "cargo fmt" applied?

jsha:
> Out of curiosity, what's the project-ambition with regards to code formatting? Separate PR:s occasionally with "cargo fmt" applied?

Yep! In generally I think we should aim to keep all new code formatted on commit, and periodically fix up existing code if necessary. I had originally thought that the diff I mentioned might have been a difference between your version of rustfmt and my version (I used rustfmt integration built-in to vscode). But when I run `rustfmt` manually on the command line, it produces the same diff you had. Not sure what's going on there.

jsha (approved):

algesten:
@rawler @jsha let's do it #151

PR #150, opened 2020-09-17
ConnectionPool::default: Explicit implementation

rawler:
The auto-derived implementation would have unexpectedly initialized the
`usize` fields to 0. Having a `::default()` leading to an unuseable value
is counter-intuitive.


algesten (approved):
You find another of these cases in `request.rs`, `redirects: 5` would get 0 if one uses default directly.

PR #151, opened 2020-09-19
cargo fmt

algesten:


jsha (approved):
Thanks! I meant to do this after our recent conversation. I'll have to figure out what's up with my editor and fix it.

rawler:
Perhaps something like this added to GitHub? https://github.com/marketplace/actions/rust-action

algesten:
I would like that!

PR #156, opened 2020-09-25
Add Send + Sync marker traits to into_reader.

jsha:
This allows the resulting Read to be shared among threads.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -288,7 +288,7 @@ impl Response {
     /// assert_eq!(bytes.len(), len);
     /// # }
     /// ```
-    pub fn into_reader(self) -> impl Read {
+    pub fn into_reader(self) -> impl Read + Send + Sync {
         //
         let is_http10 = self.http_version().eq_ignore_ascii_case("HTTP/1.0");
         let is_close = self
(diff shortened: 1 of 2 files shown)


algesten (commented):
Makes a lot of sense.

I'm unsure about `Sync` here. That means we guarantee the reader isn't only possible to pass around to other threads, but it's also consistently observable simultaneously in all threads.

In practice our impl maybe always wrap the actual thing in a mutex, so we do guarantee that. I'm thinking maybe we some future impl of our reader has some sweet optimization we can't uphold with `Sync` which means a breaking change. But maybe that's too much thinking?

Since we don't impl `Clone` and `read()` call is using `&mut self`, in practice it maybe doesn't make much difference?

@rawler maybe has some input? :)

jsha:
Pushed a change to narrow this to just `Send`, which is what I need to [unblock some work on git2-rs](https://github.com/rust-lang/git2-rs/pull/625).

In practice it looks like our Read gets Sync because [TcpStream](https://doc.rust-lang.org/beta/std/net/struct.TcpStream.html#impl-Sync), [StreamOwned](https://docs.rs/rustls/0.18.1/rustls/struct.StreamOwned.html#impl-Sync), and [Cursor](https://doc.rust-lang.org/std/io/struct.Cursor.html#impl-Sync) all get it. My thinking was that with `dyn`, once we've "cast away" the marker traits, it's impossible for user code ever to get them back (as opposed to generics, where the marker traits automatically propagate through). In that context it felt a shame to throw away properties that might be valuable to the user. But I do see your point: the `impl Read` return hides some of our implementation details (like ChunkDecoder, DeadlineStream, PoolReturnRead) from the user, which gives us flexibility; and one of the things we might want to keep hiding is the `Sync`'ness so we have flexibility there.

algesten:
Perfect. Let's merge!

PR #157, opened 2020-09-25
Fix set_max_idle_connections(0).

jsha:
This broke during some recent refactorings. The special case branch for
max_connections == 0 wasn't actually setting the max_idle_connections
field.

This change simply deletes that branch, since the existing code that
whittles down the pool size one element at a time suffices. In the
common case this will be set on an empty pool anyhow.


algesten:
👍

Issue #155, opened 2020-09-20
Fields copied from Request to Unit


jsha:
Reading through https://github.com/algesten/ureq/pull/148, I noticed that Unit makes a copy of a lot of fields that are present on Request:

            url: url.clone(),
            timeout_connect: req.timeout_connect,
            timeout_read: req.timeout_read,
            timeout_write: req.timeout_write,
            method: req.method.clone(),
            proxy: req.proxy.clone(),

But Unit already has a reference to Request. What do you think of removing these fields and delegating accesses to them to `unit.req` instead? Could save a bit of copying and memory use, but more importantly it could make it clearer where these fields are ultimately coming from.

algesten:
Sounds good! I'm not even sure why it is like this.

PR #158, opened 2020-09-26
Add `req` field to Unit and remove cloned parts from request

drivasperez:
I thought I'd give #155 go, to see more of the inside of this crate. Instead of cloning most of `Request`'s fields individually when creating a `Unit`, this PR switches to just cloning `Request` and stuffing it in `Unit`, and changes references to `unit.[field]` to `unit.req.[field]` where appropriate.

It feels like it _should_ be possible to just keep a `&'a Request` or `Cow<'a, Request>`, but trying to implement that I fell down a hole of lifetimes that felt quite likely to break everything, so I stuck with just cloning (I'm pretty new to Rust). I'm not sure how much difference it would make.


jsha (approved):
Welcome to the project! This looks good to me.

algesten:
image

mmm… nice!

Thanks!

PR #160, opened 2020-09-27
bugfix: Don't re-pool streams on drop.

jsha:
We only want streams in the pool that are ready for the next request,
and don't have any remaining bytes of the previous request. That's the
case when we return a stream due to EOF on the underlying `Read`.

However, there was also an `impl Drop` that returned the stream to the
pool on drop of a PoolReturnRead. That resulted in a bug: If a user
called `response.into_reader()`, but did not read the entire body,
a dirty stream would be returned to the pool. The next time a request
was made to the same host, that stream would be reused, and instead of
reading off "HTTP/1.1 200 OK", ureq would read a remnant of the previous
request, resulting in a BadStatus error.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -426,9 +426,3 @@ impl<R: Read + Sized + Into<Stream>> Read for PoolReturnRead<R> {
         Ok(amount)
     }
 }
-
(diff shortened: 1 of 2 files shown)


algesten:
Hmm… I merged, but maybe we want to do something where it does return to pool if it is fully read?

PR #167, opened 2020-09-28
Fix bug in force-unwrapping when resetting timers

algesten:
When running tests locally, this error can surface.

```
---- test::agent_test::custom_resolver stdout ----
thread 'test::agent_test::custom_resolver' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', src/stream.rs:60:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The problem is that setting the timeouts might fail, and this is done
in a From trait where there is not possibility to "bubble" the
io::Error.

```
socket.set_read_timeout(None).unwrap();
socket.set_write_timeout(None).unwrap();
```

This commit moves the resetting of timers to an explicit `Stream::reset()` fn
that must be called every time we're unwrapping the inner stream.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -53,14 +53,7 @@ impl DeadlineStream {
 
 impl From<DeadlineStream> for Stream {
     fn from(deadline_stream: DeadlineStream) -> Stream {
-        // Since we are turning this back into a regular, non-deadline Stream,
-        // remove any timeouts we set.
-        let stream = deadline_stream.stream;
-        if let Some(socket) = stream.socket() {
-            socket.set_read_timeout(None).unwrap();
-            socket.set_write_timeout(None).unwrap();
-        }
-        stream
+        deadline_stream.stream
     }
 }
 
@@ -157,6 +150,17 @@ impl Stream {
             Stream::Https(_) => true,
             _ => false,
         }
+    }
+
+    pub(crate) fn reset(&mut self) -> io::Result<()> {
+        // When we are turning this back into a regular, non-deadline Stream,
+        // remove any timeouts we set.
+        if let Some(socket) = self.socket() {
+            socket.set_read_timeout(None)?;
+            socket.set_write_timeout(None)?;
+        }
+
+        Ok(())
     }
 
     pub(crate) fn socket(&self) -> Option<&TcpStream> {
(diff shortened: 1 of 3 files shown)


algesten:
@jsha I tried going down a different path using rust's `std::convert::TryInto`, but it goes into a rabbit hole cause we're relying on multiple levels of nested `Into` traits.

src/stream.rs
```diff
@@ -159,6 +152,17 @@ impl Stream {
         }
     }
 
+    pub(crate) fn reset(&mut self) -> io::Result<()> {
```
jsha, line 155: I'd somewhat rather have this return `io::Result<Stream>`, and remove the `From<DeadlineStream>`. My thinking is that the `From` is no longer a perfect conversion, since you can wind up with a stream that has leftover deadlines on it. That might be a silent surprise down the line.

jsha (commented):
This looks great, and good catch!

src/stream.rs
algesten: @jsha Yes in principle I agree, however, this is what I was alluding to with "the rabbit hole".

When we do `Response::into_reader()` we get something like (and the equivalent for content-length case):

```rust
        let stream = DeadlineStream::new(stream, deadline);
...
Box::new(PoolReturnRead::new(unit, ChunkDecoder::new(stream)))
```

`PoolReturnRead(ChunkDecoder(DeadlineStream(Stream)))`

`PoolReturnRead` is defined as:

```rust
impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
    pub fn new(unit: Option<Unit>, reader: R) -> Self {
```

That `Into<Stream>` trait is the rabbit hole, cause when we go:

```rust
let stream: Stream = pool_return_read.into();
```

Rust compiler is smart and unnests all three levels down to `Stream` and in that nesting we got `ChunkDecoder` which is external to ureq.

The good rust solution would be to use the `TryFrom` trait instead of `From` (which `Into` uses). But then we're looking at bigger changes.

jsha: Ah, that makes sense! I'm convinced this is a good change for now. We can try and make improvements later. Thanks for finding this bug.

jsha (approved):
It looks like the test failure was due to the flaky test that I think was fixed by #168. Merging master and resolving the conflict should get tests to pass.

PR #168, opened 2020-09-28
Handle ConnectionReset+ConnectionAbort at any time

### LINKED ISSUE #165: Intermittent test failures on dirty_stream_not_returned
We just got a test failure on master:

```
---- test::agent_test::dirty_streams_not_returned stdout ----
thread 'test::agent_test::dirty_streams_not_returned' panicked at 'resp_to_be_dropped failed: BadStatus', src/test/agent_test.rs:214:9
```

Doesn't reproduce after a couple of tries locally. Not sure what the problem is. But we should probably consider this a blocker on releasing 1.5.0.

jsha:
Previously we had a special case for BadStatusRead that would happen
only when we got a ConnectionAborted error reading the status line.
However, sometimes we get ConnectionReset instead. Also the HTTP
spec says that idempotent requests may be retried anytime a connection
is closed prematurely.

The change treats as retryable any ConnectionAborted OR ConnectionReset
error while reading the status line and headers. It removes the special
case BadStatusRead error.

Fixes #165 (I think).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,5 @@
 use std::fmt;
-use std::io;
+use std::io::{self, ErrorKind};
 
 /// Errors that are translated to ["synthetic" responses](struct.Response.html#method.synthetic).
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


algesten (approved):
Great investigation!

PR #169, opened 2020-09-29
Use cookie_store crate instead of cookie::CookieJar

jsha:
CookieJar doesn't support the path-match and domain-match algorithms from [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.1.3), while cookie_store does.

This fixes some issues with the cookie matching algorithm currently in ureq. For instance,
the domain-match uses substring matching rather than the RFC 6265 algorithm.

This deletes two tests:

match_cookies_returns_nothing_when_no_cookies didn't test much
agent_cookies was failing because cookie_store rejects cookies on the `test:` scheme. The way around this is to set up a testserver - but it turns out cookies_on_redirect already does that, and covers the same cases and more.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ json = ["serde", "serde_json"]
 charset = ["encoding"]
 tls = ["rustls", "webpki", "webpki-roots"]
 native-certs = ["rustls-native-certs"]
-cookies = ["cookie"]
+cookies = ["cookie", "cookie_store"]
 socks-proxy = ["socks"]
 
 [dependencies]
@@ -39,7 +39,7 @@ serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
 encoding = { version = "0.2", optional = true }
 native-tls = { version = "0.2", optional = true }
-cookie_store = "0.12.0"
+cookie_store = { version = "0.12.0", optional = true }
 
 [dev-dependencies]
 serde = { version = "1", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -39,6 +39,7 @@ serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
 encoding = { version = "0.2", optional = true }
 native-tls = { version = "0.2", optional = true }
+cookie_store = "0.12.0"
```
algesten, line 42: This needs to be optional on the cookie feature or we pull in the deps for all features.

algesten (approved):
This does subtly change out behavior which could be a breaking change. However being more correct might be worth it.

jsha:
I agree it's a bit of a breaking change depending on use cases. To enumerate the changes:
 - Cookies could previously be sent to a wrong domain - e.g. a cookie set on `example.com` could go to `example.com.evil.com` or `evilexample.com`. Probably no one was relying on this, since it's quite broken.
 - A cookie with a path of `/foo` could be sent on a request to `/foobar`, but now it can't.
 - Cookies could previously be set on IP addresses, but now they can't.
 - Cookies could previously be set for domains other than the one on the request (or its parents), but now they can't.
 - When a cookie had no domain attribute, it would previously get the domain from the request, and subsequently be sent to that domain and all subdomains. Now, it will only be sent to that exact domain (host-only).

That last one is probably the most likely to break people, since someone could depend on it without realizing it was broken behavior.

algesten:
I say let's merge it.

PR #170, opened 2020-09-29
Add log dependency.

### LINKED ISSUE #163: Optional dependency on `log`
It's often useful to see what ureq's doing under the hood. I was recently working on a project that used [`log`](https://docs.rs/log/0.4.11/log/) and found it very handy. What do you think of adding the `log` package as an optional dependency?

Here are the things I'd like to log:

info:
 - When a request is made, with the method and URL
 - (maybe debug) when a response is received, with method, URL, and status code
 - (maybe debug) when a response errors

debug:
 - Outgoing request headers (with Authorization and Cookie redacted)
 - Incoming response headers (with Set-Cookie redacted)
 - When a connection is returned to the pool
 - When a connection is taken from the pool
 - When a new connection is made
 - When a cookie is set (just the name, domain, and expiration, not the contents)
 - When chunked send is chosen automatically

For reference about log levels, here's what `rustls` shows at debug level during a typical request:

```
[2020-09-27T02:26:49Z DEBUG rustls::client::hs] No cached session for DNSNameRef("github.com")
[2020-09-27T02:26:49Z DEBUG rustls::client::hs] Not resuming any session
[2020-09-27T02:26:50Z DEBUG rustls::client::hs] Using ciphersuite TLS13_CHACHA20_POLY1305_SHA256
[2020-09-27T02:26:50Z DEBUG rustls::client::tls13] Not resuming
[2020-09-27T02:26:50Z DEBUG rustls::client::tls13] TLS1.3 encrypted extensions: [ServerNameAck]
[2020-09-27T02:26:50Z DEBUG rustls::client::hs] ALPN protocol is None
[2020-09-27T02:26:50Z DEBUG rustls::client::tls13] Ticket saved
[2020-09-27T02:26:50Z DEBUG rustls::client::tls13] Ticket saved
```

It also logs at `trace` details about every TLS packet sent, but I don't think we need to do detailed trace-level logging.

jsha:
Also add log statements to unit. Each request gets one info line;
retries, redirects, and responses get logged at debug level.

Fixes #163

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,12 +39,14 @@ serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
 encoding = { version = "0.2", optional = true }
 native-tls = { version = "0.2", optional = true }
+log = "0.4.11"
 
 [dev-dependencies]
 serde = { version = "1", features = ["derive"] }
 rayon = "1.3.0"
 rayon-core = "1.7.0"
 chrono = "0.4.11"
+env_logger = "0.7.1"
 
 [[example]]
 name = "smoke-test"
--- a/examples/smoke-test/main.rs
+++ b/examples/smoke-test/main.rs
@@ -1,6 +1,5 @@
 use chrono::Local;
 use rayon::prelude::*;
-use rayon_core;
 
 use std::io::{self, BufRead, BufReader, Read};
 use std::iter::Iterator;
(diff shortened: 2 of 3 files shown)


algesten (approved):
Looks good!

We can do a lot more logging. Like protocol level on TRACE etc. But that's up to a future PR.

PR #173, opened 2020-10-01
Add test.sh.

jsha:
This emulates the test matrix that gets run in CI, making it easier to
find failures locally.

There was one conflict in the matrix: when JSON is enabled and TLS is
disabled, two of the doctests would fail. This was previous worked
around as an exclude in the github workflow. I changed the JSON doctest
to use HTTP instead.


algesten (approved):
LGTM!

PR #175, opened 2020-10-02
Allow setting TLS configs on agent.

### LINKED ISSUE #159: set_tls_config should be on Agent
Right now we have `Request.set_tls_config`. I believe this should `Agent.set_tls_config` instead, because TLS settings concern the establishment of connections, and connections are managed by Agent rather than Request. As a concrete example, someone could create an Agent, make one request for `example.com` with a TLS config containing a client certificate, then make another request with a different TLS config not containing the client certificate (that is, removing authentication from the second request). The result would be unpredictable: If a connection got reused, the request would be made on a connection that had a client certificate. But if a new connection was made, the request would be made on a connection without a client certificate.

I propose we add the setting to Agent and deprecate (but not remove) the setting on Request. When we move to 2.0.0 we could actually remove the setting on Request.

jsha:
Clears pool when set.

Fixes #159

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -324,6 +324,7 @@ impl Agent {
     /// Set the TLS client config to use for new connections.
     ///
     /// See [`ClientConfig`](https://docs.rs/rustls/latest/rustls/struct.ClientConfig.html).
+    /// This clears any existing pooled connections.
     ///
     /// Example:
     /// ```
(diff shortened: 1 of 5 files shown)


algesten:
Looks good so far!

algesten (approved):
This looks good!

For 2.0 breaking changes I'd like, at least internally, but maybe also externally, unify the two TLS methods using an enum.

```rust
pub enum Tls {
  Rustls(rustls::ClientConfig),
  NativeTls(native_tls::TlsConnector),
}

impl Tls {
  pub(crate) wrap_stream(stream: impl Stream) -> impl Stream {
   ...
  }
}
```

This way we can clean up a lot of this `#[cfg]` business that is spreading.

jsha:
Closing this in favor of the changes on the release-2.0 branch.

PR #176, opened 2020-10-04
Replace lazy_static! with once_cell Lazy

algesten:
Modern rust code bases prefer once_cell::sync::Lazy over the older
macro based lazy_static.

Notice we still have `lazy_static` in the tree, but only for tests. If we run `cargo tree` without test deps, we are now lazy_static free:

```bash
cargo tree -e normal,build
```


algesten:
@jsha just a small code hygien PR.

jsha (approved):

jsha:
Nice cleanup! 🧹

PR #180, opened 2020-10-06
Use correct host on redirect.

### LINKED ISSUE #179: 404 when downloading a GitHub release
In a project of mine I wrap a pre-built C library in a `-sys` crate (https://github.com/nbigaouette/onnxruntime-rs).

For better ergonomics I download the pre-built library in my build.rs using ureq (still on 1.4.0). This allows running tests against the library+wrapper in my CI (GitHub Actions). See: https://github.com/nbigaouette/onnxruntime-rs/blob/issue22/onnxruntime-sys/build.rs#L80-L87

The archive I want to download is `onnxruntime-linux-x64-1.4.0.tgz` from this release:
https://github.com/microsoft/onnxruntime/releases/tag/v1.4.0

The download used to work, but now it doesn't anymore. I specifically added this download so my GitHub Actions could have access to the library. Here's an example PR where the download passed: https://github.com/nbigaouette/onnxruntime-rs/pull/21

Here's an example of failure: https://github.com/nbigaouette/onnxruntime-rs/runs/1191121056?check_suite_focus=true (part of PR https://github.com/nbigaouette/onnxruntime-rs/pull/23 )

My build script prints the URL (https://github.com/nbigaouette/onnxruntime-rs/blob/issue22/onnxruntime-sys/build.rs#L210-L214) so I copy-pasted and used `curl -LO https://...` and it worked. GitHub releases uses a redirect to point to the real AWS S3 url to download. Maybe something changed in GitHub that broke how `ureq` is handling the redirects?

jsha:
As of 1.5.0, redirects to a different host are broken. This broke as part of #153,
which set the hostname from `req` instead of from `unit.url`.

Fixes #179.

### DIFF
--- a/src/test/redirect.rs
+++ b/src/test/redirect.rs
@@ -1,3 +1,9 @@
+use std::{
+    io::{self, Write},
+    net::TcpStream,
+};
+use test::testserver::{self, TestServer};
+
 use crate::test;
 
 use super::super::*;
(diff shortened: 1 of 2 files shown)


algesten (approved):
Good catch!

PR #182, opened 2020-10-08
agent: Clone resolver in `build()`

rawler:
The code `ureq::agent().set_resolver(...).build()` surprisingly uses the default resolver.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -65,9 +65,6 @@ pub(crate) struct AgentState {
 }
 
 impl AgentState {
-    fn new() -> Self {
-        Self::default()
-    }
     pub fn pool(&mut self) -> &mut ConnectionPool {
         &mut self.pool
     }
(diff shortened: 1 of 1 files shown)


rawler:
I am not sure this is the right way to solve it.
 - Perhaps `resolver` should have belonged to `Agent` all along.
 - Perhaps `build()` is a misnomer (`reset_state`).
 - Perhaps `build` is now such a fuzzy topic that it should not exist at all.

algesten (changes requested):
Good catch!

This is because we've started having config in both `Agent` and `AgentState`.

To avoid missing this if we introduce more conf, I'd like a solution that doesn't rely on `..Default::default()` – to get a type error if we don't make a decision on whether to include a prop or not.

I suggest replace `new` with `AgentState::clone_to_new() -> Self`. And it would enumerate all properties explicitly (not use `Default`), and a code comment explaining why.

algesten:

PR #188, opened 2020-10-17
Add more header validation

jsha:
This adds validation of header values on receive, and of both header names and header values on send. This doesn't change the return type of `set` to be a Result, it just validates when the request is sent. Also removes the section in the README describing handling of invalid headers, and updates a test that verified acceptance of non-ASCII headers so that it verifies rejection of them instead.


algesten (approved):
Also one of those potentially breaking changes. But probably ok.

jsha:
Good point about breaking changes. This isn't particularly urgent, so I changed the PR to target release-2.0 and will merge it there.

Issue #184, opened 2020-10-12
Build pattern on Agent


jsha:
Pulling out some conversation from https://github.com/algesten/ureq/pull/182#issuecomment-705667830:

> Perhaps build is now such a fuzzy topic that it should not exist at all.

@algesten, I'm thinking maybe this is the answer: that Agent shouldn't have a build. I've often found the .build method a bit confusing, since you can accomplish the same thing by just creating an agent and mutating it.

Alternately, we could decide we want to double down on the builder pattern and create a separate AgentBuilder type whose `.build()` returns an `Agent`. This could actually have some advantages, because we could say that certain options (like TLS config and resolver) can only be set once, when an Agent is first created. This saves some awkwardness like clearing the pool when those settings change.

Another thing of note: Right now we have a few setters on Agent that don't follow the chained method approach. We should probably clean that up if we stick with chained methods:

- set_max_pool_connections
- set_max_pool_connections_per_host
- set_cookie

(Also interesting to note, these all take `&self` rather than `&mut self`, which is made possible by interior mutability, but we should probably switch to `&mut self` for clarity and safety).

These are the existing setters that return `&mut Agent` or `&mut Self`:

- set
- auth
- auth_kind
- set_resolver
- set_proxy

Some various resources I looked up in trying to think about this:

https://randompoison.github.io/posts/returning-self/
https://doc.rust-lang.org/1.0.0/style/ownership/builders.html
https://github.com/rust-unofficial/patterns/blob/master/patterns/builder.md
https://github.com/rust-unofficial/patterns/issues/64
https://en.wikipedia.org/wiki/Builder_pattern

/cc @rawler

algesten:
Good observation!

Let's fix this for 2.0. I think we should either do a pure builder pattern or stop song chaining/build.

jsha:
This is mostly complete in the `release-2.0` branch. One thing that still needs attention: `AgentBuilder.build()` is currently infallible. I think we should probably make it a `Result<Agent, Error>` for future proofing, in case we add some future field that might cause errors at build time. For instance, we might get a set_cookie call that is invalid in some way. [Currently CookieStore discards errors](https://docs.rs/cookie_store/0.12.0/src/cookie_store/cookie_store.rs.html#59-70), but we might want to do something different someday.

jsha:
Thoughts on whether we should make `AgentBuilder.build()` fallible even though it doesn't currently fail? I'm leaning towards "what we have now, where it's infallible, is fine."

[...]

PR #190, opened 2020-10-18
Move Agent construction to new AgentBuilder.

jsha:
In the process, rename set_foo methods to just foo, since methods on the
builder will always be setters.

Adds a new() method on ConnectionPool so it can be constructed directly
with the desired limits. Removes the setter methods on ConnectionPool
for those limits. This means that connection limits can only be set when
an Agent is built.

There were two tests that verify Send and Sync implementations, one for
Agent and one for Request. This PR moves the Request test to request.rs,
and changes both tests to more directly verify the traits. There may be
another way to do this, I'm not sure.

Related to #184

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -690,3 +690,17 @@ fn no_hostname() {
     );
     assert!(req.get_host().is_err());
 }
+
+#[test]
+fn request_implements_send_and_sync() {
+    let _request: Box<dyn Send> = Box::new(Request::new(
+        &Agent::default(),
+        "GET".to_string(),
+        "https://example.com/".to_string(),
+    ));
+    let _request: Box<dyn Sync> = Box::new(Request::new(
+        &Agent::default(),
+        "GET".to_string(),
+        "https://example.com/".to_string(),
+    ));
+}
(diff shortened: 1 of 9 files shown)


src/request.rs
```diff
@@ -690,3 +690,17 @@ fn no_hostname() {
...
+    let _request: Box<dyn Send> = Box::new(Request::new(
+        &Agent::default(),
+        "GET".to_string(),
+        "https://example.com/".to_string(),
+    ));
+    let _request: Box<dyn Sync> = Box::new(Request::new(
+        &Agent::default(),
+        "GET".to_string(),
+        "https://example.com/".to_string(),
+    ));
```
algesten, line 705: I've learnt a better pattern for this using an empty body generic `fn`.

```rust
  fn ensure_send(_v: impl Send) {}
  fn ensure_sync(_v: impl Sync) {}

#[test]
 fn request_implements_send_and_sync() {
    ensure_send(Request::new(…));
    ensure_sync(Request::new(…));
 }
```

algesten (commented):
Looks good!

algesten (approved):

src/request.rs
jsha: Ooh, that's a nice technique!

PR #193, opened 2020-10-19
Push mutexes down into pool and cookie store.

jsha:
Previously, Agent stored most of its state in one big
Arc<Mutex<AgentState>>. This separates the Arc from the Mutexes.
Now, Agent is a thin wrapper around an Arc<AgentState>. The individual
components that need locking, ConnectionPool and CookieStore, now are
responsible for their own locking.

There were a couple of reasons for this. Internal components that needed
an Agent were often instead carrying around an Arc<Mutex<AgentState>>.
This felt like the components were too intertwined: those other
components shouldn't have to care quite so much about how Agent is
implemented. Also, this led to compromises of convenience: the Proxy on
Agent wound up stored inside the `Arc<Mutex<AgentState>>` even though it
didn't need locking. It was more convenient that way because that was
what Request and Unit had access too.

The other reason to push things down like this is that it can reduce
lock contention. Mutations to the cookie store don't need to lock the
connection pool, and vice versa. This was a secondary concern, since I
haven't actually profiled these things and found them to be a problem,
but it's a happy result of the refactoring.

Now all the components outside of Agent take an Agent instead of
AgentState.

In the process I removed `Agent.cookie()`. Its API was hard to use
correctly, since it didn't distinguish between cookies on different
hosts. And it would have required updates as part of this refactoring.
I'm open to reinstating some similar functionality with a refreshed API.

I kept `Agent.set_cookie`, but updated its method signature to take a
URL as well as a cookie.

Many of ConnectionPool's methods went from `&mut self` to `&self`,
because ConnectionPool is now using interior mutability.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "ureq"
 version = "1.5.1"
-authors = ["Martin Algesten <martin@algesten.se>"]
+authors = ["Martin Algesten <martin@algesten.se>", "Jacob Hoffman-Andrews <ureq@hoffman-andrews.com>"]
 description = "Minimal HTTP request library"
 license = "MIT/Apache-2.0"
 repository = "https://github.com/algesten/ureq"
--- a/examples/smoke-test/main.rs
+++ b/examples/smoke-test/main.rs
@@ -61,7 +61,7 @@ fn get_and_write(agent: &ureq::Agent, url: &String) -> Result<()> {
 }
 
 fn get_many(urls: Vec<String>, simultaneous_fetches: usize) -> Result<()> {
-    let agent = ureq::Agent::default().build();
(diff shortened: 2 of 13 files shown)


algesten (approved):
One thing, why is it `CookieStoreLocked`? If we're saying that the locked-ness is an internal detail, shouldn't it be just `ColkirStore` which also makes it analogous to the pool which also has an internal lock.

jsha:
Good point, will change that in the morning. I wanted it to have a name that was distinct from the `CookieStore` we import, so someone looking at a call site and not the corresponding `use` statement would know what one they were looking at. Maybe `CookieTin`? :-D

algesten:
`CookieTin`!!!! 🎉🎉🎉🎉

Issue #196, opened 2020-10-20
Timeout is reset on stream when calling `into_reader()`


birkenfeld:
When calling `into_reader()` on a request, the timeout set with `set_timeout_read` (NOT the deadline `set_timeout`) on the request is reset to no timeout by `unit::connect()` calling `DeadlineStream::reset` before the socket is returned, and there is no way to set it again.

This makes `set_timeout_read` basically unused except for reading the headers.

birkenfeld:
In fact, it seems that `timeout_read` also has a bug when a read timeout appears in the header part.

Below is a reproducing example (using version 1.5.1 here).

```
use std::{net, thread, time, io::prelude::*};

fn main() {
    thread::spawn(server);
    thread::sleep(time::Duration::from_millis(100));

    let x = ureq::get("http://localhost:12345/").timeout_read(100).call();
    println!("Response string: {:?}", x.into_string());
}

const WAIT_HEAD: time::Duration = time::Duration::from_millis(0);
const WAIT_BODY: time::Duration = time::Duration::from_millis(0);

fn server() {
    let listener = net::TcpListener::bind(("localhost", 12345)).unwrap();
    let (mut sock, _) = listener.accept().unwrap();
    let mut buf = [0; 128];
    sock.read(&mut buf).unwrap();

    thread::sleep(WAIT_HEAD);
    sock.write_all(b"HTTP/1.1 200 OK\r\nContent-Length: 10\r\n").unwrap();
    sock.write_all(b"Connection: close\r\nContent-Type: text/html\r\n\r\n").unwrap();

    thread::sleep(WAIT_BODY);
    sock.write_all(b"xxxxxxxxxx").unwrap();
}
```

If this is run as given, with no sleeps, everything looks fine:
```
Response string: Ok("xxxxxxxxxx")
```

If run with `WAIT_BODY` set to 1000 milliseconds:
```
Response string: Ok("xxxxxxxxxx")
```
instead of a timeout error.

If run with `WAIT_HEAD` set to 1000 milliseconds:
```
Response string: Ok("HTTP/1.1 200 OK\r\nContent-Length: 10\r\nConnection: close\r\nContent-Type: text/html\r\n\r\nxxxxxxxxxx")
```

jsha:
Thanks for the report, and the functional repro case! I'm looking into it.

jsha:
[...]

PR #197, opened 2020-10-21
Check error immediately after Response::from_read

jsha:
This is a fix for the issue reported in the second comment at https://github.com/algesten/ureq/issues/196#issuecomment-712863514. Note that it's a draft because this breaks at least one existing test: `overall_timeout_during_headers`. That's because previously, the overall timeout was getting preserved on the stream, and `into_string()` was returning an I/O error. With this fix, the stream is not getting assigned to the Response at all, and so `into_string()` is successfully reading the synthetic error body.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -221,6 +221,11 @@ impl Response {
         &self.error
     }
 
+    // Internal-only API, to allow unit::connect to return early on errors.
+    pub(crate) fn into_error(self) -> Option<Error> {
+        self.error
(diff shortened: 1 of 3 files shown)


algesten:
Thanks for this! If it isn't solved by the weekend, I'll have time to pitch in then.

jsha:
Fixed test; ready for review.

algesten (approved):
Good fix!

PR #200, opened 2020-10-23
Remove some dev-dependencies.

jsha:
We don't really need rayon for the parallelism in smoke-test.

Also, now that we use log, smoke-test can just use that.


algesten (approved):
✨

PR #201, opened 2020-10-25
API changes for 2.0

algesten:
* Remove Request::build
* All mutations on Request follow builder pattern

The previous `build()` on request was necessary because mutating
functions did not follow a proper builder pattern (taking `&mut self`
instead of `mut self`). With a proper builder pattern, the need for
`.build()` goes away.

* All Request body and call methods consume self

Anything which "executes" the request will now consume the `Request`
to produce a `Result<Response>`.

* Move all config from request to agent builder

Timeouts, redirect config, proxy settings and TLS config are now on
`AgentBuilder`.

* Rename max_pool_connections -> max_idle_connections
* Rename max_pool_connections_per_host ->  max_idle_connections_per_host

Consistent internal and external naming.

* Introduce new AgentConfig for static config created by builder.

`Agent` can be seen as having two parts. Static config and a mutable
shared state between all states. The static config goes into
`AgentConfig` and the mutable shared state into `AgentState`.

* Replace all use of `Default` for `new`.

Deriving or implementing `Default` makes for a secondary instantiation
API.  It is useful in some cases, but gets very confusing when there
is both `new` _and_ a `Default`. It's especially devious for derived
values where a reasonable default is not `0`, `false` or `None`.

* Remove feature native_tls, we want only native rustls.

This feature made for very clunky handling throughout the code. From a
security point of view, it's better to stick with one single TLS API.
Rustls recently got an official audit (very positive).

https://github.com/ctz/rustls/tree/master/audit

Rustls deliberately omits support for older, insecure TLS such as TLS
1.1 or RC4. This might be a problem for a user of ureq, but on balance
not considered important enough to keep native_tls.

* Remove auth and support for basic auth.

The API just wasn't enough. A future reintroduction should at least
also provide a `Bearer` mechanism and possibly more.

* Rename jar -> cookie_store
* Rename jar -> cookie_tin

Just make some field names sync up with the type.

* Drop "cookies" as default feature

The need for handling cookies is probably rare, let's not enable it by
default.

* Change all feature checks for "cookie" to "cookies"

The outward facing feature is "cookies" and I think it's better form
that the code uses the official feature name instead of the optional
library "cookies".

* Keep `set` on Agent level as well as AgentBuilder.

The idea is that an auth exchange might result in a header that need
to be set _after_ the agent has been built.


algesten:
@jsha Got a mega-PR for you here. I initially tried to keep it in different commits, but things got _very involved_ with this much changes. I've tried to clearly document all the changes though.

jsha:
This is very exciting, thanks! Reviewing now.

src/agent.rs
```diff
@@ -296,27 +331,125 @@ impl AgentBuilder {
...
-        Some(idx) => &user[..idx],
-        None => user,
-    };
-    base64::encode(&format!("{}:{}", safe, pass))
+#[cfg(feature = "tls")]
+#[derive(Clone)]
+pub(crate) struct TLSClientConfig(pub(crate) Arc<rustls::ClientConfig>);
+
+#[cfg(feature = "tls")]
+impl std::fmt::Debug for TLSClientConfig {
```
jsha, line 449: Nit: elsewhere in the codebase we do `use std::fmt;` and `impl fmt::Debug`; I'd like to do that here to be consistent.

src/pool.rs
```diff
@@ -9,8 +9,8 @@ use crate::Proxy;
 
 use url::Url;
 
-const DEFAULT_MAX_IDLE_CONNECTIONS: usize = 100;
-const DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST: usize = 1;
+pub const DEFAULT_MAX_IDLE_CONNECTIONS: usize = 100;
```
jsha, line 12: Let's actually move these values over into `agent.rs`, since notionally it's the Agent that provides the defaults now.

src/request.rs
```diff
@@ -146,11 +118,12 @@ impl Request {
     /// }
     /// ```
     #[cfg(feature = "json")]
-    pub fn send_json(&mut self, data: SerdeValue) -> Result<Response> {
-        if self.header("Content-Type").is_none() {
-            self.set("Content-Type", "application/json");
+    pub fn send_json(self, data: SerdeValue) -> Result<Response> {
+        let mut this = self;
```
jsha, line 122: You can avoid the `let mut this = self` by declaring self to be `mut` in the method signature:

```
pub fn send_json(self, data: SerdeValue) -> Result<Response> {
```

src/test/agent_test.rs
```diff
@@ -166,17 +164,17 @@ fn dirty_streams_not_returned() -> Result<(), Error> {
...
         stream.write_all(b"\r\n")?;
         Ok(())
     });
     let url = format!("http://localhost:{}/", testserver.port);
-    let agent = Agent::default();
+    let agent = Agent::new();
     let resp = agent.get(&url).call()?;
     let resp_str = resp.into_string()?;
-    assert_eq!(resp_str, "corgidachsund");
+    assert_eq!(resp_str, "corgidachshund");
```
jsha, line 177: Thank you for the fix! 😁

src/test/testserver.rs
```diff
@@ -57,12 +57,12 @@ impl TestServer {
                     eprintln!("testserver: handling just-accepted stream: {}", e);
                     break;
                 }
-                thread::spawn(move || handler(stream.unwrap()));
-                if done.load(Ordering::Relaxed) {
+                if done.load(Ordering::SeqCst) {
```
jsha, line 60: I didn't see a a mention of this change in the PR description. What was it for? I'm curious about the choice of `SeqCst` vs `Relaxed`; I didn't give it much thought when I originally wrote this code.

jsha (approved):
Overall looks great. I had some minor comments, but none blocking. I'd like to merge, and then you or I can follow up with additional PRs addressing those comments.

src/agent.rs
```diff
@@ -75,19 +86,47 @@ pub struct Agent {
...
-    pub(crate) jar: CookieTin,
+    #[cfg(feature = "cookies")]
+    pub(crate) cookie_tin: CookieTin,
     pub(crate) resolver: ArcResolver,
 }
 
 impl Agent {
+    /// Creates an Agent with default settings.
+    ///
+    /// Same as `AgentBuilder::new().build()`.
```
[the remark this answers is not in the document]
algesten: Agree!

src/agent.rs
```diff
@@ -175,25 +227,29 @@ impl AgentBuilder {
...
+                #[cfg(feature = "cookies")]
+                cookie_tin: CookieTin::new(
+                    self.cookie_store.unwrap_or_else(|| CookieStore::default()),
                 ),
-                proxy: self.proxy.clone(),
-                #[cfg(feature = "cookie")]
-                jar: CookieTin::new(self.jar),
                 resolver: self.resolver,
             }),
+            config,
```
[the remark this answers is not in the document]
algesten: Agree. But borrow rules won't let me :(

src/agent.rs
algesten: Sure. I opted for this because it means conditional on the tls feature. But I don't feel strongly.

src/pool.rs
algesten: Ok!

src/request.rs
algesten: Can you reassign self?

src/test/agent_test.rs
algesten: :) most important

src/test/testserver.rs
algesten: I got intermittent test errors saying connection closed. I couldn't repro consistently and decided for a belts are bracers.

src/agent.rs
jsha: Oh! I missed that. Is it possible to do:

```rust
#[cfg(feature = "tls")]
{
  use std::fmt;
  impl fmt::Debug for TLSClientConfig {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
  ...
}
```

src/request.rs
jsha: I believe so! Haven't double-checked; I could be wrong.

src/test/testserver.rs
jsha: I'm working on merging the recent timeout fixes into `release-2.0` and am also getting connection closed errors. I believe these are actually caused because `dribble_headers_respond` does not properly finish its headers with `\r\n`, so ureq sees that as a connection that was closed while headers were still being read. The underlying issue is that `dribble_headers_respond` shouldn't be getting to the end of writing its headers at all; we should time out before then. I've fixed the test so it emits a clearer error in that case, and am working on figuring out why it's reaching the end of headers (i.e. why it's not timing out).

Issue #202, opened 2020-10-25
why requires 'static in send() from a reader 


keepsimple1:
Hi, 

I am trying to use [`send()`](https://docs.rs/ureq/1.5.1/ureq/struct.Request.html#method.send) , and wondering why it requires `'static` since the reader is already owned ?  Thanks.

jsha:
Fixed in #205! Thanks for the report @keepsimple1 :-)

PR #205, opened 2020-10-25
Make Request::send more general.

steffahn:
Removes `+ 'static` constraint from the `impl Read` parameter.
For this, lifetime parameters are added to `Payload` and `SizedReader`.

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -17,11 +17,11 @@ use super::SerdeValue;
 /// *Internal API*
 pub(crate) enum Payload<'a> {
     Empty,
-    Text(String, String),
+    Text(&'a str, String),
     #[cfg(feature = "json")]
     JSON(SerdeValue),
     Reader(Box<dyn Read + 'a>),
-    Bytes(Vec<u8>),
+    Bytes(&'a [u8]),
 }
 
 impl fmt::Debug for Payload<'_> {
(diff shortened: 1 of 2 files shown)


jsha:
Welcome to the repository, and thanks for the patch! I assume this is related to #202? Can you give some more background and an example of the problem you're trying to solve? I think this is probably good but I'd like to understand the problem space better.

steffahn:
> Can you give some more background and an example of the problem you're trying to solve?

Sure, I’m coming from [this URLO thread](https://users.rust-lang.org/t/static-lifetime-with-owned-object/50637) where I originally planned on explaining to OP why the `'static` constraint is needed here. I found it _isn’t_ really needed, so I made this PR.

I guess it is related to #202 since that’s the same person as OP in the linked URLO thread.

Since there are `Read` impls in `std` for reference types like [`&'_ u8`](https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-10) or [`&'_ mut R` where `R: Read`](https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-14), I think it is a valuable addition to the API to start supporting those types.

jsha (approved):
Excellent, thanks for the context. That helps me understand a lot better. By the same token, we could (in a separate change) turn the `Bytes(Vec<u8>),` variant into `Bytes(&'a [u8])`, saving a copy.

This looks good to me. I'd like to wait on @algesten's review too. In the meantime, would you mind adding a short unittests that calls `send` on a reader built from ` &'_ u8`, both to demonstrate the functionality and to ensure we don't break it in the future? Thanks!

steffahn:
I’m not sure where exactly the unit test is supposed to go.

Also, I’m going offline for today, it’s getting late here. I’ll probably check back in tomorrow. Edits by maintainers are allowed on this PR, so feel free to make a unit test yourself if you don’t want to wait for me :wink:.

jsha:
Sounds good, have a good night!

algesten (approved):
This is a very good change. It's like a 2-in-1 with both fixing the 'static lifetime and the `&[u8]`.

Thanks!

PR #211, opened 2020-10-29
Agent tweaks for 2.0

jsha:
Replace `this` idiom with `mut self`.

Move idle connections constants from pool.rs to agent.rs.

Remove Agent.set and some convenience request methods
(leaving get, post, and put).

Move max_idle_connections setting from AgentConfig to AgentBuilder
(since the builder passes these to ConnectionPool and the Agent
doesn't subsequently need them).

Eliminate duplicate copy of proxy in AgentState; use the one in
AgentConfig.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -14,6 +14,8 @@ use {crate::cookies::CookieTin, cookie::Cookie, cookie_store::CookieStore, url::
 pub struct AgentBuilder {
     headers: Vec<Header>,
     config: AgentConfig,
+    max_idle_connections: usize,
+    max_idle_connections_per_host: usize,
     /// Cookies saved between requests.
     /// Invariant: All cookies must have a nonempty domain and path.
     #[cfg(feature = "cookies")]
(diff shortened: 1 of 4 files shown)


src/agent.rs
```diff
@@ -157,11 +133,6 @@ impl Agent {
         self.request("GET", path)
     }
 
-    /// Make a HEAD request from this agent.
-    pub fn head(&self, path: &str) -> Request {
-        self.request("HEAD", path)
-    }
-
```
algesten, line 164: I think we should keep HEAD. It's useful against S3 to check for existence.

src/agent.rs
```diff
@@ -171,35 +142,16 @@ impl Agent {
     pub fn put(&self, path: &str) -> Request {
         self.request("PUT", path)
     }
-
-    /// Make a DELETE request from this agent.
-    pub fn delete(&self, path: &str) -> Request {
-        self.request("DELETE", path)
-    }
-
```
algesten, line 179: I think we should keep DELETE. since without it one can't use a classic REST API (CRUD)

src/request.rs
```diff
@@ -118,12 +118,11 @@ impl Request {
     /// }
     /// ```
     #[cfg(feature = "json")]
-    pub fn send_json(self, data: SerdeValue) -> Result<Response> {
-        let mut this = self;
-        if this.header("Content-Type").is_none() {
-            this = this.set("Content-Type", "application/json");
+    pub fn send_json(mut self, data: SerdeValue) -> Result<Response> {
+        if self.header("Content-Type").is_none() {
+            self = self.set("Content-Type", "application/json");
```
algesten, line 123: very nice!

algesten (commented):
Looks good! I want to keep HEAD and DELETE but otherwise I agree.

Issue #203, opened 2020-10-25
Should Agent have a set method as well as AgentBuilder?


jsha:
Pulling out a conversation from https://github.com/algesten/ureq/pull/201: 

> Keep set on Agent level as well as AgentBuilder.
>
> The idea is that an auth exchange might result in a header that need
to be set after the agent has been built.

I'd like to approach this use case a different way:

```rust
let builder = ureq::AgentBuilder::new()
  .set("Some-Header", "foo"));
let agent = builder.build();
let resp = agent.get("https://example.com/fetch-token")?;
let authed_agent = builder
  .set("Token", resp.header("Token").unwrap())
  .build();
let resp2 = agent.post("https://example.com/authed_endpoint")?;
```

In other words, make `AgentBuilder.build()` not consume self. Alternately, we could offer an `AgentBuilder.clone()`, but that is a little more awkward.

Here's my reasoning: We're intentionally making an Agent really easy to share across threads by cloning it. We can protect the Agent's internal state with a Mutex so we don't have memory-level race conditions. But if an Agent can have headers added after construction, we could have application-level race conditions: One thread might be performing an auth handshake while another is making a request, and that request would be authed or not depending on the random ordering of the two. By requiring the headers to be set at build time, we make it easier to reason about any given Agent: it's authed, or it's not authed, and that won't change.

I have to acknowledge, of course, that the above logic could equally well apply to auth cookies, but the way cookies work means we kind of need to allow mutation of an Agent's cookie jar during its lifetime.

Also, if we allow setting headers on an already-built Agent, we will probably eventually get requests to allow removing headers from an already-built Agent; it has the potential to get quite complicated. By keeping our API surface for setting headers to two places (AgentBuilder and RequestBuilder) we can hopefully keep things simple.

The main thing standing in the way of the code snippet I posted above is that `AgentBuilder.build()` _does_ consume the builder. That's because `CookieStore` is not cloneable. We could ask upstream about adding a clone or (my preference for now) we could manually implement a clone for our `CookieTin` wrapper by copying out all the cookies.

algesten:
Not sure I entirely agree here. I admit I have put the `headers` vec in a bad place, however, `Agent` is absolutely a mutable shared state. Breaking the builder pattern + potentially requiring all conf in the builder to be clonable, sounds on balance worse to me.

Like, if we are reusing connections and have a mutable cookie store, why not some mutable headers?

As long as we make the contract clear: headers are copied into `Request` when the request is created.

[...]

PR #213, opened 2020-10-30
Release 2.0 agent headers and cookie tweaks

algesten:
These are the tweaks I'm talking about in #203 

It also provides the functions needed to implement an external cookie persistence.


jsha (approved):
Nice!

Issue #126, opened 2020-08-03
Error should be cloneable


sergiimk:
Hi, I was giving this library a try but found error handling extremely painful.

I'd like to be able to compose errors (for example re-emit `ureq::Error::ConnectionFailed` as `myapp::Error::ResourceUnavailable`) while preserving the original error as `source` not to lose any detailed information for debugging.

Library currently prevents any composition because `ureq::Error` does not implement `Clone` and `Response.synthetic_error()` returns `&Option<Error>`. This means I can inspect the error but neither move it out of the response nor clone it.

Any attempt to box it via `Box::new(e)` or `e.into()` results in rather confusing error message:
```
Box::new(e)
^^^^^^^^^^^ the trait `std::error::Error` is not implemented for `&ureq::error::Error`
```

I'd also recommend taking a look at `thiserror` crate as it can fill some gaps in your `Error` implementation (e.g. providing `Error.source()` for `Io(IoError)`) while maintaining backwards compatibility.

jsha:
Thanks for the report! As a quick fix, what do you think of adding `impl std::error::Error for &Error {}` in addition to the current `impl std::error::Error for Error {}`? I thnk that would allow a boxed `ureq::Error` to serve as a source.

BurntSushi:
I don't think implementing `std::error::Error` for `&Error` would help much here, because it would imply returning a borrowed error as a source and `source` has a `'static` bound. I do think that enabling composition is pretty important. I think you have a few choices available to you to do this:

* Make your error type cloneable. (This may be difficult if your error type embeds an `io::Error`.)
* Provide an `into_synthetic_error` API that consumes the response and returns a `Result<Response, Error>`, where the return value is the response itself if there is no error and the error itself otherwise.
* Provide some kind of mutable API that permits one to extract an error out of a response if it exists. e.g., I think `&mut Option<ureq::Error>` would do the trick, because then callers could use [`Option::take`](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) to transfer ownership. However, this is a bit unwieldy, and it's not clear what role a response that errored but has no error would play.
* My guess is that #128 would also somewhat provide a way out of this, but it would be a little weird to require the use of a convenience API just to enable error composition.

Personally, I think an `into_synthetic_error` is probably the best option here.

algesten:
@BurntSushi 

[...]

PR #214, opened 2020-11-06
Add Response::into_synthetic_error

jyn514:
This is an alternative to https://github.com/algesten/ureq/pull/132 which does not require breaking changes.

Closes https://github.com/algesten/ureq/issues/126.

Note that this is a backport to the 1.5.0 branch, so that I can use the new API without having to wait for a 2.0 release.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -213,14 +213,34 @@ impl Response {
         self.error.is_some()
     }
 
-    /// Get the actual underlying error when the response is
+    /// Get a reference to the actual underlying error when the response is
     /// ["synthetic"](struct.Response.html#method.synthetic).
     pub fn synthetic_error(&self) -> &Option<Error> {
         &self.error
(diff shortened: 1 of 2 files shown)


jyn514:
Hmm, it looks like 1.5.0 is missing a lot of things from 1.5.1 ... I'm not sure how to make a backport. Only the last commit that github shows here is by me.

jsha:
I changed the base branch to `master`, since we're still using `master` for the 1.x series. I like the idea for this change! Looks like you've got a conflict in `src/response.rs` though; mind cleaning that up?

jyn514:
Heh, the conflict was that you'd already added this API, but for internal use only. I changed it to public and kept my name (it was `into_error()` before). Let me know if you like `into_error` better.

jsha (approved):
Nice!

PR #217, opened 2020-11-11
Split out RequestBuilder from Request.

jsha:
This formalizes the builder pattern with regards to a Request. All
methods like `ureq::get` that formerly returned a Request now return a
RequestBuilder, so you can add headers and params the same as before.
Also, RequestBuilder implements `call()` and `send*()` as shortcuts to
immediately build and send a Request, so call chains don't get longer.

Add the ability to set a URL on a request directly, in addition to the
ability to pass an unparsed `&str`. Remove some of the individual field
getters from Request, like get_method, get_url, get_scheme, and get_host.

Remove the `qstring` dependency. We can achieve the same thing
using Url's query_pairs_mut.

Delete the `.proxy()` convenience method on Request.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ base64 = "0.13"
 chunked_transfer = "1.2.0"
 cookie = { version = "0.14", features = ["percent-encode"], optional = true}
 once_cell = "1"
-qstring = "0.7"
 url = "2"
 socks = { version = "0.3.2", optional = true }
 rustls = { version = "0.18", optional = true, features = [] }
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use crate::pool::ConnectionPool;
 use crate::proxy::Proxy;
-use crate::request::Request;
+use crate::request::RequestBuilder;
 use crate::resolve::{ArcResolver, StdResolver};
 use std::time::Duration;
 
@@ -106,32 +106,35 @@ impl Agent {
     ///     .call();
     /// println!("{:?}", r);
     /// ```
-    pub fn request(&self, method: &str, path: &str) -> Request {
-        Request::new(self.clone(), method.into(), path.into())
+    pub fn request(&self, method: &str, url: &str) -> RequestBuilder {
+        RequestBuilder::new()
+            .agent(self.clone())
+            .method(method)
+            .url_str(url)
     }
 
     /// Make a GET request from this agent.
-    pub fn get(&self, path: &str) -> Request {
+    pub fn get(&self, path: &str) -> RequestBuilder {
         self.request("GET", path)
     }
 
     /// Make a HEAD request from this agent.
-    pub fn head(&self, path: &str) -> Request {
+    pub fn head(&self, path: &str) -> RequestBuilder {
         self.request("HEAD", path)
     }
 
     /// Make a POST request from this agent.
-    pub fn post(&self, path: &str) -> Request {
+    pub fn post(&self, path: &str) -> RequestBuilder {
         self.request("POST", path)
     }
 
     /// Make a PUT request from this agent.
-    pub fn put(&self, path: &str) -> Request {
+    pub fn put(&self, path: &str) -> RequestBuilder {
         self.request("PUT", path)
     }
 
     /// Make a DELETE request from this agent.
-    pub fn delete(&self, path: &str) -> Request {
+    pub fn delete(&self, path: &str) -> RequestBuilder {
         self.request("DELETE", path)
     }
 
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -330,7 +330,7 @@ impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
             stream.reset()?;
 
             // insert back into pool
-            let key = PoolKey::new(&unit.url, unit.req.proxy());
+            let key = PoolKey::new(&unit.url, unit.req.agent.config.proxy.clone());
             unit.req.agent.state.pool.add(key, stream);
         }
 
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,7 +1,6 @@
 use std::fmt;
-use std::io::Read;
+use std::io::{self, Read};
 
-use qstring::QString;
 use url::{form_urlencoded, Url};
 
 use crate::agent::Agent;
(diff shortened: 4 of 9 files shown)


Cargo.toml
```diff
@@ -28,7 +28,6 @@ base64 = "0.13"
 chunked_transfer = "1.2.0"
 cookie = { version = "0.14", features = ["percent-encode"], optional = true}
 once_cell = "1"
-qstring = "0.7"
```
algesten, line 31: Good! I like getting rid of this dep!

src/agent.rs
```diff
@@ -106,32 +106,35 @@ impl Agent {
     ///     .call();
     /// println!("{:?}", r);
     /// ```
-    pub fn request(&self, method: &str, path: &str) -> Request {
-        Request::new(self.clone(), method.into(), path.into())
+    pub fn request(&self, method: &str, url: &str) -> RequestBuilder {
+        RequestBuilder::new()
+            .agent(self.clone())
+            .method(method)
+            .url_str(url)
```
algesten, line 113: The `RequestBuilder` I'm not so sure about, the reasoning being: who are we doing this for?

1. From the user's perspective, I'd like them to interact with two main ureq objects, `Request` and `Response`. However with the `RequestBuilder`, it's like `Request` becomes an almost invisible intermediate object that the user only would interact with if they for some reason `.build()` instead of call it directly. 
2. Optics of the API. `pub fn get(path: &str) -> Request` conveys the simplicity I want ureq to have. I like there is no `Result` and no intermediate object that becomes a mental hurdle to grasp (granted, builders are a common pattern).
3. The Request was already a builder pattern. Not in name, but I feel spelling that out is more a formalism for our benefit than the user's.

What are the arguments for a `RequestBuilder`?

src/pool.rs
```diff
@@ -330,7 +330,7 @@ impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
             stream.reset()?;
 
             // insert back into pool
-            let key = PoolKey::new(&unit.url, unit.req.proxy());
+            let key = PoolKey::new(&unit.url, unit.req.agent.config.proxy.clone());
```
algesten, line 333: :+1:

src/request.rs
```diff
@@ -31,67 +50,288 @@ pub type Result<T> = std::result::Result<T, Error>;
...
+        self.url = Some(Urlish::Url(url));
+        self
+    }
+
+    /// Set the URL to send this request to. If you have an already-parsed Url
+    /// object, use `.url()` instead.
+    pub fn url_str(mut self, url: &str) -> Self {
+        self.url = Some(Urlish::Str(url.to_string()));
+        self
+    }
```
algesten, line 95: The "native" Rust datatype is `&str`, exposing `Url` is for me secondary. I would like that to be reflected in the naming. As an aside, one reason I didn't like the Agent::set_cookie with `Url` was that it forces the user to get their head around `Url`, when the rest of `ureq` is about `&str`. 

Since the URL is introduced in `ureq::get` or `agent::get`, I think it's a somewhat strange API to have override methods for that on the builder level (regardless of whether the builder is `Request` or `RequestBuilder`). How would the user use it?

```rust
let url = Url::parse("http://foo.com").unwrap();
ureq::get("ignore me").url(url).call();
```

src/request.rs
```diff
@@ -31,67 +50,288 @@ pub type Result<T> = std::result::Result<T, Error>;
...
+        }
+        let req = Request {
+            agent,
+            method,
+            url,
+            return_error_for_status: self.return_error_for_status,
+            headers: self.headers,
+        };
+        Ok(req)
+    }
```
algesten, line 194: The validation here is good.

algesten (changes requested):
Some comments, let's discuss!

src/request.rs
algesten: Maybe we can do something like this to support uses of `Url`?

```rust
impl Agent {
    pub fn request_url(&self, method: &str, url: Url) -> Request {
…
}

impl Request {
  pub fn new_with_url(method: &str, url: Url) -> Request {
…
}
```

jsha:
> The RequestBuilder I'm not so sure about, the reasoning being: who are we doing this for?

> From the user's perspective, I'd like them to interact with two main ureq objects, Request and Response. However with the RequestBuilder, it's like Request becomes an almost invisible intermediate object that the user only would interact with if they for some reason .build() instead of call it directly.
> Optics of the API. pub fn get(path: &str) -> Request conveys the simplicity I want ureq to have. I like there is no Result and no intermediate object that becomes a mental hurdle to grasp (granted, builders are a common pattern).
> The Request was already a builder pattern. Not in name, but I feel spelling that out is more a formalism for our benefit than the user's.

Good questions indeed! My idea for this change came from what we did with Agent, splitting it into AgentBuilder and Agent. To me, that made the relationship between the two clearer, and each of their APIs smaller and cleaner. I expected a similar benefit here. However, I agree with you that making users think about two request types is not great.

One of the advantages for us as authors is that a RequestBuilder can have lots of things that aren't quite ready to actually perform a request, or they might be one thing or another. For instance, a Urlish might be a `Url` or a `&str`, but we make sure to get to a `Url` before we perform the Request. Or some fields might be invalid, and we validate them before performing the request. However, convenience for the developers is not a good reason to choose an API that is worse for users.

I'll close out this PR and split it into a few smaller PRs to:
 - Get rid of the qstring dependency.
 - Get rid of the proxy() convenience method.
 - Offer methods that take in a Url alongside the methods that take in a `&str`.

And what would you think of defining an internal-only `BuiltRequest` struct? Then in `do_call`, we would create that BuiltRequest, doing the validation and collapsing the various options. Then in internal things like `Unit` we would make reference to BuildRequest instead of Request, so we could rely on things being in a validated state.

algesten:
> * Get rid of the qstring dependency.
> * Get rid of the proxy() convenience method.
> * Offer methods that take in a Url alongside the methods that take in a &str.

Perfect!

>And what would you think of defining an internal-only BuiltRequest struct? Then in do_call, we would create that BuiltRequest, doing the validation and collapsing the various options. Then in internal things like Unit we would make reference to BuildRequest instead of Request, so we could rely on things being in a validated state.

That sounds ideal to me. `Request` becomes a de-facto builder pattern (without being called builder), but the actual built artefact is not something the user is exposed to.

PR #218, opened 2020-11-11
Add support for using testserver in doctests.

jsha:
Doctests run against a normally-built copy of the crate, i.e. one
without #[cfg(test)] set, so we can't use the conditional compilation
feature.

Instead, define a static var that indicates whether the library is
running in test mode or not. For each doctest, insert a hidden call that
sets this var to true. Then, when ureq::agent() is called, it returns a
test_agent instead.

This required moving testserver out of the test mod and into src/, so
that it can be included unconditionally (i.e. when cfg(test) is false).

This PR converts one doctest as an example. If we land this PR, I'll
send a followup to convert the rest.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,8 @@ pub use serde_json::json;
 
 #[cfg(test)]
 mod test;
+#[doc(hidden)]
+mod testserver;
 
 pub use crate::agent::Agent;
 pub use crate::agent::AgentBuilder;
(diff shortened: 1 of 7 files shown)


jsha:
Odd; this is consistently failing in CI, but the same command is consistently failing locally. I'll look in more detail tomorrow.

algesten:
Looks good! I tried a fix for the failing tests here, https://github.com/algesten/ureq/tree/jsha-doctest-override
but got even more absurd errors (SIGILL, really?!)

jsha:
Thanks for kicking off that branch! That was exactly the next approach I wanted to try. It looks like the SIGILL is a result of "thread panicked while panicking; aborting". That was caused by the `.unwrap()` in the Drop implementation. I removed that and went from reliably failing locally to reliably passing locally. I also removed the conn_count stuff because it was breaking other tests, and I think it's harmless for the first connection to hit the normal handler.

jsha:
Looks like localhost is not `127.0.0.1` in GitHub Actions (some of the time?). I changed TestServer to bind to 127.0.0.1, and it made the tests pass. 🎉 !

algesten (approved):
Super!!!

PR #222, opened 2020-11-14
Implement more realistic doctests.

jsha:
Add `is_test` and `fn main` headers to various doctests, and use real
URLs along with the `?` operator.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -304,6 +304,8 @@ impl AgentBuilder {
     /// If both this and `.timeout()` are both set, `.timeout()`
     /// takes precedence.
     ///
+    /// The default is no timeout. In other words, requests may block forever on reads by default.
+    ///
     /// ```
     /// use std::time::Duration;
     /// # fn main() -> Result<(), ureq::Error> {
@@ -323,6 +325,8 @@ impl AgentBuilder {
     /// Timeout for the individual writes to the socket.
     /// If both this and `.timeout()` are both set, `.timeout()`
     /// takes precedence.
+    ///
+    /// The default is no timeout. In other words, requests may block forever on writes by default.
     ///
     /// ```
     /// use std::time::Duration;
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -151,8 +151,8 @@ impl Agent {
     /// agent.get("https://www.google.com/").call()?;
     ///
     /// // Saves (persistent) cookies
-    /// let file = File::create("cookies.json")?;
-    /// agent.cookie_store().save_json(&mut file)?;
+    /// let mut file = File::create("cookies.json")?;
+    /// agent.cookie_store().save_json(&mut file).unwrap();
     /// # Ok(())
     /// # }
     /// ```
@@ -418,7 +418,7 @@ impl AgentBuilder {
     /// with some pre-existing cookies.
     ///
     /// Example
-    /// ```
+    /// ```no_run
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
     /// use cookie_store::CookieStore;
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -285,7 +285,7 @@ impl AgentBuilder {
     /// The default is 30 seconds.
     ///
     /// ```
-    /// # use std::time::Duration;
+    /// use std::time::Duration;
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
     /// let agent = ureq::builder()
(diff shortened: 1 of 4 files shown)


src/agent.rs
```diff
@@ -274,10 +285,15 @@ impl AgentBuilder {
     /// The default is 30 seconds.
     ///
     /// ```
+    /// # use std::time::Duration;
```
algesten, line 288: I think we should show this import.

src/agent.rs
```diff
@@ -288,13 +304,16 @@ impl AgentBuilder {
     /// If both this and `.timeout()` are both set, `.timeout()`
     /// takes precedence.
     ///
-    /// The default is `0`, which means it can block forever.
-    ///
```
algesten, line 292: The doc was wrong, but we should say what the default behavior is.

src/agent.rs
```diff
@@ -305,13 +324,16 @@ impl AgentBuilder {
...
```
algesten, line 309: Ditto

src/agent.rs
```diff
@@ -346,12 +372,15 @@ impl AgentBuilder {
     /// If the redirect count hits this limit (and it's > 0), TooManyRedirects is returned.
     ///
     /// ```
-    /// let r = ureq::builder()
-    ///     .redirects(10)
```
algesten, line 350: Can we use redirects(0) to turn off redirects completely?

src/agent.rs
```diff
@@ -362,11 +391,15 @@ impl AgentBuilder {
     ///
     /// Example:
     /// ```
-    /// let tls_config = std::sync::Arc::new(rustls::ClientConfig::new());
+    /// # fn main() -> Result<(), ureq::Error> {
+    /// # ureq::is_test(true);
+    /// # use std::sync::Arc;
```
algesten, line 396: Show this import

src/agent.rs
```diff
@@ -305,13 +324,16 @@ impl AgentBuilder {
     /// If both this and `.timeout()` are both set, `.timeout()`
     /// takes precedence.
     ///
-    /// The default is `0`, which means it can block forever.
-    ///
     /// ```
+    /// # use std::time::Duration;
```
algesten, line 328: Show this import

src/agent.rs
```diff
@@ -288,13 +304,16 @@ impl AgentBuilder {
...
```
algesten, line 308: Show this import

src/agent.rs
```diff
@@ -381,22 +414,24 @@ impl AgentBuilder {
...
-    /// use cookie_store::CookieStore;
-    /// use std::fs::File;
-    /// use std::io::BufReader;
-    ///
-    /// let file = File::open("cookies.json").unwrap();
+    /// ```
+    /// # fn main() -> Result<(), ureq::Error> {
+    /// # ureq::is_test(true);
+    /// # use cookie_store::CookieStore;
+    /// # use std::fs::File;
```
algesten, line 421: Show these imports

src/request.rs
```diff
@@ -62,13 +64,16 @@ impl Request {
     /// Use `.timeout_connect()` and `.timeout_read()` to avoid blocking forever.
     ///
     /// ```
-    /// let r = ureq::builder()
-    ///     .timeout_connect(std::time::Duration::from_secs(10)) // max 10 seconds
+    /// # use std::time::Duration;
```
algesten, line 67: show

src/request.rs
```diff
@@ -194,13 +208,14 @@ impl Request {
...
     ///
     /// ```
-    /// use std::io::Cursor;
-    ///
-    /// let read = Cursor::new(vec![0x20; 100_000]);
-    ///
-    /// let resp = ureq::post("http://localhost/example-upload")
-    ///     .set("Content-Type", "text/plain")
-    ///     .send(read);
+    /// # use std::io::Cursor;
```
algesten, line 211: show

src/response.rs
```diff
@@ -213,22 +227,23 @@ impl Response {
     /// Example:
     ///
     /// ```
-    /// # #[cfg(feature = "tls")] {
-    /// use std::io::Read;
-    ///
-    /// let resp =
-    ///     ureq::get("https://ureq.s3.eu-central-1.amazonaws.com/hello_world.json")
-    ///         .call().unwrap();
+    /// # use std::io::Read;
```
algesten, line 230: show

algesten (approved):
Very good!✨

Overall I don't think we should hide imports of std stuff. I want ureq to be accessible also for beginners who don't know what a BufRead or a Duration is.

jsha:
Good call on showing the imports. Do you have thoughts on showing them as a separate line (good style) vs using fully-qualified names inline (shorter examples)?

algesten:
I think it's cleaner to have them on separate lines separated by one blank line to the actual example. In some respects I want the examples to be copy-paste into a user's code and I think one should use imports (although I'm a bit lax on it myself).

PR #223, opened 2020-11-14
Turn Unit into a built Request

jsha:
This involved removing the Request reference from Unit, and adding an
Agent, a method, and headers.

Also, move is_retryable to Unit.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -330,8 +330,8 @@ impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
             stream.reset()?;
 
             // insert back into pool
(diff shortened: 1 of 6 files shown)


algesten (approved):
So Unit effectively becomes out BuiltRequest?

jsha:
> So Unit effectively becomes out BuiltRequest?

Yep! I realized we already had a type that was filling 90% of the role, and it was simplest to just take it to 100%. If we land this, we may want to rename Unit to BuiltRequest or ReadyRequest or similar; but I figured that would just distract from the changes here.

algesten:
Sounds good!

PR #224, opened 2020-11-14
Remove Headers from the public API.

jsha:
It turns out Headers is actually an internal-only API. None of the
user-facing types use it.

Unfortunately, making it unexported also required deleting the doctests,
since doctests can only run against a public interface.


jsha:
Ping on this (if you have time :-))

algesten (approved):
Agree. Sad to lose doc tests. Can be changed into unit tests, but happy to merge as is.

jsha:
Good point about losing the doctests. I'll change those to unittests before merging.

PR #226, opened 2020-11-14
Add ureq::request_url and Agent::request_url.

jsha:
These let a user pass an already-parsed Url.

An open question: Should we `pub use url::Url`? My inclination is to not do so. The documentation that results doesn't make it clear that `Url` comes from a different crate. As a beginner, I was very confused about why each HTTP package seemed to have its own implementation of `Url`. The downside is that the user has to be using a version selector for `url::Url` that is compatible with the version we are using (and then Cargo's maximal version selection should take care of the rest). If they're using an incompatible version they get a pretty clear error message from Rust:

```
error[E0308]: mismatched types
  --> src/main.rs:49:45
   |
49 |     let response = agent.request_url("GET", &u).call()?;
   |                                             ^^ expected struct `url::Url`, found struct `Url`
   |
   = note: expected reference `&url::Url`
              found reference `&Url`
   = note: perhaps two different versions of crate `url` are being used?
```


algesten:
> An open question: Should we pub use url::Url? My inclination is to not do so.

I agree. Also because to provide access to all the knobs, we probably had to expose `pub use url`, in which case the surface API of that crate will overshadow that of ureq.

jsha:
FYI, I'm waiting on approval of this PR if you have time to take a look.

algesten (approved):
Sorry missed it.

Yes. This looks great.

PR #230, opened 2020-11-15
bikeshed: rename error_for_status -> error_on_non_2xx

algesten:
From a user's perspective, reading "Request:: error_for_status" feels unclear. "Error for what status?" "What is status?"

I suggest we should rename this to "error_on_non_2xx". It's a bit unwieldy, but maybe clearer what it means?

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -31,7 +31,7 @@ pub struct Request {
     agent: Agent,
     method: String,
     url: String,
-    return_error_for_status: bool,
+    error_on_non_2xx: bool,
     headers: Vec<Header>,
     query_params: Vec<(String, String)>,
 }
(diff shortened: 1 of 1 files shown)


jsha (approved):
This is fine by me, and I agree the current name isn't great. I'll join you a bit in bikeshedding and mention that 3xx errors complicate this, since we generally follow those rather than error.

Also one of the things that's been bugging me about this API is that the default is true, and you call `error_on_non_2xx(false)` to change it. Somehow it feels like it would be more intuitive if the method name said the thing you are changing it to do. For instance `allow_4xx` / `allow_5xx` or `accept_4xx` / `accept_5xx`.

At any rate, I don't have strong feeling about this and I am `200 OK` with the approach in this PR.

algesten:
>  I am `200 OK` with the approach in this PR.

Wihoo! `FIN`

PR #233, opened 2020-11-16
Update README and docs.

jsha:
This makes src/lib.rs the primary source for crate-level documentation.
I've generated README.md with `cargo readme > README.md`. Since links to
specific documentation items should be relative when possible, but must
be absolute in README.md, I've used the new syntax for intra-rustdoc
links
(https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md),
along with a README.tpl that sets up those links to point at the
absolute versions. `cargo readme` uses the README.tpl by default.

I've also rewritten the crate level docs, removing some TODO information
at the bottom, and moving the license information to CONTRIBUTING.md.

Also, update the Cargo.toml keywords and description to make ureq
more findable in searches for "http client."


algesten (approved):
Looks good! I think we should keep one paragraph about sync and why this lib is sync forever. Maybe talking up that there is no need for an async runtime.

jsha:
I wound up adding three paragraphs. :-) Let me know what you think!

algesten (approved):

algesten:
Looks good! Nice to get some depth on why. Small note on form, it's structured _why not async_ -> _why sync_ – which maybe gives it a slightly negative tone. Unsure how hard it is to reverse to start with why sync is good. It's a minor nit, I'm totally happy to merge as is.

jsha:
You're right about the structure, but I haven't been able to figure out a way to phrase it in the positive without first explaining the negative. I'll merge as-is and keep thinking about it; we can always revise later.

PR #234, opened 2020-11-20
Rewrite the Error type.

### LINKED ISSUE #232: URL and method on errors
I was just thinking - it would be useful if our Error type included the URL that was being request, and the method. That way instead of printing, e.g.

```
HTTP status 404
```

We could print

```
HTTP status 404 on GET http://example.com
```

The alternative would be for the caller to add the URL as context, but that would involve duplicating the URL in most cases.

jsha:
This adds a source field to keep track of upstream errors and allow
backtraces, plus a URL field to indicate what URL an error was
associated with.

The enum variants we used to use for Error are now part of a new
ErrorKind type. For convenience within ureq, ErrorKinds can be turned
into an Error with `.new()` or `.msg("some additional information")`.

Error acts as a builder, so additional information can be added after
initial construction. For instance, we return a DnsFailed error when
name resolution fails. When that error bubbles up to Request's
`do_call`, Request adds the URL.

Fixes #232.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,10 +6,11 @@ use std::io::{self};
 
 use crate::Response;
 
+/// An error that may occur when processing a Request.
 #[derive(Debug)]
 pub struct Error {
     kind: ErrorKind,
-    message: String,
+    message: Option<String>,
     url: Option<Url>,
     source: Option<Box<dyn error::Error>>,
     response: Option<Box<Response>>,
@@ -23,7 +24,10 @@ impl Display for Error {
         if let Some(response) = &self.response {
             write!(f, "status code {}", response.status())?;
         } else {
-            write!(f, "{:?}: {}", self.kind, self.message)?;
+            write!(f, "{:?}", self.kind)?;
+        }
+        if let Some(message) = &self.message {
+            write!(f, ": {}", message)?;
         }
         if let Some(source) = &self.source {
             write!(f, ": {}", source)?;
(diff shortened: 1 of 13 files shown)


algesten (approved):
Nice cleanup! And super nice to have Url in there.

src/error.rs
```diff
@@ -1,25 +1,111 @@
...
+        Error {
+            kind,
+            message,
+            url: None,
+            source: None,
+            response: None,
+        }
+    }
+
+    pub(crate) fn url(mut self, url: Url) -> Self {
```
algesten, line 52: In two minds about exposing `Url` as a type here. Since our entry point is mainly `&str` 🤔

src/error.rs
```diff
@@ -31,44 +117,60 @@ pub enum Error {
...
-    // Return true iff the error was due to a connection closing.
-    pub(crate) fn connection_closed(&self) -> bool {
-        match self {
-            Error::Io(e) if e.kind() == ErrorKind::ConnectionAborted => true,
-            Error::Io(e) if e.kind() == ErrorKind::ConnectionReset => true,
-            _ => false,
-        }
+impl ErrorKind {
+    pub(crate) fn new(self) -> Error {
+        Error::new(self, "".to_string())
```
algesten, line 125: Should we maybe have `Optional<String>` in `Error`?

src/error.rs
jsha: In this case it's only exposed at the crate level, so doesn't affect the public-facing API; we can always change it later.

PR #235, opened 2020-11-21
Merge into_json_deserialize into into_json.

jsha:
Followup to #56. At the time, doing this would have been an API break; but we can do this as part of 2.0. This simplifies the API nicely and creates better symmetry between sending bodies and receiving them.


algesten (approved):
Nice! So much cleaner.

PR #236, opened 2020-11-22
Make Error Send + Sync

jsha:


algesten (approved):
Good catch! So easy to forget.

PR #245, opened 2020-11-23
Cleanups and coverage improvements.

jsha:
I looked at some grcov output locally, and found some code that was
unused: The BufRead impl on Stream, and the Debug and Default impls on
Payload. Deleted those.

Also, the timeout_write doctest on Agent was testing the wrong thing.
Fixed that.

Added some test cases to exercise Debug impls.

Used the Display formatter rather than the Debug formatter to write
ErrorKind in Error's Display impl.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -366,7 +366,7 @@ impl AgentBuilder {
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
     /// let agent = ureq::builder()
-    ///     .timeout_read(Duration::from_secs(1))
+    ///     .timeout_write(Duration::from_secs(1))
     ///     .build();
     /// let result = agent.get("http://httpbin.org/delay/20").call();
     /// # Ok(())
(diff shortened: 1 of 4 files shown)


algesten (commented):
This seems strange. Isn't that BufRead implicatitly used when using it as a Read Trait?

Doesn't this bring us back to many many syscalls?

jsha:
I will double-check on the syscalls, but I'm pretty sure not. The BufReader inside Stream takes care of that, I think: https://github.com/algesten/ureq/blob/master/src/stream.rs#L27-L35

algesten (approved):
Ok! When you're certain just go ahead and merge.

algesten:
Please redo this PR against "main".

jsha:
> This seems strange. Isn't that BufRead implicatitly used when using it as a Read Trait?

FYI, what I found in #261 was that the BufRead methods (`fill_buf` and `consume`) weren't getting called because we weren't actually using the BufRead aspects of our reader. Implementing the `BufRead` trait doesn't automatically make the `.read()` method use the `BufRead` methods. For BufReader, `.read()` is specifically written so that it calls `.fill_buf()` and then `.consume()`. For other structs that want to act like a BufReader, they need to do the same. That's what I wound up doing for DeadlineStream in #261.

Issue #187, opened 2020-10-16
Redirect 307 / 308 correctly


oberien:
This feature was already tried to be implemented. However, when sending the request, the body is consumed (due to being `dyn Read`), so it can't be sent again for the 307 or 308 redirect, which requires the same request to be sent again.

https://github.com/algesten/ureq/blob/5b75deccef13f1ceb2b8d546a735c0d3bc3bfc83/src/unit.rs#L237-L238

From what I can tell so far after looking at the source, we can reset any `Payload` back to the beginning, except the generic `Reader(Box<dyn Read + 'static>)`. If this reader also implemented `Seek`, we would be able to get its position before starting the body-write, and reset it to that position on a 307 / 308 to read from it again. However, this would be a breaking API change to `Request::send(&mut self, reader: impl Read + 'static) -> Response` as the argument now also needs to implement `Seek`. (Additionally, `io::Empty` currently doesn't implement `Seek`, for which I opened https://github.com/rust-lang/rust/issues/78029.)  
Another option would be to require the reader to be `Clone`, however I'd see that as a strictly worse version of `Seek`.

jsha:
Thanks for filing, and welcome to the repo! I've been wanting to get this done for a while. It sounds like your proposal has two parts:

 - For all input types other than `Reader(Box<dyn Read + 'static>)`, just reset in the straightforward way.
 - Add another input type like `Reader(Box<dyn Read + Seek + 'static>)` so that certain Read inputs (like files) can be seek'ed to reset.

Is that right?

algesten:
Can we solve this with a buffer of configurable size? (This is an idea @jsha started)

The default buffer size would be some multiple of a probable initial TCP window size, meaning the server would have a chance to send the 307/308 response while the client sending more than the buffer size is "held back" by TCP flow control.

Obviously _probable initial TCP window size_ is a bit moot, so the user should be able to configure this for `Agent` and/or `Request`.

In addition, such a buffer could means for small `Read` bodies, we can pre-buffer the entire contents into memory, avoiding transfer-encoding chunked.

oberien:
> For all input types other than `Reader(Box<dyn Read + 'static>)`, just reset in the straightforward way.

[...]

PR #252, opened 2020-11-23
Follow 307/308 redirects

jyn514:
This is the simplest part of https://github.com/algesten/ureq/issues/187. I'm not sure what the expect-100 discussion was about, but maybe it can be fixed later?

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -408,14 +408,24 @@ impl AgentBuilder {
     ///
     /// If the redirect count hits this limit (and it's > 0), TooManyRedirects is returned.
     ///
+    /// WARNING: for 307 and 308 redirects, this value is ignored for methods that have a body.
+    /// You must handle 307 redirects yourself when sending a PUT, POST, PATCH, or DELETE request.
+    ///
     /// ```
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
     /// let result = ureq::builder()
     ///     .redirects(1)
     ///     .build()
-    ///     .get("http://httpbin.org/redirect/3")
-    ///     .call();
+    ///     .get("http://httpbin.org/status/301")
+    ///     .error_on_non_2xx(false)
+    ///     .call()?;
+    /// assert_ne!(result.status(), 301);
+    ///
+    /// let result = ureq::post("http://httpbin.org/status/307")
+    ///     .error_on_non_2xx(false)
+    ///     .send_bytes(b"some data")?;
+    /// assert_eq!(result.status(), 307);
     /// # Ok(())
     /// # }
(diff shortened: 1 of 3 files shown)


algesten:
Thank! I need to refresh my memory on why we disabled 307/308. I suspect it's because the spec requires us to resend any body, which in turn means buffer potentially a lot of data to be able to "replay", or break spec.

jyn514:
Another option is to make this work *most* of the time by only worrying about GET requests. Then there's no body to worry about resending, and only PUT/POST will have to deal with 307 responses manually.

jyn514:
Would love to get some eyes on this :) AFAIK there are a few ways forward:

1. The current situation, don't follow 307/308 and return them to the caller as an error. This is the most flexible for the caller but also the least ergonomic, especially since ureq doesn't currently document that 307 isn't handled.
2. Handle 307/308, but only for GET/HEAD/DELETE (any method without a body). This handles the common case while putting off a decision about bodies for later. However it makes the API pretty inconsistent, which is unfortunate.
3. Handle 307/308 on PUT by resending the request. This requires bounding the API to require `Seek`, not just `Read`, which is more restrictive for the caller but prevents spurious copies. Personally I think requiring the caller to use `read_to_end` first is not a giant deal, but I also don't use ureq for PUT requests so my opinion doesn't matter very much :P
4. possibly other ideas ???

While writing this up I realized I don't actually need this to be fixed on the ureq end to fix https://github.com/deadlinks/cargo-deadlinks/issues/110, I can handle the 307/308 explicitly on my end.

jsha:
Hi, sorry for the delay in looking at this! I have been meaning to swap back into my head what the semantics of 307/308 are. I think this is the best option for a short term fix:

> Handle 307/308, but only for GET/HEAD/DELETE (any method without a body). This handles the common case while putting off a decision about bodies for later. However it makes the API pretty inconsistent, which is unfortunate.

Are you interested in updating this PR to take this approach?

jyn514:
Yes, definitely! I'll add that tomorrow.

jsha:
FYI, you're getting a test failure because we do something tricky to make our doctests not hit live websites. If you look at the hidden `is_test(true)` at the top of the doctest, it sets of a test agent that overrides DNS lookups to point at a TestServer listening on localhost. We have a set of handlers in testserver.rs for various known test URLs. To make your doctest work you'll need to add a handler for /308.

Why, then, do we use httpbin as an example URL? If someone copies and pastes the code, we'd like it to work on the real Internet too.

jyn514:
@jsha thanks, that fixed it!

algesten (approved):
I'm happy to go ahead with this change. Thanks both for the work!

PR #254, opened 2020-11-24
Replace unmaintained encoding crate with encoding_rs

messense:
https://github.com/hsivonen/encoding_rs


algesten (approved):
All good! Let's merge!

PR #255, opened 2020-11-26
Add Error::{kind, status, into_response}.

jsha:
Also, remove Response::{ok, error, client_error, server_error,
redirect}. The idea is that you would access the status code through
the Error object instead.

I fetched all the reverse dependencies of ureq on crates.io and looked
for uses of the methods being removed. I found none.

I'm also considering removing the error_on_non_2xx method entirely. If
it's easy to get the underlying response for errors, it would be nice to
make that the single way to do things rather than support two separate
ways of handling HTTP errors.

Followup from @jyn514's comment at https://github.com/algesten/ureq/issues/128#issuecomment-732470836.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -21,14 +21,18 @@ use crate::Response;
 /// use ureq::{Response, Error};
 /// # fn main(){ ureq::is_test(true); get_response(); }
 ///
+/// // An example of a function that handles HTTP 500 errors differently
+/// // than other errors.
 /// fn get_response() -> Result<Response, Error> {
-///   let mut result = ureq::get("http://httpbin.org/status/500").call();
+///   let fetch = || ureq::get("http://httpbin.org/status/500").call();
+///   let mut result = fetch();
 ///   for _ in 1..4 {
 ///     match result {
+///       // Retry 500's after waiting for two seconds.
 ///       Err(e) if e.status() == 500 => thread::sleep(Duration::from_secs(2)),
 ///       r => return r,
 ///     }
-///     result = ureq::get("http://httpbin.org/status/500").call();
+///     result = fetch();
 ///   }
 ///   println!("Failed after 5 tries: {:?}", &result);
 ///   result
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,6 +7,32 @@ use std::io::{self};
 use crate::Response;
 
 /// An error that may occur when processing a Request.
+///
+/// This can represent connection-level errors (e.g. connection refused),
+/// protocol-level errors (malformed response), or status code errors
+/// (e.g. 404 Not Found). For status code errors, kind() will be
+/// ErrorKind::HTTP, status() will return the status code, and into_response()
+/// will return the underlying Response. You can use that Response to, for
+/// instance, read the full body (which may contain a useful error message).
+///
+/// ```
+/// use std::{result::Result, time::Duration, thread};
+/// use ureq::{Response, Error};
+/// # fn main(){ ureq::is_test(true); get_response(); }
+///
+/// fn get_response() -> Result<Response, Error> {
+///   let mut result = ureq::get("http://httpbin.org/status/500").call();
+///   for _ in 1..4 {
+///     match result {
+///       Err(e) if e.status() == 500 => thread::sleep(Duration::from_secs(2)),
+///       r => return r,
+///     }
+///     result = ureq::get("http://httpbin.org/status/500").call();
+///   }
+///   println!("Failed after 5 tries: {:?}", &result);
+///   result
+/// }
+/// ```
 #[derive(Debug)]
 pub struct Error {
     kind: ErrorKind,
--- a/src/error.rs
+++ b/src/error.rs
@@ -13,7 +13,7 @@ pub struct Error {
     message: Option<String>,
     url: Option<Url>,
     source: Option<Box<dyn error::Error + Send + Sync + 'static>>,
-    response: Option<Box<Response>>,
+    response: Option<Response>,
 }
 
 impl Display for Error {
@@ -67,11 +67,51 @@ impl Error {
     }
 
     pub(crate) fn response(mut self, response: Response) -> Self {
-        self.response = Some(Box::new(response));
+        self.response = Some(response);
         self
     }
-    pub(crate) fn kind(&self) -> ErrorKind {
+
+    /// The type of this error.
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// let err = ureq::get("http://httpbin.org/status/500")
+    ///     .call().unwrap_err();
+    /// assert_eq!(err.kind(), ureq::ErrorKind::HTTP);
+    /// ```
+    pub fn kind(&self) -> ErrorKind {
         self.kind
+    }
+
+    /// For Errors of type HTTP (i.e. those that result from an HTTP status code),
+    /// return the status code of the response. For all other Errors, return 0.
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// let err = ureq::get("http://httpbin.org/status/500")
+    ///     .call().unwrap_err();
+    /// assert_eq!(err.kind(), ureq::ErrorKind::HTTP);
+    /// assert_eq!(err.status(), 500);
+    /// ```
+    pub fn status(&self) -> u16 {
+        match &self.response {
+            Some(response) => response.status(),
+            None => 0,
+        }
+    }
+
+    /// For an Error of type HTTP (i.e. those that result from an HTTP status code),
+    /// turn the error into the underlying Response. For other errors, return None.
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// let err = ureq::get("http://httpbin.org/status/500")
+    ///     .call().unwrap_err();
+    /// assert_eq!(err.kind(), ureq::ErrorKind::HTTP);
+    /// assert_eq!(err.into_response().unwrap().status(), 500);
+    /// ```
+    pub fn into_response(self) -> Option<Response> {
+        self.response
     }
 
     /// Return true iff the error was due to a connection closing.
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,14 +6,15 @@ use std::io::{self};
 
 use crate::Response;
 
-/// An error that may occur when processing a Request.
+/// An error that may occur when processing a [Request](crate::Request).
 ///
 /// This can represent connection-level errors (e.g. connection refused),
 /// protocol-level errors (malformed response), or status code errors
-/// (e.g. 404 Not Found). For status code errors, kind() will be
-/// ErrorKind::HTTP, status() will return the status code, and into_response()
-/// will return the underlying Response. You can use that Response to, for
-/// instance, read the full body (which may contain a useful error message).
+/// (e.g. 404 Not Found). For status code errors, [kind()](Error::kind()) will be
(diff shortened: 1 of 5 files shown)


src/error.rs
```diff
@@ -67,13 +67,53 @@ impl Error {
...
+    /// # ureq::is_test(true);
+    /// let err = ureq::get("http://httpbin.org/status/500")
+    ///     .call().unwrap_err();
+    /// assert_eq!(err.kind(), ureq::ErrorKind::HTTP);
+    /// ```
+    pub fn kind(&self) -> ErrorKind {
         self.kind
     }
 
+    /// For Errors of type HTTP (i.e. those that result from an HTTP status code),
```
jyn514, line 86:
```suggestion
    /// For [`Error`]s of type HTTP (i.e. those that result from an HTTP status code),
```
or maybe
```suggestion
    /// For errors of type HTTP (i.e. those that result from an HTTP status code),
```

src/error.rs
```diff
@@ -67,13 +67,53 @@ impl Error {
...
+    /// return the status code of the response. For all other Errors, return 0.
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// let err = ureq::get("http://httpbin.org/status/500")
+    ///     .call().unwrap_err();
+    /// assert_eq!(err.kind(), ureq::ErrorKind::HTTP);
+    /// assert_eq!(err.status(), 500);
+    /// ```
+    pub fn status(&self) -> u16 {
```
jyn514, line 96: Maybe return an `Option` instead? Or an `Option<NonZeroU16>`?

jyn514 (commented):
> The idea is that you would access these through the Error object instead.

Maybe this should be documented somewhere? I had to think about why they'd only be relevant for an `Error`: if it succeeded, it will always be `ok()` and all the other functions will return `false`.

jyn514:
I just tried it out and this works great :) Thanks so much for putting thought into this API! https://github.com/jyn514/cargo-deadlinks/commit/4a7739523c7040ccd89ec9d3994ffa245ba7ce52

src/error.rs
jsha: Oh interesting! I did not know about NonZeroU16. I went back and forth about whether to use Option instead of having 0 as a sentinel value. I think doing it this way simplifies calling code a lot, since callers are usually checking for specific values rather than trying to handle every possible outcome. For instance, with the sentinel value:

```
let res = ureq::get("http://example.com").call();
if let Err(e) = res {
  if e.status() > 500 {
    retry();
  }
}
```

vs (with Option):

```
let res = ureq::get("http://example.com").call();
if let Err(e) = res {
  if let Some(status) = e.status() {
    if status > 500 {
      retry();
    }
  }
}
```

Or, a caller could simplify that with:

```
let res = ureq::get("http://example.com").call();
if let Err(e) = res {
  if e.status().unwrap_or_default() > 500 {
    retry();
  }
}
```

But if we expect the caller to always call `unwrap_or_default`, we could save them the trouble!

jsha:
> Maybe this should be documented somewhere? I had to think about why they'd only be relevant for an Error: if it succeeded, it will always be ok() and all the other functions will return false.

Good point. Also I updated the wording of the PR description a bit to say:

>  The idea is that you would access the status code through the Error object instead.

Since I didn't move the affected methods to Error but deleted them entirely.

src/error.rs
jyn514: Ok, if the focus is for ergonomics I agree a sentinel value is nicer :)

jsha:
Here's an alternative to into_response I considered just now and rejected:

```rust
    /// Returns Ok(Response) if this Error is of type HTTP (i.e. it results from an
    /// HTTP status code). Otherwise return Err(self). Consumes self.
    ///
    /// This allows overriding ureq's default behavior of treating all 4xx and 5xx
    /// status codes as errors.
    ///
    /// ```
    /// # ureq::is_test(true);
    /// let response = ureq::get("http://httpbin.org/status/500")
    ///     .call()
    ///     .or_else(|e| e.accept_response())?;
    /// if response.status() == 429 || response.status() == 500 {
    ///   thread::sleep(Duration::from_secs(2));
    ///   retry();
    /// } else {        
    ///   return MyError::new(format!("response status {}", response.status()))
    /// }
    /// ```
    pub fn accept_response(self) -> Result<Response, Error> {
        if let Some(response) = self.response {
            Ok(response)
        } else {
            Err(self)
        }
    }
```

I like the neatness of being able to do `.or_else(|e| e.accept_response())` as a way of saying "I don't care about the status code, just give me all the Responses." But as the doctest suggests, that's not usually what people want - usually they want to single out specific status codes (like 429 and 500) for special treatment, while leaving other status codes (like 401 and 404) to be errors.

And the disadvantage of the `accept_response` approach is that by accepting all the responses, you lose the niceties of `ureq::Error`: the url and source. One could do:

```rust
let response = ureq::get("http://httpbin.org/status/500")
    .call()
    .or_else(|e| match e.status() {
       429 | 500 => e.accept_response(),
       _ => e
     })?
 if response.status() == 429 || response.status() == 500 {
   ...
```

But that's awkward due to duplicating the `429` and `500`. Another possibility to consider: a variant of `accept_response` that takes a list of statuses to accept.

src/error.rs
algesten: Yeah. I like sentinel.

src/error.rs
```diff
@@ -67,13 +98,53 @@ impl Error {
...
+    /// For an Error of type HTTP (i.e. those that result from an HTTP status code),
+    /// turn the error into the underlying Response. For other errors, return None.
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// let err = ureq::get("http://httpbin.org/status/500")
+    ///     .call().unwrap_err();
+    /// assert_eq!(err.kind(), ureq::ErrorKind::HTTP);
+    /// assert_eq!(err.into_response().unwrap().status(), 500);
+    /// ```
```
algesten, line 143: Could we get rid of this Optional? 

I know there are some "early" errors like validating the URL which means constructing a Response could be awkward.

I'm just thinking ergonomics.

src/response.rs
```diff
@@ -147,30 +147,6 @@ impl Response {
...
-    /// Whether the response status is: 500 <= status <= 599
-    pub fn server_error(&self) -> bool {
-        self.status >= 500 && self.status <= 599
-    }
-
-    /// Whether the response status is: 400 <= status <= 599
-    pub fn error(&self) -> bool {
-        self.client_error() || self.server_error()
-    }
-
```
algesten, line 173: Why did these go?

algesten (commented):
I like where this is going. If we can make it ergonomic enough, I agree we should drop error_on_non_2xx.

src/error.rs
jsha: There are actually a lot of error types for which a response is unavailable: All of the Io ones, DnsFailed, ConnectionFailed, BadStatus, BadHeader, etc. I think you're suggesting that for those cases, if someone called into_response(), we'd construct a fake Response with e.g. status 500. I would prefer to avoid that - it mixes up the error layers more than I'd like.

The optional is indeed a bit unfortunate because it encourages checking the ErrorKind and then doing `.unwrap()`. That's safe, but it requires a little reasoning to see its safety and might cause the user to ignore other instances of `.unwrap()`. That said, short of constructing fake Responses, I don't see a way around it.

src/response.rs
jsha: As @jyn514 pointed out at https://github.com/algesten/ureq/issues/128#issuecomment-732470836, it's confusing to have e.g. `.client_error()` on Response, because the main way you get a Response is via `Ok(response)`, in which case you know ``client_error()` will always be false.

The other way you can get a Response is via `err.into_response()`, but if you are calling `.into_response()` you've probably already checked the status code to see whether it's one you're interested in.

The other answer is that I checked all the published crates for users of these and found they're not used at all. People prefer `.status()`. I figure, let's head into 2.0.0 with a clean slate and a narrow API; if we find there's user demand for methods like `.client_error()`, we can always add them back, but we can't delete them if we launch with them.

jsha:
By the way, @jyn514, I realized that the variant I showed above, where `accept_response` returns a `Result`:

    pub fn accept_response(self) -> Result<Response, Error> {

Was exactly what you originally proposed in https://github.com/algesten/ureq/issues/128#issuecomment-732470836. Sorry for not acknowledging that!

jyn514:
No problem! The main reason I suggested that is it means you don't have to check `kind()` beforehand: you can call `err.into_http_error()` directly, and if it returns `Ok()`, you know the `kind` must have been `HTTP`. Like you noticed above, the current setup with an `Option` requires an `unwrap()`.

> And the disadvantage of the accept_response approach is that by accepting all the responses, you lose the niceties of ureq::Error: the url and source.

Hmm, I don't think the URL is currently exposed in the API - I wish it were, it would make https://github.com/jyn514/cargo-deadlinks/commit/4a7739523c7040ccd89ec9d3994ffa245ba7ce52#diff-d04787b0b6844467601179d863fe4139b6fd271bed6292b73e6374c063637e2aR85 a little nicer instead of having to pass in the URL from the calling function.

jsha:
> > And the disadvantage of the accept_response approach is that by accepting all the responses, you lose the niceties of ureq::Error: the url and source.

> Hmm, I don't think the URL is currently exposed in the API

Ah, I meant that the URL gets printed in the Display impl.

I put up a couple of other alternatives in #256 and #257. Let me know what you think!

jsha:
I keep thinking about this problem, and I'm starting to like the idea of returning `Result`. I think there is one small tweak that would make it quite nice and usable: Implement `From<Response> for Error`, so it's easy for users who extract a Response to turn it back into an Error if they decide not to handle it after all. The resulting Error could get its url field from the Response itself (and we could make Response include a chain of redirect responses).

Here's an example of what that would look like, with some code that retries 429s after a delay:

```rust
impl Error {
    ...
    pub fn into_response(self) -> Result<Response, Error> {
        match self.response {
            Some(r) => Ok(r),
            None => Err(self),
        }
    }
}

impl From<Response> for Error {
    fn from(resp: Response) -> Error {
        ErrorKind::HTTP.new().response(resp).url(resp.get_url())
    }
}
```

```rust
// Fetch a URL. If the response is a 429, retry up to 4 times after
// an appropriate backoff.
fn get_response(agent: &ureq::Agent, url: &str) -> result::Result<Response, ureq::Error> {
    let retry_secs = |r: &Response| {
        r.header("retry-after")
            .and_then(|h| h.parse().ok())
            .unwrap_or(5)
    };
    for i in 1.. {
        let result: Result<Response, Error> = agent.get(url).call();
        if i >= 5 {
            return result;
        }
        // protocol and transport errors get returned immediately.
        let response: Response = result.or_else(|e: Error| e.into_response())?;
        // if the error contained a response, and the response was 429, retry.
        match response.status() {
            429 => {
                let delay = Duration::from_secs(retry_secs(&response));
                eprintln!("429, sleeping for {:?}", delay);
                thread::sleep(delay);
            }
            _ => return Err(response)?,
        }
    }
    unreachable!();
}
```

Another minor variation: instead of defining `fn into_response`, we could define `impl TryFrom<Error> for Response`. That would let users say `result.or_else(|r| r.try_into())?`, which is basically the same. I'm not sure if we'd get any particular benefits from the `impl`, so I'm inclined to avoid it for fear of surprising side effects.

jsha:
Closing in favor of #257

PR #256, opened 2020-11-26
Add `.http()` method to Error to extract HTTP-type errors

jsha:
An alternate approach to #255, meant to reduce the need to call `.unwrap()`. This method returns a `Result<(u16, Response), Error>`.


jsha:
Closing in favor of #257

PR #257, opened 2020-11-27
Make Error an enum: Status(code, response) or Transport

jsha:
This is a third alternative to #255 and #256.

It makes Error an enum with two variants: `Status`, to represent a status-code error (404, 500, etc) and `Transport` to represent all other errors.

This is similar to #255 but has some advantages:
 - It is more explicit about the meaning of the two variants, rather than repurposing Result to express the notion of "either X or Y."
 - The match syntax is somewhat simpler since you don't need an extra layer of parens to indicate a tuple.
 - The caller gets to decide whether they want to move the Response out of an Error or take a reference to the Response.

The main disadvantage that I see is that the variant names are somewhat unwieldy:

```
        match err {
            ureq::Error::Status(429, r) => {
```

This can be solved by importing them, but the bare names `Status` and `Transport` are a bit ambiguous.

### DIFF
--- a/examples/smoke-test/main.rs
+++ b/examples/smoke-test/main.rs
@@ -1,11 +1,11 @@
-use std::io::{self, BufRead, BufReader, Read, Write};
+use std::io::{self, BufRead, BufReader, Read};
 use std::sync::{Arc, Mutex};
 use std::thread::{self, JoinHandle};
 use std::time::Duration;
-use std::{env, error, fmt, result, result::Result};
+use std::{env, error, fmt, result};
 
 use log::{error, info};
-use ureq::{self, Response};
+use ureq;
 
 #[derive(Debug)]
 struct Oops(String);
@@ -30,13 +30,14 @@ impl fmt::Display for Oops {
     }
 }
 
-fn get(agent: &ureq::Agent, url: &str) -> std::result::Result<(), Oops> {
-    let r = get_response(agent, url)?;
-    let mut reader = r.into_reader();
+type Result<T> = result::Result<T, Oops>;
+
+fn get(agent: &ureq::Agent, url: &str) -> Result<Vec<u8>> {
+    let response = agent.get(url).call()?;
+    let mut reader = response.into_reader();
     let mut bytes = vec![];
     reader.read_to_end(&mut bytes)?;
-    std::io::stdout().write_all(&bytes)?;
-    Ok(())
+    Ok(bytes)
 }
 
 fn get_and_write(agent: &ureq::Agent, url: &str) {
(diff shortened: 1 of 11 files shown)


examples/smoke-test/main.rs
```diff
@@ -48,7 +47,31 @@ fn get_and_write(agent: &ureq::Agent, url: &str) {
...
+        match err {
+            ureq::Error::Status(429, r) => {
+                let retry: Option<u64> = r.header("retry-after").and_then(|h| h.parse().ok());
+                eprintln!("429 for {}, retrying in {}", r.status(), r.get_url());
+                thread::sleep(Duration::from_secs(retry.unwrap_or(5)));
+                eprintln!("error status {} body {}", r.status(), r.into_string()?);
+            }
+            ureq::Error::Status(_, r) => return Err(r.into()),
+            ureq::Error::Transport(e) => return Err(e.into()),
+        };
```
algesten, line 67: I think this might be more succinctly expressed as:

```rust
if let ureq::Error::Status(429, r) = err {
 // ... retry
} else {
  return Err(err);
}
```

algesten:
I quite like where this is going!

jyn514:
I ran into a few issues trying to use this:

1. `Transport` is private, so I can't move out of `Error::Transport`:

```
error[E0603]: module `error` is private
   --> src/check.rs:24:26
    |
24  |     HttpFetch(Url, ureq::error::Transport),
    |                          ^^^^^ private module
```

I think this was just an oversight? The struct is marked public, you could re-export it in the root.

2. `` error[E0277]: `ureq::error::Transport` doesn't implement `std::fmt::Display` ``
3. Is there a reason `status` is part of the enum directly? You can get it easily by calling `response.status()`, which also avoids a state mismatch where the status in `Status(code)` doesn't match `response.status()`.

jyn514:
Overall though I like this :) It makes it very clear that there are two types of errors that might need to be handled differently.

jyn514:
If it helps, this is ideally what I'd like the api to be: https://github.com/deadlinks/cargo-deadlinks/pull/129/files#diff-d04787b0b6844467601179d863fe4139b6fd271bed6292b73e6374c063637e2aR248
where it returns a `Result` and the `Error` distinguishes between status errors and IO errors (and the whole thing handles 307/308 redirects, too). So this is on the right track for that :)

algesten:
@jyn514 Thanks for that input. I've got the fixes, but since I didn't open the original PR, I'll merge this as is and follow up.

jsha:
> Is there a reason status is part of the enum directly? You can get it easily by calling response.status(), which also avoids a state mismatch where the status in Status(code) doesn't match response.status().

I did it this way for ergonomics on match statements. I figure:

```
            Err(Status(503, r) => ...
```

is easier to type and to read than:

```
            Err(Status(r)) is r.status() = 503 =>
```

birkenfeld:
Very nice, and being able to match the status this easily really makes `error_on_non_2xx` quite obsolete. Thanks!

PR #258, opened 2020-11-28
Remove status methods on Response.

jsha:
Now that Responses with non-2xx statuses get turned into `Error`, there is less need for these. Also, surveying the set of public crates that depend on ureq, none of them use these methods. It seems that users tend to prefer checking the status code directly.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -121,7 +121,7 @@ impl Request {
         let response =
             unit::connect(unit, true, 0, reader, false).map_err(|e| e.url(url.clone()))?;
 
-        if response.error() && self.error_on_non_2xx {
+        if self.error_on_non_2xx && response.status() >= 400 {
             Err(ErrorKind::HTTP.new().url(url.clone()).response(response))
         } else {
             Ok(response)
--- a/src/response.rs
+++ b/src/response.rs
@@ -147,30 +147,6 @@ impl Response {
             .collect()
     }
 
-    /// Whether the response status is: 200 <= status <= 299
-    pub fn ok(&self) -> bool {
-        self.status >= 200 && self.status <= 299
-    }
-
-    pub fn redirect(&self) -> bool {
-        self.status >= 300 && self.status <= 399
-    }
-
-    /// Whether the response status is: 400 <= status <= 499
-    pub fn client_error(&self) -> bool {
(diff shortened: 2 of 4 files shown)


jyn514 (approved):

algesten (commented):
In my view these are higher level API. I think they are useful also with the changes to Error.

algesten:
For context, they were lifted from superagent.

jsha:
Here is my thinking on each of these individually:

`.ok()` -- With the new Result API, any Request you get back will by `.ok()`. Also, I think the name `.ok()` is a little confusing with `Result::ok()`.

`.error()` - with the new Result API, this is an exact overlap with anything that would return Error. People will just check for whether a Result is `Err(...)` rather than call `.error()`.

`.client_error()` - most of the time, if someone wants to specially handle a 4xx error, they want to handle specific ones, because the response to them is different. For instance a specialized response to a 404 would be "delete this from the list of URLs to check in the future," where a specialized response to a 401 would be "try and load updated credentials." For instance:

https://github.com/msfjarvis/healthchecks-rs/blob/4200edb9ed73bed56a05d81cdd3b0a65badc7ede/healthchecks/src/manage.rs#L70-L84

https://github.com/SirWindfield/cargo-free/blob/75d4b363b63fa2b6f3ac12a196a2cb80f0afc848/src/lib.rs#L59-L63

https://github.com/lukehsiao/netlify-ddns-rs/blob/1d7daea38bf78d9f1bbfba9d38ea1863ed70ed1d/src/netlify.rs#L101-L112

`.server_error()` - I don't have as much objection to this one, since it's reasonable to want to treat all server errors (500, 502, 503) more or less the same. Although even at that, [501 Not Implemented](https://tools.ietf.org/html/rfc7231#section-6.6.2) seems like people would want to handle it differently. I guess that doesn't come up much in practice - I've never seen a 501 in the wild.

`.redirect()` - Usually redirects are handled under the hood, unless someone disables automatic redirect handling. I'm not terribly opposed to this one, but given that no-one's using it and it's just as easy to do `300..399.contains(resp.status())`, I'm mildly inclined towards deletion.

algesten:
Alright. Let's drop them.

PR #259, opened 2020-11-28
De-redundantize Error kinds.

jsha:
Change "Bad" to "Invalid" in error names, mimicking io::Error::ErrorKind.

Change InvalidProxyCreds to ProxyUnauthorized.

Change DnsFailed to just Dns (the fact that there was a failure is implicit
in the fact that this was an error).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -109,9 +109,9 @@ pub enum ErrorKind {
     /// Too many redirects.
     TooManyRedirects,
     /// A status line we don't understand `HTTP/1.1 200 OK`.
-    Status,
+    BadStatus,
     /// A header line that couldn't be parsed.
-    Header,
+    BadHeader,
     /// Some unspecified `std::io::Error`.
     Io,
     /// Proxy information was not properly formatted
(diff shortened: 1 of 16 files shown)


src/error.rs
```diff
@@ -99,29 +99,27 @@ impl Error {
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub enum ErrorKind {
     /// The url could not be understood.
-    BadUrl,
+    Url,
```
algesten, line 102: InvalidUrl?

algesten (commented):
In two minds about this.

I think the archetype is `std::io::ErrorKind`, which very clearly has stuff like `InvalidInput`, `InvalidData`.

"Bad" might not be a good word. "InvalidUrl" seems ok.

jsha:
Good point about io::Error having Invalid. I've changed to InvalidUrl. Are there any others that you think should have `Invalid`? E.g. InvalidStatus or InvalidHeader?

jyn514:
'InvalidStatus' seems to mean the status is itself invalid, like a status of 0 or something. I think 'ErrorStatus' or something would be more descriptive.

jsha:
> 'InvalidStatus' seems to mean the status is itself invalid, like a status of 0 or something. I think 'ErrorStatus' or something would be more descriptive.

Actually, the ErrorKind `InvalidStatus` (or `Status` in this PR) reflects a failure to parse the status line - for instance a status of 0. For this PR, I'm going to revert it to `BadStatus` just so we don't have lots of conflicts with #266. Later, though, we should change this to merge BadStatus and BadHeader into just one ErrorKind, `InvalidHeader`. That avoids any potential confusion with `error::Error::Status(u16, Response)`.

algesten (approved):
Let's do it!

PR #261, opened 2020-11-29
Use BufRead in more places

jsha:
This moves Stream's enum into an `Inner` enum, and wraps a single BufReader around the whole thing. This makes it easier to consistently treat the contents of Stream as being wrapped in a BufReader.

Also, implement BufRead for DeadlineStream. This means when a timeout is set, we don't have to set that timeout on the socket with every small read, just when we fill up the buffer. This reduces the number of syscalls.

Remove the `Cursor` variant from Stream. It was strictly less powerful than that `Test` variant, so I've replaced the handful of Cursor uses with `Test`. Because some of those cases weren't test, I removed the `#[cfg(test)]` param on the `Test` variant.

Now that all inputs to `do_from_read` are `impl BufRead`, add that as a type constraint. Change `read_next_line` to take advantage of `BufRead::read_line`, which may be somewhat faster (though I haven't benchmarked).

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -229,7 +229,7 @@ fn pool_connections_limit() {
         proxy: None,
     });
     for key in poolkeys.clone() {
-        pool.add(key, Stream::Cursor(std::io::Cursor::new(vec![])));
+        pool.add(key, Stream::from_vec(vec![]))
     }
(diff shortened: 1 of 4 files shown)


algesten (approved):
Looks great!

PR #263, opened 2020-12-02
Remove Response::status_line.

jsha:
The necessary functionality is available via `status_text`, and removing this method means we have more flexibility in our internal storage.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -408,14 +408,24 @@ impl AgentBuilder {
     ///
     /// If the redirect count hits this limit (and it's > 0), TooManyRedirects is returned.
     ///
+    /// WARNING: for 307 and 308 redirects, this value is ignored for methods that have a body.
+    /// You must handle 307 redirects yourself when sending a PUT, POST, PATCH, or DELETE request.
+    ///
     /// ```
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
     /// let result = ureq::builder()
     ///     .redirects(1)
     ///     .build()
-    ///     .get("http://httpbin.org/redirect/3")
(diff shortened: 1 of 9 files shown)


birkenfeld:
This seems a bit excessive, there is no good reason to make this information completely inaccessible?
Especially with servers that return "unusual" status codes, where the status text is the best indication of what is actually meant.

What's also missing (AFAICS) is a way to turn a status code into its equivalent standard text (500 => Internal server error), i.e. some way to create a meaningful error description.

algesten (commented):
Not sure about this.

I know the status text _should_ only double up the status code, but in practice I have encountered quite a few REST services where it conveys additional information.

Granted this was some time ago, and with HTTP/2 totally dropping status text, the problem is rather moot.

jsha:
Fair enough. I don't feel too strongly about removing `status_text`. I just happened to notice it in the spec and figured we could tidy up. For reference, I checked some other clients: Python requests doesn't seem to offer a way to get it: https://requests.readthedocs.io/en/latest/api/. Go's net/http _does_ offer a way: https://golang.org/pkg/net/http/#StatusText. Node's HTTP client _does_ offer a way: https://nodejs.org/api/http.html#http_response_statusmessage

What do you think about removing `status_line` and keeping `status_text`?

algesten:
sounds good. let's provide a way to get it if absolutely wanted.

birkenfeld:
JFTR, `requests` does provide it as `response.reason`.

jsha:
Thanks for the correction, @birkenfeld. Looks like I missed it in my quick look-through. Now I'm even more convinced that keeping status_text is right. 👍🏻

algesten (approved):

PR #266, opened 2020-12-02
Do more validation on status line.

jsha:
Status code must be exactly three digits.
HTTP version must be "HTTP/", digit, dot, digit.

https://tools.ietf.org/html/rfc7230#section-3.1.2
     status-line = HTTP-version SP status-code SP reason-phrase CRLF

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -41,7 +41,8 @@ pub const DEFAULT_CHARACTER_SET: &str = "utf-8";
 /// ```
 pub struct Response {
     url: Option<String>,
-    version: String,
+    status_line: String,
+    index: ResponseStatusIndex,
     status: u16,
     headers: Vec<Header>,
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -439,28 +439,44 @@ impl Response {
 fn parse_status_line(line: &str) -> Result<(ResponseStatusIndex, u16), Error> {
     //
 
-    let mut split = line.splitn(3, ' ');
+    let err = Err(BadStatus.new());
```
algesten, line 442: Could we maybe move this inline and give some relevant error messages? I.e.

`BadStatus.msg("Status line is not ASCII")`, `BadStatus.msg("Version does not start HTTP/")`, etc.

algesten (approved):
Totally for it. Just think we could be more specific to signal what went wrong.

PR #268, opened 2020-12-05
Add debug logs for stream pooling.

jsha:
Now each phase in a stream's life cycle should be logged at debug: creation, adding to pool, removing from pool, and dropping.

This adds one .expect() for a condition that was checked on the line above (streams in pool > max_idle). It also tweaks the debug formatter for Stream so it prints the TcpStream directly. For now, since there's not much other data in the Stream, that's a nice succinct way to get at the useful information, like the source port and fd, and the remote IP address.

```
[2020-12-05T05:28:56Z DEBUG ureq::pool] adding stream to pool: https|www.example.com|443 -> TcpStream { addr: 10.255.129.68:50004, peer: 93.184.216.34:443, fd: 4 }
```

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -7,6 +7,7 @@ use crate::stream::Stream;
 use crate::unit::Unit;
 use crate::Proxy;
 
+use log::debug;
 use url::Url;
 
 /// Holder of recycled connections.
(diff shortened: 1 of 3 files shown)


algesten (approved):
LGTM!

PR #269, opened 2020-12-05
Fixes to error::Transport

algesten:
This is a followup to #257 from @jyn514 feedback.


jyn514 (approved):
This works well :) It still has the redundant `status` code but that's not a big deal for me.

jsha (approved):

PR #272, opened 2020-12-05
Remove error_on_non_2xx.

jsha:
After the recent changes in #257, it's probably not necessary. It's now
quite easy to use a match statement to extract responses for certain
status codes, or all status codes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,12 @@ Ureq supports sending and receiving json, if you enable the "json" feature:
       .into_string()?;
 ```
 
+### Error handling
+
+ureq returns errors via `Result<T, ureq::Error>`. That includes I/O errors,
+protocol errors, and status code errors (when the server responded 4xx or
+5xx). More details on the [Error] type.
+
 ### Features
 
 To enable a minimal dependency tree, some features are off by default.
@@ -189,6 +195,7 @@ If ureq is not what you're looking for, check out these other Rust HTTP clients:
 [post()]: https://docs.rs/ureq/latest/ureq/fn.post.html
 [put()]: https://docs.rs/ureq/latest/ureq/fn.put.html
 [Request]: https://docs.rs/ureq/latest/ureq/struct.Request.html
+[Error]: https://docs.rs/ureq/latest/ureq/enum.Error.html
 [Request::call()]: https://docs.rs/ureq/latest/ureq/struct.Request.html#method.call
 [Request::send()]: https://docs.rs/ureq/latest/ureq/struct.Request.html#method.send
 [Request::send_bytes()]: https://docs.rs/ureq/latest/ureq/struct.Request.html#method.send_bytes
--- a/README.tpl
+++ b/README.tpl
@@ -12,6 +12,7 @@
 [post()]: https://docs.rs/ureq/latest/ureq/fn.post.html
 [put()]: https://docs.rs/ureq/latest/ureq/fn.put.html
 [Request]: https://docs.rs/ureq/latest/ureq/struct.Request.html
+[Error]: https://docs.rs/ureq/latest/ureq/enum.Error.html
 [Request::call()]: https://docs.rs/ureq/latest/ureq/struct.Request.html#method.call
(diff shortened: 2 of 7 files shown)


algesten (commented):
I'm not sure about removing this. I refer to this discussion https://github.com/algesten/ureq/issues/128#issuecomment-739497894

Whilst I personally prefer non-200 to take an error path, this might be surprising to other users. `error_on_non_2xx(false)` is a path to make ureq behave the way those users expect.

If we are removing it, is there maybe some other mechanic we could put in place for users who want that behavior? Maybe move it to agent?

Maybe we can make a more obvious name and turn around the true/false `handle_http_errors_as_success`.

jsha:
I'm thinking the other mechanic should be:

```rust
let resp = ureq::get("http://example.com/")
  .call()
  .or_else(|e| match e { 
    Status(_, r) => Ok(r), // turn status errors into Ok Responses.
    e => e,
  })?
```

Of course, we should document that clearly so people know how to do it. If enough people want this behavior and find the `or_else` method too onerous, we can later add back in `error_on_non_2xx` or an equivalent method. But if we launch 2.0.0 with the method included and find it's underused, we can't later delete it without breaking compatibility. It's easier to add than to remove. :-)

We did get one piece of feedback that matching on status makes error_on_non_2xx obsolete: https://github.com/algesten/ureq/pull/257#issuecomment-739329610

algesten:
Alright. If we document your above solution.

algesten (approved):

eeeebbbbrrrr:
This change caught me by surprise.  Not complaining per se, however...

Am I just blind or is the failure reason (HTTP status code, etc) not actually available on the `Error`?

It seems to me that the pattern to handle errors must be something like this, but I'm lost:

```rust
let r:Result<ureq::Response, ureq::Error> = ...;

match r.as_ref() {
   Ok(response) => ...,
   Err(e) => {
      match e.kind() {
          // I'm lost now
      }
   }
}
```

`<long pause>`

> quite easy to use a match statement to extract responses for certain status codes

Oh wait, is the pattern supposed be like?:

```rust
let r:Result<ureq::Response, ureq::Error> = ...;

match r {
   Ok(response) if response.status() == 200 => { /* it worked */ },
   Ok(response) => { /* the server returned an unexpected status */ },
   Err(e) => { /* some kind of io/transport error */ }
}
```

If the latter, then, yeah, okay, that makes sense.  

I guess maybe where I'm confused is that internally in `ureq` the property for this is `error_on_non_2xx`, but that doesn't generate a `Result:Err()`?  I am a little confused still...

jsha:
As of today's release (2.0.0-rc4), the pattern is supposed to be like this:

```rust
match r {
  Ok(response) => { /* it worked */},
  Err(Status(code, response)) => { /* the server returned an unexpected status */ }
  Err(_) => { /* some kind of io/transport error */ }
}
```

I've tried to document that at https://docs.rs/ureq/2.0.0-rc4/ureq/enum.Error.html and https://docs.rs/ureq/2.0.0-rc4/ureq/index.html#error-handling, but please let me know if the docs are confusing or the approach to error handling is un-ergonomic for you. We're trying to make this as usable as possible, so feedback is very valuable!

algesten:
@jsha @eeeebbbbrrrr I tried to document the simplest case a bit better here: https://github.com/algesten/ureq/pull/280

eeeebbbbrrrr:
Thanks!  I missed `ureq::Error` being an enum?  Was it always?

I have to admit tho, I don't know that I don't really like successful (but non-200) HTTP responses being considered a `Result::Err`.  I swapped out `reqwest` for `ureq` b/c `ureq` is easier to use and faster, but `reqwest` does consider non-200 HTTP responses a `Result::Ok()`.

In a way it feels like `ureq` is making a decision for me.  It's like, I don't consider a non-200 response an error with `ureq`, but I might want to consider it an error in my application (or not, depending on the code).

Anyways, thanks for a fantastic project.

jsha:
> I missed ureq::Error being an enum? Was it always?

It was not. We changed this in a recent release candidate after some discussions of how to make it easy to handle non-200 responses.

PR #253, opened 2020-11-23
Follow 307/308 redirects (1.x backport)

jyn514:
This backports https://github.com/algesten/ureq/pull/252 to the 1.x branch. I confirmed locally that this fixes the deadlinks issue (https://github.com/deadlinks/cargo-deadlinks/issues/110).

PR #274, opened 2020-12-06
[1.x] Fix broken behavior for 308 redirects

jyn514:
Previously, this would unconditionally panic:

```
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/joshua/src/rust/ureq/src/unit.rs:244:59
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Follow-up to #253. Sorry for not testing earlier!

### DIFF
--- a/src/test/redirect.rs
+++ b/src/test/redirect.rs
@@ -116,3 +116,17 @@ fn redirect_post() {
     assert!(resp.has("x-foo"));
     assert_eq!(resp.header("x-foo").unwrap(), "bar");
 }
+
+#[test]
+fn redirect_308() {
+    test::set_handler("/redirect_get3", |_| {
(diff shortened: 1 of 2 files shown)


jsha:
Thanks for the fix! Can you add a unittest that would detect the bug?

jsha:
By the way, for the `main` branch, #275 refactors the redirect-counting code and happens to fix this.

jyn514:
Done, I made sure it failed on master.

jsha (approved):

jsha:
Great, I'll cut a release with the fix.

jsha:
Published: https://crates.io/crates/ureq/1.5.4

PR #277, opened 2020-12-12
Clarify example from README.

jsha:
We should include the `fn main() -> Result` part, because people may not
know they need the `-> Result` in order for `?` to work.

Also, I replaced `Accept: text/html` with `Example-Header: header
value`. My thinking is that if someone doesn't know they need to remove
that Accept for non-HTML URLs, they might get unexpected results.


algesten (approved):
Yeah agree, let's show the working example.

PR #280, opened 2020-12-20
Document simplest error example

algesten:
Document the simplest possible error non-2xx error handling after some feedback in https://github.com/algesten/ureq/pull/272#issuecomment-748537427


jsha (approved):
Thanks!

PR #281, opened 2020-12-20
OrAnyStatus::or_any_status ergonomic helper

algesten:
Some users might prefer to handle all HTTP responses as Response regardless of status code.

This helper might be controversial since it means we become a bit locked down to having two variants of the `Error` enum.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -63,18 +63,17 @@ use crate::Response;
 /// ```
 ///
 /// If you'd like to treat all status code errors as normal, successful responses,
-/// you can use [Result::or_else](std::result::Result::or_else) like this:
+/// you can use [OrAnyStatus::or_any_status] like this:
 ///
 /// ```
 /// use ureq::Error::Status;
-/// # fn main() -> std::result::Result<(), ureq::Error> {
+/// # fn main() -> std::result::Result<(), ureq::Transport> {
 /// # ureq::is_test(true);
+/// use ureq::OrAnyStatus;
+///
 /// let resp = ureq::get("http://example.com/")
 ///   .call()
-///   .or_else(|e| match e {
-///     Status(_, r) => Ok(r), // turn status errors into Ok Responses.
-///     _ => Err(e),
-///   })?;
+///   .or_any_status()?;
 /// # Ok(())
 /// # }
 /// ```
@@ -96,6 +95,43 @@ pub struct Transport {
     url: Option<Url>,
     source: Option<Box<dyn error::Error + Send + Sync + 'static>>,
     response: Option<Response>,
+}
+
+/// Extension to [`Result<Response, Error>`] for handling all status codes as [`Response`].
+pub trait OrAnyStatus {
+    /// Ergonomic helper for handling all status codes as [`Response`].
+    ///
+    /// By default, ureq returns non-2xx responses as [`Error::Status`]. This
+    /// helper is for handling all responses as [`Response`], regardless
+    /// of status code.
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// # fn main() -> Result<(), ureq::Transport> {
+    /// // Bring trait into context.
+    /// use ureq::OrAnyStatus;
+    ///
+    /// let response = ureq::get("http://httpbin.org/status/500")
+    ///     .call()
+    ///     // Transport errors, such as DNS or connectivity problems
+    ///     // must still be dealt with as `Err`.
+    ///     .or_any_status()?;
+    ///
+    /// assert_eq!(response.status(), 500);
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn or_any_status(self) -> Result<Response, Transport>;
+}
+
+impl OrAnyStatus for Result<Response, Error> {
+    fn or_any_status(self) -> Result<Response, Transport> {
+        match self {
+            Ok(response) => Ok(response),
+            Err(Error::Status(_, response)) => Ok(response),
+            Err(Error::Transport(transport)) => Err(transport),
+        }
+    }
 }
 
 impl Display for Error {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -293,7 +293,7 @@ mod testserver;
 
 pub use crate::agent::Agent;
 pub use crate::agent::AgentBuilder;
-pub use crate::error::{Error, ErrorKind, Transport};
+pub use crate::error::{Error, ErrorKind, OrAnyStatus, Transport};
 pub use crate::header::Header;
 pub use crate::proxy::Proxy;
 pub use crate::request::Request;


algesten:
Hey @jsha did you see this?

src/error.rs
```diff
@@ -167,6 +167,22 @@ impl Error {
         }
     }
 
+    /// Ergonomic helper for handling all status codes as [Response].
+    ///
+    /// ```
+    /// # ureq::is_test(true);
+    /// let response = ureq::get("http://httpbin.org/status/500").call()
+    ///     .or_else(|e| e.into_response()).unwrap();
```
jsha, line 175: Could you use a `?` instead of `unwrap`? That matches our other doctests better. It requires putting `# fn main() -> Result<(), ...>` at the top, and `# Ok(()` and `# }` at the bottom.

jsha (commented):
Thanks for the reminder! This looks ok to me.

> This helper might be controversial since it means we become a bit locked down to having two variants of the Error enum.

I think we're locked to having only two variants of `Error` anyhow, since we choose not to put `#[non_exhaustive]` on the enum. I think that's right and good. There are fundamentally two types of error that act quite different, and we reflect that with the two top-level enums. For more detailed varieties of protocol-level enums, we have `ErrorKind` inside the `Transport` variant.

I'm slightly reluctant to add this on the grounds that it adds multiple ways to do the same thing - you could do the enum matching style, or you could do into_response. That's not the worst thing, though, and it seems like some folks may want this style. I guess it's a way of hedging our bets on the unusual API choice of making non-2xx errors by default.

Could you update the doc comment at the top of the file that uses `or_else` so it calls this function?

A couple of minor things:

If a user has a function that returns `Result<Response, Error>`, they won't be able to do `.into_response()?`. I think that's fine and good. People who are opting into this behavior are essentially saying they only care about `error::Transport`, so they should write their functions to return `Result<Response, Transport>`.

Right now `Transport` doesn't `impl error::Error`. We should add that.

algesten:
> I'm slightly reluctant to add this on the grounds that it adds multiple ways to do the same thing - you could do the enum matching style, or you could do into_response. That's not the worst thing, though, and it seems like some folks may want this style. I guess it's a way of hedging our bets on the unusual API choice of making non-2xx errors by default.

Maybe we could do one more double take on that decision before finalizing 2.0? If it genuinely is so surprising to users, maybe it's a bad choice? I dunno. I want to make the most user friendly, least surprising lib… 

> Could you update the doc comment at the top of the file that uses or_else so it calls this function?

Will do.

> Right now Transport doesn't impl error::Error. We should add that.

Good point! I will fix.

> Could you use a ? instead of unwrap?

Sure!

jsha:
Another variant that occurred to me: We could define a trait that we attach to `Result<Response, Error>` that accepts all Status errors and hoists them to the `Response`. This would make the "accept all status errors" case even shorter to express. Slight downside: The user would have to `use ureq::AcceptStatus`.

```patch
diff --git a/src/error.rs b/src/error.rs
index e912d8a..832432e 100644
--- a/src/error.rs
+++ b/src/error.rs
@@ -47,15 +47,12 @@ use crate::Response;
 /// you can use [Result::or_else](std::result::Result::or_else) like this:
 ///
 /// ```
-/// use ureq::Error::Status;
+/// use ureq::AcceptStatus;
 /// # fn main() -> std::result::Result<(), ureq::Error> {
 /// # ureq::is_test(true);
 /// let resp = ureq::get("http://example.com/")
 ///   .call()
-///   .or_else(|e| match e { 
-///     Status(_, r) => Ok(r), // turn status errors into Ok Responses.
-///     _ => Err(e),
-///   })?;
+///   .accept_status()?;
 /// # Ok(())
 /// # }
 /// ```
@@ -68,6 +65,20 @@ pub enum Error {
     Transport(Transport),
 }
 
+pub trait AcceptStatus {
+    fn accept_status(self) -> Result<Response, Transport>;
+}
+
+impl AcceptStatus for Result<Response, Error> {
+    fn accept_status(self) -> Result<Response, Transport> {
+        match self {
+            Ok(r) => Ok(r),
+            Err(Error::Status(_, r)) => Ok(r),
+            Err(Error::Transport(t)) => Err(t),
+        }
+    }
+}
+
 // Any error that is not a status code error. For instance, DNS name not found,
 // connection refused, or malformed response.
 #[derive(Debug)]
diff --git a/src/lib.rs b/src/lib.rs
index 4ba9b6a..efd8bba 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -226,7 +226,7 @@ mod testserver;
 
 pub use crate::agent::Agent;
 pub use crate::agent::AgentBuilder;
-pub use crate::error::{Error, ErrorKind, Transport};
+pub use crate::error::{Error, ErrorKind, Transport, AcceptStatus};
 pub use crate::header::Header;
 pub use crate::proxy::Proxy;
 pub use crate::request::Request;
```

tl-alex-butler:
I found myself wanting this too, I also reached for the trait solution though it'll be a minor annoyance having to import it.

I popped this into a project:
```rust
trait UreqResult {
    fn or_any_response(self) -> Result<ureq::Response, ureq::Transport>;
}
impl UreqResult for Result<ureq::Response, ureq::Error> {
    fn or_any_response(self) -> Result<ureq::Response, ureq::Transport> {
        match self {
            Ok(r) | Err(ureq::Error::Status(_, r)) => Ok(r),
            Err(ureq::Error::Transport(t)) => Err(t),
        }
    }
}
```

Maybe a `ureq::ResultExt` would work. More convenience methods could be added later too perhaps. E.g. `or_synthetic_response` if that v1 functionality is desired with v2.

algesten:
@jsha @tl-alex-butler After your feedback I've reworked this PR as an extension trait. Naming is always the hardest, I stuck with `into_response`, but am not dead set on it.

src/error.rs
```diff
@@ -98,6 +97,43 @@ pub struct Transport {
     response: Option<Response>,
 }
 
+/// Extensions to [`Error`].
+pub trait ErrorExt {
```
tl-alex-butler, line 101: Maybe `ResultExt` is more accurate

```suggestion
/// Extensions to [`Result<Response, Error>`].
pub trait ResultExt {
```

tl-alex-butler:
Yep naming is hard, good luck!

Just to explain my suggestions a little further: I was thinking in terms of a `Result` which has `or` methods that can return other results. So `or_any_response()` / `or_any_status()`.
```rust
// A valid response with 2xx status code
let response = ureq::get("...").call()?;
// A valid response with any status code
let response = ureq::get("...").call().or_any_status()?;
```

`Result` also has `unwrap_or` methods for returning the lhs type. I quite like the idea of having v1-style synthetic responses too, it makes a nice upgrade story for v1 user to v2.
```rust
// A valid response with any status code or a synthetic response generated from the transport error
let response = ureq::get("...").call().unwrap_or_synthetic();
```

I think `into_response()` & `into_synthetic_response()` would work fine too. Though maybe `try_into_response()` is more idiomatic as it returns a `Result`. For me these seem a _little_ harder to read though, a reader may need to look at the docs to see this wasn't just some boiler plate conversion code etc. This probably isn't a huge deal though.

src/error.rs
algesten: true true

algesten:
Thanks. Very good thoughts. I lean towards `or_any_status` because it feels most user friendly. 

Ideally for 2.x i want to leave all "synthetic errors".  

`try_into_response` i agree feels harder to read. It also feels a bit off since the happy code path doesn't use it.

jsha:
I do like `or_any_status`. I think it makes sense to call the extension trait `OrAnyStatus`, so it's clear at the import point what it's used for. Otherwise, I look at `use ureq::ErrorExt` and think "huh, what extensions need to be specially imported?" Also, `ErrorExt` suggests it's an extension trait for Error, but really it's an extension trait for Result.

algesten:
@jsha if you're happy with this PR, I'd like to rebase/reshape the commits and then merge.

src/error.rs
```diff
@@ -98,6 +97,43 @@ pub struct Transport {
     response: Option<Response>,
 }
 
+/// Extension to [`Result<Response, Error>`] for handling all status codes as [`Response`].
+pub trait OrAnyStatus {
+    /// Ergonomic helper for handling all status codes as [`Response`].
+    ///
+    /// By default, ureq handles non-2xx responses as [`Error::Status`]. This
```
jsha, line 104:
```suggestion
    /// By default, ureq returns non-2xx responses as [`Error::Status`]. This
```

jsha (approved):
Looks good! One small wording suggestion, which you can address as part of your rebasing if you agree.

PR #287, opened 2021-01-03
add documentation link to Cargo.toml

ckampfe:
Hey, thank you for the excellent library, it's a pleasure to use.
This change allows the documentation link to show up on crates.io.
Thanks!


algesten:
Thanks for the kind words!

You might have caught crates.io just when I published the new version. I think it puts the documentation link there by itself, it's just that docs.rs is a bit behind when publishing.

ckampfe:
Ah! That's great. I think my page was somehow cached or something. 😄  Regardless, thanks again for the excellent library.

PR #288, opened 2021-01-04
Fix 307/308 redirects (again, for 2.0)

jyn514:
I guess the unit test got backported but never added to `main`.

Fixes the issue in https://github.com/deadlinks/cargo-deadlinks/pull/134.

Also add some improvements to registration of test handlers, and add env_logger support to tests.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -271,7 +271,9 @@ pub(crate) fn connect(
                 307 | 308 if ["GET", "HEAD", "OPTIONS", "TRACE"].contains(&method.as_str()) => {
                     let empty = Payload::Empty.into_read();
                     debug!("redirect {} {} -> {}", resp.status(), url, new_url);
-                    return connect(unit, use_pooled, empty, Some(Arc::new(resp)));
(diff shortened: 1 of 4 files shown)


src/unit.rs
```diff
@@ -273,7 +271,9 @@ pub(crate) fn connect(
                 307 | 308 if ["GET", "HEAD", "OPTIONS", "TRACE"].contains(&method.as_str()) => {
                     let empty = Payload::Empty.into_read();
                     debug!("redirect {} {} -> {}", resp.status(), url, new_url);
-                    return connect(unit, use_pooled, empty, Some(Arc::new(resp)));
+                    // recreate the unit to get a new hostname and cookies for the new host.
+                    let new_unit = Unit::new(&unit.agent, "GET", &new_url, &unit.headers, &empty);
```
jyn514, line 275: Oops, this explicitly defeats the purpose of 308, sorry, it should use `unit.method` instead.

jsha (approved):

jsha:
Thanks for the fix, and for the test improvements! 🔧

Issue #290, opened 2021-01-04
`connect` stack overflows on thousands of redirects


jyn514:
I'm aware this is a very silly issue and if the response is "that isn't supported" I'm perfectly ok with it.

The following test case causes a stack overflow:

```rust
#[test]
fn too_many_redirects() {
    for i in 0..10_000 {
        test::set_handler(&format!("/malicious_redirect_{}", i), move |_| {
            let location = format!("Location: /malicious_redirect_{}", i+1);
            test::make_response(302, "Go here", vec![&location], vec![])
        });
    }
    let req = crate::builder().redirects(10001).build();
    let resp = req.get("test://host/malicious_redirect_0").call().unwrap();
    assert_eq!(resp.get_url(), "test://host/malicious_redirect_10000");
}
```
```
thread 'test::redirect::too_many_redirects' has overflowed its stack
fatal runtime error: stack overflow
```
I had an idea how to fix it (which is why I originally thought of this; in practice I doubt it will come up):
<details>

```diff
diff --git a/src/unit.rs b/src/unit.rs
index f0d75a0..a1dadcb 100644
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -163,15 +163,30 @@ impl Unit {
     }
 }
 
-/// Perform a connection. Used recursively for redirects.
+/// Used for retrying a failed connection, or for following a redirect.
+struct Retry<'a>(Unit, bool, SizedReader<'a>, Option<Arc<Response>>);
+
+/// Perform a connection. Follows redirects.
 pub(crate) fn connect(
     unit: Unit,
     use_pooled: bool,
     body: SizedReader,
     previous: Option<Arc<Response>>,
 ) -> Result<Response, Error> {
-    //
+    let mut req = Retry(unit, use_pooled, body, previous);
+    loop {
+        match connect_inner(req)? {
+            Ok(resp) => return Ok(resp),
+            Err(retry) => req = retry,
+        }
+    }
+}
 
+/// Perform a connection. Does not follow redirects.
+///
+/// This return type is a misuse of `Result`; really it should be `Either`.
+fn connect_inner(Retry(unit, use_pooled, body, previous): Retry) -> Result<Result<Response, Retry>, Error> {
+    let retry = |u, p, b, prev| Ok(Err(Retry(u, p, b, prev)));
     let host = unit
         .url
         .host_str()
@@ -194,7 +209,7 @@ pub(crate) fn connect(
             debug!("retrying request early {} {}: {}", method, url, err);
[...]

PR #291, opened 2021-01-04
Use iteration instead of recursion for `connect`

jyn514:
Look at this beautiful waste of time!

Closes https://github.com/algesten/ureq/issues/290.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -103,7 +103,7 @@ impl Display for Error {
         match self {
             Error::Status(status, response) => {
                 write!(f, "{}: status code {}", response.get_url(), status)?;
-                if let Some(original) = response.history.iter().next() {
+                if let Some(original) = response.history.get(0) {
                     write!(f, " (redirected from {})", original)?;
                 }
             }
--- a/src/error.rs
+++ b/src/error.rs
@@ -103,7 +103,7 @@ impl Display for Error {
         match self {
             Error::Status(status, response) => {
                 write!(f, "{}: status code {}", response.get_url(), status)?;
-                if let Some(original) = response.history().iter().last() {
+                if let Some(original) = response.history.iter().next() {
                     write!(f, " (redirected from {})", original)?;
                 }
             }
--- a/src/error.rs
+++ b/src/error.rs
@@ -300,13 +300,12 @@ fn status_code_error() {
 
 #[test]
 fn status_code_error_redirect() {
-    use std::sync::Arc;
     let mut response0 = Response::new(302, "Found", "").unwrap();
     response0.set_url("http://example.org/".parse().unwrap());
     let mut response1 = Response::new(302, "Found", "").unwrap();
-    response1.set_previous(Arc::new(response0));
+    response1.history_from_previous(response0);
     let mut response2 = Response::new(500, "Internal Server Error", "server overloaded").unwrap();
-    response2.set_previous(Arc::new(response1));
+    response2.history_from_previous(response1);
     response2.set_url("http://example.com/".parse().unwrap());
     let err = Error::Status(response2.status(), response2);
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -103,8 +103,8 @@ impl Display for Error {
         match self {
             Error::Status(status, response) => {
                 write!(f, "{}: status code {}", response.get_url(), status)?;
-                if let Some(original) = response.history().last() {
-                    write!(f, " (redirected from {})", original.get_url())?;
+                if let Some(original) = response.history().iter().last() {
+                    write!(f, " (redirected from {})", original)?;
                 }
             }
             Error::Transport(err) => {
--- a/src/error.rs
+++ b/src/error.rs
@@ -300,18 +300,22 @@ fn status_code_error() {
 
 #[test]
 fn status_code_error_redirect() {
-    let mut response0 = Response::new(302, "Found", "").unwrap();
-    response0.set_url("http://example.org/".parse().unwrap());
-    let mut response1 = Response::new(302, "Found", "").unwrap();
-    response1.history_from_previous(response0);
-    let mut response2 = Response::new(500, "Internal Server Error", "server overloaded").unwrap();
-    response2.history_from_previous(response1);
-    response2.set_url("http://example.com/".parse().unwrap());
-    let err = Error::Status(response2.status(), response2);
+    use crate::{get, test};
 
+    test::set_handler("/redirect_a", |unit| {
+        assert_eq!(unit.method, "GET");
+        test::make_response(302, "Go here", vec!["Location: test://example.edu/redirect_b"], vec![])
+    });
+    test::set_handler("/redirect_b", |unit| {
+        assert_eq!(unit.method, "GET");
+        test::make_response(302, "Go here", vec!["Location: http://example.com/status/500"], vec![])
+    });
+
+    let err = get("test://example.org/redirect_a").call().unwrap_err();
+    assert_eq!(err.kind(), ErrorKind::HTTP, "{:?}", err);
     assert_eq!(
         err.to_string(),
-        "http://example.com/: status code 500 (redirected from http://example.org/)"
+        "http://example.com/status/500: status code 500 (redirected from test://example.org/redirect_a)"
     );
 }
 
--- a/src/response.rs
+++ b/src/response.rs
@@ -460,7 +460,9 @@ impl Response {
 
     #[cfg(test)]
     pub fn set_previous(&mut self, previous: Arc<Response>) {
-        self.previous = vec![previous.get_url().to_owned()];
+        let mut history = previous.previous.clone();
+        history.push(previous.get_url().to_string());
+        self.previous = history;
     }
 }
 
--- a/src/response.rs
+++ b/src/response.rs
@@ -51,9 +51,12 @@ pub struct Response {
     headers: Vec<Header>,
     unit: Option<Unit>,
     stream: Stream,
-    // If this Response resulted from a redirect, the Response containing
-    // that redirect.
-    pub(crate) previous: Vec<String>,
+    /// The redirect history of this response, if any. The history starts with
+    /// the first response received and ends with the response immediately
+    /// previous to this one.
+    ///
+    /// If this response was not redirected, the history is empty.
+    pub(crate) history: Vec<String>,
 }
 
 /// index into status_line where we split: HTTP/1.1 200 OK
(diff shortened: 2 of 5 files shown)


src/response.rs
```diff
@@ -54,7 +53,7 @@ pub struct Response {
     stream: Stream,
     // If this Response resulted from a redirect, the Response containing
     // that redirect.
-    previous: Option<Arc<Response>>,
+    pub(crate) previous: Vec<String>,
```
jsha, line 56: This comment needs updating. In particular it should document which way the vector goes (first request to last request, or vice versa).

src/error.rs
```diff
@@ -103,8 +103,8 @@ impl Display for Error {
         match self {
             Error::Status(status, response) => {
                 write!(f, "{}: status code {}", response.get_url(), status)?;
-                if let Some(original) = response.history().last() {
-                    write!(f, " (redirected from {})", original.get_url())?;
+                if let Some(original) = response.history().iter().last() {
```
jsha, line 106: If you remove the `Hist` iterator, you'll want to change this `.last()` to `.first()`.

jsha (changes requested):
Thanks for this! I like the general concept. A couple of quick thoughts:

The main thing we want is for the latest Response in a chain at any given point to have a history of the redirects that led to it. Right now, each response gets a cloned Vec. It might be sufficient for each Response to take the history Vec from the previous response and add one to it. That said, in the process of typing that out I realized it would make the relationship between one Request and the next a little more complicated, since we'd have to mutate the previous Request. So let's not do that.

We still have the `Hist` iterator type, but I don't think we need that anymore. That was just a hack to do iteration over a chain of Arcs. Now that we have a Vec, we can use Vec's iterator. We don't even strictly speaking need an iterator - but since Vec::iter() returns an ExactSizeIterator, using `.last()` is a perfectly efficient way to get the last item.

I'm not totally sold on the Retry concept, because it combines two somewhat unrelated purposes - retrying due to broken connection, and following a redirect. What do you think of leaving "retry due to broken connection" as a recursive operation (since it should never go more than one deep for any given Unit), and treating redirects as a more traditional loop?

For instance, we could have a `connect_inner` that returns `Result<Response, Transport>`, and `connect` would have a loop that calls `connect_inner`, and if the `Response` is a 3xx, it performs the rI don't totally like edirect logic and repeats the loop.

src/response.rs
jyn514: `hist.iter()` returns a reverse iterator. I agree this is confusing, I'll change it to return a normal iterator and use `first()` here instead.

jyn514 (commented):
> Right now, each response gets a cloned Vec.

The vec is not cloned. This is the reason I moved `previous` out of do_from_request, so I would be able to access it directly even in case of an IO error.

> What do you think of leaving "retry due to broken connection" as a recursive operation (since it should never go more than one deep for any given Unit)

Why will it never go more than one deep? I don't see any state tracking that.

Assuming it never goes more than one deep, I agree separating the two makes sense.

> For instance, we could have a connect_inner that returns Result<Response, Transport>, and connect would have a loop that calls connect_inner, and if the Response is a 3xx, it performs the redirect logic and repeats the loop.

As in, move the 300 handling out of `connect_inner`? Yeah, that sounds a lot simpler.

src/response.rs
```diff
@@ -466,7 +460,9 @@ impl Response {
 
     #[cfg(test)]
     pub fn set_previous(&mut self, previous: Arc<Response>) {
-        self.previous = Some(previous);
+        let mut history = previous.previous.clone();
```
jsha: > > Right now, each response gets a cloned Vec.

> The vec is not cloned. This is the reason I moved previous out of do_from_request, so I would be able to access it directly even in case of an IO error.

Isn't this (line 463) cloning the vec?

Also, I noticed this is still taking an `Arc<Response>`, but I think it can just be a `&Response` now, right?

jsha (commented):
> Why will it never go more than one deep? I don't see any state tracking that.

[RFC 7230 says this](https://tools.ietf.org/html/rfc7230#section-6.3.1):

> A client SHOULD NOT automatically retry a failed automatic retry.

This is implemented by the `use_pooled` parameter to `connect`. When we first arrive, it's true. If we do an automatic retry, it's false on the second call. We also gate both retry paths on `is_recycled`, which can only be true if we asked for a pooled connection and got one. This is definitely a bit convoluted and spread over more locations than it should be. We should clarify the code.

src/response.rs
jyn514: This is cloning the vec, but it's for tests only, because it takes a `Vec`. The main logic in `connect_inner` doesn't do any clones.

> Also, I noticed this is still taking an Arc<Response>, but I think it can just be a &Response now, right?

I think so, yes. I'll try changing that.

src/unit.rs
```diff
@@ -163,12 +160,69 @@ impl Unit {
...
 }
 
-/// Perform a connection. Used recursively for redirects.
+/// Perform a connection. Follows redirects.
 pub(crate) fn connect(
-    unit: Unit,
+    mut unit: Unit,
+    use_pooled: bool,
+    mut body: SizedReader,
+    mut previous: Vec<String>,
```
jsha, line 168: This is now state entirely internal to the `connect` fn, so it can be moved to a local variable.

jsha (changes requested):
I'm finding the ownership flow of `previous` back and forth between the two functions to be fairly confusing. Is it possible for ownership of `previous` to remain entirely within the outer `connect` fn?

jyn514:
> Is it possible for ownership of previous to remain entirely within the outer connect fn?

Yes, I think so - it just needs to be set before returning from `connect()`, it doesn't matter very much what `resp.history` is before then.

src/unit.rs
jyn514: It's not entirely internal state - `send_prelude` needs to know if this is a redirect or not. But it can do that just as easily with a slice instead.

jyn514: Oh wait, I just realized you meant this should only ever be called with `vec![]`. Yes, it can be removed from `connect`, just not `connect_inner`.

jyn514:
I'm pretty happy with this :) let me know when/if I should squash the commits.

jsha (approved):
Yeah, this is looking really nice! I think this is ready to go. No need to squash the commits. When I go to merge, I'll use GitHub's "squash and merge" function.

I'll come back and give it a second look with fresh eyes in the morning before hitting the button.

algesten:
Wow. This is quite a PR. Thanks! I'll go through it later today too.

src/unit.rs
```diff
@@ -163,12 +160,69 @@ impl Unit {
...
 
-/// Perform a connection. Used recursively for redirects.
+/// Perform a connection. Follows redirects.
 pub(crate) fn connect(
-    unit: Unit,
+    mut unit: Unit,
+    use_pooled: bool,
+    mut body: SizedReader,
+) -> Result<Response, Error> {
+    let mut previous = vec![];
```
jsha, line 169: Nit: I'd like a type annotation on this. Otherwise you have to read pretty far to figure out what's stored in this vec.

src/unit.rs
```diff
@@ -163,12 +160,69 @@ impl Unit {
...
+    loop {
+        let mut resp = connect_inner(&unit, use_pooled, body, &previous)?;
+        previous.push(unit.url.to_string());
+        resp.history = previous;
+        // handle redirects
+        if (300..399).contains(&resp.status()) && unit.agent.config.redirects > 0 {
```
jsha, line 175: Nit: I'd like to reorder the cases in this loop to favor early returns and reduce indentation. Specifically, something like:

```rust
if !(300..399).contains(&resp.status()) || unit.agent.config.redirects == 0 {
  return Ok(resp);
}
... otherwise, handle a redirect ...
```

This also allows getting rid of the `continue`, which IMO makes this feel more like a natural loop, where falling through to the bottom means the loop continues by default.

src/unit.rs
```diff
@@ -163,12 +160,69 @@ impl Unit {
...
-    unit: Unit,
+    mut unit: Unit,
+    use_pooled: bool,
+    mut body: SizedReader,
+) -> Result<Response, Error> {
+    let mut previous = vec![];
+    loop {
+        let mut resp = connect_inner(&unit, use_pooled, body, &previous)?;
+        previous.push(unit.url.to_string());
+        resp.history = previous;
```
jsha, line 173: I think this is slightly early to push the unit's URL onto `previous`. Doesn't this mean a resp with no redirects contains itself in its history? We should make sure we have tests that catch this, if it's the case (I might also be misunderstanding here).

jsha (changes requested):

src/unit.rs
jyn514: I'm confused why you say that. `history.push(unit.url.to_string())` is only a couple lines below: https://github.com/algesten/ureq/blob/de669f65eca0d3d5a19b6b750449760e8a50f5ae/src/unit.rs#L172

src/unit.rs
jyn514: Done, although I used `break` instead of return to avoid repeating the `Ok` a lot.

I also noticed that this returns `Ok(resp)` when `redirects == 0`, but `Err(TooManyRedirects)` for any other value of redirects. Is that intentional?

src/unit.rs
jyn514: Ooh, excellent catch. Yes, this should only push the history if we're sure we're going to redirect (near the bottom, after calculating `new_method`.

That brings up another issue: I don't think the history tests actually test anything. Because they add items to the history manually, it doesn't matter what `connect` does, they'll pass no matter what.

jsha (approved):
Looks good to me! @algesten, since you expressed interest in reviewing, I'll wait on your approval.

src/unit.rs
jsha: Yep, that's intentional. `0` is a special value that means "turn off redirect handling."

src/unit.rs
jsha: Heh, true. I got hung up on the line before that:

https://github.com/algesten/ureq/blob/de669f65eca0d3d5a19b6b750449760e8a50f5ae/src/unit.rs#L171

And thought "I wonder what is being passed to connect_inner? I guess I'd have to go look at the definition for that function.

In general I've been developing a preference for explicit type annotations in a lot of cases. But I won't be picky about it in this specific spot.

algesten (approved):
Yep. This all looks good to me. Merge away!

PR #292, opened 2021-01-04
Fix some clippy lints

jyn514:
This commit can be replicated with `cargo +nightly clippy --fix -Z unstable-options`.

I am 100% ambivalent about whether this lands, feel free to close.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -279,16 +279,16 @@ pub fn is_test(is: bool) -> bool {
         IS_TEST.store(true, Ordering::SeqCst);
     }
     let x = IS_TEST.load(Ordering::SeqCst);
-    return x;
+    x
 }
 
 /// Agents are used to hold configuration and keep state between requests.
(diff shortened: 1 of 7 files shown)


jsha (approved):
I'm all in favor of staying clippy-clean. We should probably add a clippy phase to our CI.

src/lib.rs
```diff
@@ -279,16 +279,16 @@ pub fn is_test(is: bool) -> bool {
...
     #[cfg(not(test))]
     if is_test(false) {
-        return testserver::test_agent();
+        testserver::test_agent()
     } else {
-        return AgentBuilder::new().build();
+        AgentBuilder::new().build()
     }
     #[cfg(test)]
     return testserver::test_agent();
```
jsha, line 294: Can you remove this return as well? That would make for symmetry with the cases above. Presumably clippy didn't catch it due to the `#[cfg(test)]` guard.

PR #293, opened 2021-01-04
Response: Use ErrorKind::UnexpectedEof for "premature close"

rawler:
In a client application, we're explicitly trying to differentiate "invalid data" scenarios from "broken transfer". `ureq` calling a broken transfer "invalid data" becomes a bit confusing. We can of course solve it locally for us, but I'd at least want to bring it up for discussion first.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -618,7 +618,7 @@ impl<R: Read> Read for LimitedRead<R> {
             // received, the recipient MUST consider the message to be
             // incomplete and close the connection.
             Ok(0) => Err(io::Error::new(
-                io::ErrorKind::InvalidData,
+                io::ErrorKind::UnexpectedEof,
                 "response body closed before all bytes were read",
             )),
             Ok(amount) => {
@@ -636,7 +636,7 @@ fn short_read() {
     let mut lr = LimitedRead::new(Cursor::new(vec![b'a'; 3]), 10);
     let mut buf = vec![0; 1000];
     let result = lr.read_to_end(&mut buf);
-    assert!(result.is_err());
+    assert!(result.err().unwrap().kind() == io::ErrorKind::UnexpectedEof);
 }
 
 impl<R: Read> From<LimitedRead<R>> for Stream


rawler:
Other alternatives, is to use `ConnectionAborted` or `ConnectionReset` instead.

algesten:
@rawler I see you've made the changes against 1.x and we released 2.0.0 yesterday. Do you mean you want us to make more releases of the 1.x branch?

jsha:
FWIW, I think this is a good idea, though my preference would be to apply the change to the main branch and release it as part of 2.0.1.

Docs for io::ErrorKind::UnexpectedEof:

> UnexpectedEof
> An error returned when an operation could not be completed because an "end of file" was reached prematurely.
>
> This typically means that an operation could only succeed if it read a particular number of bytes but only a smaller number of bytes could be read.

rawler:
We're currently on 1.x. Haven't assessed upgrading, wanted to start with discussing the concept.

Depending on how you see it, it's also an API-breaking change, so not sure it should technically be applied to 1.x OR 2.x.

rawler:
(I just realized I don't see my "point of discussion rather than suggestion to merge"-comment from yesterday. Must have forgotten to submit, or something went wrong)

rawler:
(Re-based and re-targeting main)

algesten:
The concept is good. The breaking change aspect is a bit annoying given we just released 2.0.0.

@jsha any thoughts on that?

jsha:
I think this is a small enough change to include in the 2.0.0 series. The behavior still triggers an error, it's just a different error. We don't guarantee in the documentation that an early EOF returns InvalidData, or that an InvalidData error indicates early EOF. If someone took a dependency based on that behavior, it would be based on reading the source.

Also, I think given the short amount of time since 2.0.0 was released, it's unlikely that anyone has taken a dependency on this specific error type resulting from this specific error since moving to 2.0.0.

This is a good conversation to have, though, because it suggests that now that we are intentionally making an API guarantee that EOF == UnexpectedEof, we should document that.

jsha (approved):
@algesten any thoughts before I merge?

algesten:
since it's early days for 2.x, let's do it.

PR #300, opened 2021-01-11
Add request Display impl `$method $url`

### LINKED ISSUE #297: Request::get_method, get_url gone in v2
I was using these for terse logging of my test requests & responses, e.g.
```rust
eprintln!("-> {} {}", req.get_method(), req.get_url());
```
So I could build the `Request`s normally in many places and have a single place I logged the request & response info. 

I could use the `Debug` impl of `Request` but it includes, in my case, lots of auth header noise that i didn't want. So it would be nice to have a way to access the `Request` method & url again.

tl-alex-butler:
Add `impl Display for Request` so that `ureq::get("https://example.com")` -> `GET https://example.com`. Ignores all other request properties. Terse + easy to read.

Resolves #297

If this wasn't how you guys wanted to go about it, feel free to close.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+ * Add `Display` impl for `Request` of form _`"$method $url"`_.
+
 # 2.0.1
  * Fix handling of 308 redirects (port from 1.5.4 branch).
  * Return UnexpectedEof instead of InvalidData on short responses. (#293)
--- a/src/request.rs
+++ b/src/request.rs
@@ -21,6 +21,15 @@ enum Urlish {
     Str(String),
 }
 
+impl fmt::Display for Urlish {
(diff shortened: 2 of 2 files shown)


jsha (approved):
This looks good to me. For completeness' sake, another possibility is Debug's "alternate format specifier" `{#?}`; we could put the detailed stuff we have today as an alternate format debug, and have a more concise version here as `{:?}`. But even the concise version should probably have `Request { ... }`, and you're looking for something shorter, so this seems good.

https://doc.rust-lang.org/std/fmt/trait.Debug.html

algesten:
Since we haven't implemented Display consistently across the board for our public data structures, I'm more in favor of amending Debug.

algesten:
Closing due to inactivity. Still happy to accept a PR based on implementing reasonable `Debug` for all public structs.

PR #303, opened 2021-01-16
cargo fmt

algesten:


algesten:
Wonder if we can build into the ci run that `cargo fmt` doesn't make any changes to the source tree.

jsha (approved):
Yeah, we can definitely incorporate that check into CI! Would be nice alongside a `cargo clippy` check.

PR #306, opened 2021-01-18
Fix some clippy warnings.

jsha:
This doesn't fix all of them, but some of the remaining ones are more
involved to fix.


algesten (approved):
sweet

algesten:
got rustfmt lint before merge.

PR #310, opened 2021-01-26
Additional Request functions

### LINKED ISSUE #297: Request::get_method, get_url gone in v2

algesten:
Provide functions on `Request` to inspect the state. 

`.method()` returns GET, POST etc and `.url()` returns the url _as a &str;_, since this is our "native" url format.

Sadly there's an inconsistency carried over from old where `.set` sets a header, and `.query` sets a query parameter. To inspect headers, we have `.header()` and `.header_names()` (the consistent naming would have been `.header()` to set the header).

The new functions for query parameters are `.query_params()` to list all set query parameters, `.query_value()` to read the first value for a `param` and `.query_values()` to get all query values for a name.

The PR is big because there are two ways to set query parameters. These are equivalent:

1. `ureq::get("http://some/?my=param")`
2. `ureq::get("http://some/").query("my". "param")`

`.query_params()`, `.query_value()` etc should list all parameters, regardless of how they were set. We don't want to parse query multiple times, so the PR moves the parse of `url: &str` from being done last in the "build" methods `.do_call()` to be already at instantiation. An additional hurdle is that `Url` gives us the parsed parameters as `Cow<String>`, while we want the `ureq` API to operate on `&str`.

Close #297

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -4,7 +4,6 @@ use std::{fmt, time};
 use url::{form_urlencoded, Url};
 
 use crate::body::Payload;
-use crate::error::ErrorKind;
 use crate::header::{self, Header};
 use crate::unit::{self, Unit};
 use crate::Response;
@@ -15,10 +14,29 @@ use super::SerdeValue;
 
 pub type Result<T> = std::result::Result<T, Error>;
 
-#[derive(Debug, Clone)]
-enum Urlish {
-    Url(Url),
-    Str(String),
+#[derive(Clone)]
+struct ParsedUrl(std::result::Result<Url, url::ParseError>);
+
+impl fmt::Display for ParsedUrl {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Ok(url) = &self.0 {
+            write!(f, "{}", url.as_str())
+        } else {
+            write!(f, "{:?}", self.0)
+        }
+    }
+}
+
+impl From<String> for ParsedUrl {
+    fn from(s: String) -> Self {
+        ParsedUrl(s.parse())
+    }
+}
+
+impl From<Url> for ParsedUrl {
+    fn from(url: Url) -> Self {
+        ParsedUrl(Ok(url))
+    }
 }
 
 /// Request instances are builders that creates a request.
@@ -36,53 +54,38 @@ enum Urlish {
 pub struct Request {
     agent: Agent,
     method: String,
-    url: Urlish,
+    parsed_url: ParsedUrl,
     error_on_non_2xx: bool,
     headers: Vec<Header>,
-    query_params: Vec<(String, String)>,
     timeout: Option<time::Duration>,
-}
-
-impl fmt::Display for Urlish {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            Urlish::Url(u) => write!(f, "{}", u),
-            Urlish::Str(s) => write!(f, "{}", s),
-        }
-    }
 }
 
 impl fmt::Debug for Request {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "Request({} {} {:?}, {:?})",
-            self.method, self.url, self.query_params, self.headers
+            "Request({} {}, {:?})",
+            self.method, self.parsed_url, self.headers
         )
     }
 }
 
 impl Request {
     pub(crate) fn new(agent: Agent, method: String, url: String) -> Request {
-        Request {
-            agent,
-            method,
-            url: Urlish::Str(url),
-            headers: vec![],
-            error_on_non_2xx: true,
-            query_params: vec![],
-            timeout: None,
-        }
+        Self::_new(agent, method, url.into())
     }
 
     pub(crate) fn with_url(agent: Agent, method: String, url: Url) -> Request {
+        Self::_new(agent, method, url.into())
+    }
+
+    fn _new(agent: Agent, method: String, parsed_url: ParsedUrl) -> Request {
         Request {
             agent,
             method,
-            url: Urlish::Url(url),
+            parsed_url,
             headers: vec![],
             error_on_non_2xx: true,
-            query_params: vec![],
             timeout: None,
         }
     }
@@ -115,17 +118,8 @@ impl Request {
         for h in &self.headers {
             h.validate()?;
         }
-        let mut url: Url = match self.url {
-            Urlish::Url(u) => u,
-            Urlish::Str(s) => s.parse().map_err(|e| {
-                ErrorKind::InvalidUrl
-                    .msg(&format!("failed to parse URL: {:?}", e))
-                    .src(e)
-            })?,
-        };
-        for (name, value) in self.query_params {
-            url.query_pairs_mut().append_pair(&name, &value);
-        }
+        let url = self.parsed_url.0?;
+
         let deadline = match self.timeout.or(self.agent.config.timeout) {
             None => None,
             Some(timeout) => {
@@ -349,7 +343,7 @@ impl Request {
     /// ```
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
-    /// let resp = ureq::get("http://httpbin.org/response-headers")
+    /// let resp = ureq::get("http://httpbin.org/get")
     ///     .query("format", "json")
     ///     .query("dest", "/login")
     ///     .call()?;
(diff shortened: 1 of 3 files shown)


jsha:
I would prefer if `.url()` returned a `Url`. Having `&str` as our input parameter makes sense, since it's super common to want to write a constant string as input to `get()`. But for the case of "I have a Request, and want some details about what URL it's going to go to," I think `Url` is better. As a purely practical matter, this lets us skip the complexity of `.query_params()` and `.query_values()` since Url already has a good way to get those. It also lets us skip the complexity of `ParsedUrl`. When `.url()` is called, we can parse our Urlish at that time, mix in any query params, and return the resulting Url (or Error).

algesten:
> As a purely practical matter, this lets us skip the complexity of .query_params() and .query_values() since Url already has a good way to get those.

I'm not sure about making Url _the way_ to get data out of our API because I don't think Url has a particularly user friendly surface. It's written trying to do as little data copying as possible; performance above all. `Url::query_pairs` is point in case returning an _iterator_ of pairs of `Cow`, which is a pretty awful type to handle for users. Implementing "What is the value of query parameter 'foo'", is a rather tedious exercise in that API.

Also we are deliberately delaying `Result` to happen only when we convert `Request` -> `Response`, which I think is nice. For me it would be surprising that state inspection functions like `.url()` would a) give me `Result` and b) give me a copy of `Url` (instead of a ref).

I'm trying to figure out some compromise way forward, like… Url + extension trait? However I keep getting stuck at those darn `Cows` 🐮🐮🐮.

```
impl UrlExt for Url {
    fn query_params(&self) -> Vec<&str> {
        let mut v = vec![];
        for (k, _) in self.query_pairs() {
            v.push(k.as_ref());
        }
        v
    }
```

```
error[E0515]: cannot return value referencing local variable `k`
   --> src/request.rs:448:9
    |
446 |             v.push(k.as_ref());
    |                    - `k` is borrowed here
447 |         }
448 |         v
    |         ^ returns a value referencing data owned by the current function
```

Maybe there's something I'm missing here. It feels like this should be possible somehow.

jsha:
I spent some time trying to get your extension trait version working and returning `&str`. I wasn't able to do it, but this String version is relatively straightforward:

```rust
trait UrlExt {
    fn query_params(&self) -> Vec<String>;
}

impl UrlExt for Url {
    fn query_params(&self) -> Vec<String> {
        self.query_pairs().into_owned().map(|p| p.0).collect()
    }
}
```

Here's my attempt to simplify things down and explicitly annotate lifetimes to better understand what's going on:

```rust
fn qstr<'a>(url: &'a Url) -> &'a str {
    let mut qp: Parse<'a> = url.query_pairs();
    let cow: Cow<'a, str> = qp.next().unwrap().0;
    let bcow: &'a Cow<'a, str> = &cow;
    let br: &'a str = bcow.borrow();
    br
}
```

This errors on the `let bcow` line with:
```
`cow` does not live long enough
borrowed value does not live long enoughrustcE0597
error.rs(124, 1): `cow` dropped here while still borrowed
error.rs(118, 9): lifetime `'a` defined here
error.rs(121, 15): type annotation requires that `cow` is borrowed for `'a`
```

Note that `cow` is not a reference with lifetime `'a`, it's an owned value within the function, whose type is parameterized by `'a`. That means: "`cow` is either a fully owned `String` _or_ it's a reference: a `&'a str`." In the latter case, it should be possible to return the `&'a str`. It has the right type and lifetime.

However, in the case that `cow` is owned, it's not possible to return a reference to its inner String, since that String will stop existing when the function exits. From the type system's perspective, once we own that `Cow`, it would be perfectly valid for us to mutate it, which would definitely turn it into an owned `String`. For example, imagine we wanted to lowercase it:

```rust
fn qstr<'a>(url: &'a Url) -> &'a str {
    let mut qp: Parse<'a> = url.query_pairs();
    let mut cow: Cow<'a, str> = qp.next().unwrap().0;
    cow.to_mut().to_lowercase();
    let bcow: &'a Cow<&'a str> = &cow;
    let br: &'a str = bcow.borrow();
    br
}
```

This demonstrates why you can't take a reference to `&cow` and assign it to `bcow: &'a Cow<'a, str>`. The lifetime of `&cow` is not `&a`, it's a much shorter lifetime that only exists within the function. Probably `bcow` should be `&'b Cow<'a, str>`, but that wouldn't really help since `'b` is too short.

So now we have a question: Why does `Url::query_pairs()` return `Cow<'a, str>`? It's not useful for us to modify the thing we get back. Since any query pair we get back in theory exists within `Url`, or `Url` has a reference to it, `Url` _should_ be able to give us a reference that lives as long as Url. This may be an API design issue, but more likely I'm just not fully understanding. One theory I have is that `Cow` is an infectious type. If you have `Cow` anywhere in your implementation, you're going to wind up with it everywhere, because of the lifetime issues I described above. The `Cow` type is introduced by `form_urlencoded`'s Parse: https://docs.rs/form_urlencoded/1.0.0/form_urlencoded/struct.Parse.html. I'm not sure why that package uses Cow, but I'm curious now so I'll dig and find out.

So, what to do for our specific use case? Now that I've spent a while wrestling with `Cow`, I'm more convinced than ever it's not a user-friendly type, even if there's a quick way to turn `Cow<str>` into `String`. My ideal would be to change Url upstream to have a more user-friendly interface by default. But that could take a while. I think the extension trait approach is a decent stopgap.

jsha:
I found out why `form_urlencoded` uses Cow: https://github.com/servo/rust-url/commit/9f05a0108025a1c9002e8977a42dce764895d170#diff-f22a711df8ea59f8d7c4ee4242a7a2b9477943fb90526ec1f086e2a6ed70b162L65-L72

If there is something percent-encoded in the input, the parser will need to allocate a new String to store the decoded version. However, if there's nothing percent-encoded in the input, the parser can skip the allocation and return a reference to the underlying bytes. So my assumption above "Since any query pair we get back in theory exists within Url, or Url has a reference to it, Url should be able to give us a reference that lives as long as Url" was incorrect. For inputs that record decoding, there are no underlying bytes that give us exactly what we need, and a String has to be generated at parse time.

algesten:
> If there is something percent-encoded in the input, the parser will need to allocate a new String to store the decoded version. However, if there's nothing percent-encoded in the input, the parser can skip the allocation and return a reference to the underlying bytes.

@jsha feels like `Url` should be able to deal with that without `Cow`. I mean either have a pair of offsets into the original parsed bytes, or have an allocated `String`. Either way they could return a `&str`.

But as you say, getting this changed upstreams is probably a long while away. I think we should focus on _ureq_ ideal API, even if that means we have a somewhat inefficient impl until we can get some changes upstream.

algesten:
@jsha 

I'd like to make progress on this PR. Let's focus on the surface API of ureq and worry about the efficiency of the internals later.  Let's summarize what we got since I'm quite happy with these additions:

* `pub fn method(&self) -> &str`

I think `method is a `&str` since we got [`ureq::request`](https://docs.rs/ureq/2.0.2/ureq/fn.request.html) and the equivalent on agent. No `Method` enum etc.

* `pub fn url(&self) -> &str`

I argue`url` for ureq also primarily is a `&str`. The `Url` variants are convenience secondary accepted types. Specifically I don't consider `Url` API particularly user friendly, but if people want to interact with them, we can accept them as input. A further complication is that we want potential url parsing errors from `Request`, being a builder pattern, to surface at "build time" (`call()`, `send()`).

All in all, I'm liking `&str` here.

* `pub fn query_params(&self) -> Vec<&str>`
* `pub fn query_value(&self, param: &str) -> Option<&str>`
* `pub fn query_values(&self, param: &str) -> Vec<&str>`

The names are up for bike shedding, but I like that these work from both parameters passed in the url, as well as those added later via `query()`. The types being `&str` also seems natural if `url()` returns `&str`.

An alternative would be to have some wrapper type `QueryParameters` that encapsulates all these functions as well as mutations (add, remove).

jsha:
Agreed that we should make progress here with what works best for `ureq`'s API. I'll take a deeper look at this later today. I generally agree with using `&str` here. My one question is how we arrange to make that available? I think we may have to push URL parsing much earlier (and memoize a possible error for later). Note that we could also choose to return `String`, which has the advantage that we don't have to memoize anything on the object. I suspect it might work out a lot nicer to use `String`.

algesten:
If you look at my PR, it's doing exactly that. Moving URL parsing earlier, but surfacing the error later.

jsha:
> If you look at my PR, it's doing exactly that. Moving URL parsing earlier, but surfacing the error later.

Ah, right, I remember seeing that previously; sorry for forgetting!

I'm onboard with the current approach, in terms of returning query information as `&str` pointing to some internal data structure. I still feel like we're asking `Request` to do the work of two types - a request and a URL. It would be more elegant to have the accessor methods on a separate type that does just the one job of being a URL. Of course, the downside of doing that is we'd have our own `UreqUrl` or similar object, which would be confusing since there's another thing called `Url` that we use as input. So, let's go ahead.

For the specific details here, having `all_query_params`, `query_params`, `query_value`, and `query_values` seems like a lot. Can we get rid of `query_params`? It can be quickly implemented by the user as a map over `all_query_params`. Also, `query_value` should define which value it picks (i.e. the first).

> I argue url for ureq also primarily is a &str. The Url variants are convenience secondary accepted types. Specifically I don't consider Url API particularly user friendly, but if people want to interact with them, we can accept them as input.

FWIW, I consider the `Url` variants not as convenience variants but as correctness ones. For someone building a system that manipulates URLs (adding and removing query params, modifying path components, etc), passing the manipulated Url directly into ureq rather than round-tripping it through a `&str` gives better confidence that the Url you constructed is the Url being fetched.

Is the `url(&self) -> &str` accessor critical to the request-signing work you're trying to unblock? What do you think about splitting out the url accessor into a separate conversation?

algesten:
@jsha Let's try another iteration.

In this version there's only one single function for query parameters. `pub fn query_params(&mut self) -> &mut Vec<(String, String)>`. It allows inspection and manipulation without a flurry of extra functions.

I also pulled out the url inspection to make that a separate PR.

algesten:
Sorry I'm going in circles here. Now the `query_params` is:

`pub fn query_params(&self) -> Vec<(String, String)>`

It stores query manipulation back into the parsed url.

src/request.rs
```diff
@@ -111,21 +114,12 @@ impl Request {
         self.do_call(Payload::Empty)
     }
 
-    fn do_call(&self, payload: Payload) -> Result<Response> {
+    fn do_call(self, payload: Payload) -> Result<Response> {
```
jsha, line 117: This makes this a breaking change, since we're consuming the Request rather than using a reference to it. Can we make this work with `&self`?

jsha (commented):
This is looking good! I like the current revision. One comment about a breaking change. Otherwise I think we're good to go.

src/request.rs
algesten: It's actually not. This is not `pub`, but used from `call`, `send`, `send_form` etc, and all those are consuming. This change is simply avoiding a clone that was here earlier, but doesn't change the public API.

jsha (approved):

PR #311, opened 2021-01-27
Add a user_agent() method to AgentBuilder to configure the default User-Agent header

### LINKED ISSUE #308: Configure the user-agent via Agent
Looking at [`unit.rs`](https://github.com/algesten/ureq/blob/06abdff/src/unit.rs#L381) it seems the intended way to override the user-agent is by setting the header manually on each request (e.g. `.set("User-Agent", "foo-bar")`). This get's a bit tedious if you're sending multiple requests since you need to set the same user-agent string over and over.

It'd be great if `AgentBuilder` had a `user_agent()` method that set the user-agent for all requests from that agent. Would you be open to a PR adding something like this?

dimo414:
A few questions:

1. Should there be any error checking, e.g. that the string is non-empty?
2. Should there be a way to disable the User-Agent header? I opted not to add a `no_user_agent()` method since it doesn't seem necessary (and could always be added later) but wanted to confirm.
3. What test coverage would you like me to add, if any? I tried to add a test to `simple.rs` however since the `User-Agent` and `Accept` headers are written directly to the stream they can't be inspected in the same way. Removing [lines 381-390](https://github.com/algesten/ureq/blob/06abdff/src/unit.rs#L381) doesn't cause any tests to fail presently.

Fixes #308

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -36,6 +36,7 @@ pub(crate) struct AgentConfig {
     pub timeout_write: Option<Duration>,
     pub timeout: Option<Duration>,
     pub redirects: u32,
+    pub user_agent: String,
     #[cfg(feature = "tls")]
     pub tls_config: Option<TLSClientConfig>,
 }
@@ -208,6 +209,7 @@ impl AgentBuilder {
                 timeout_write: None,
                 timeout: None,
                 redirects: 5,
+                user_agent: format!("ureq/{}", env!("CARGO_PKG_VERSION")),
                 #[cfg(feature = "tls")]
                 tls_config: None,
             },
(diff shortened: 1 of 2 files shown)


src/unit.rs
```diff
@@ -342,8 +342,6 @@ fn connect_socket(unit: &Unit, hostname: &str, use_pooled: bool) -> Result<(Stre
 /// Send request line + headers (all up until the body).
 #[allow(clippy::write_with_newline)]
 fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> io::Result<()> {
-    //
-
```
algesten, line 346: unrelated change?

src/agent.rs
```diff
@@ -434,6 +436,32 @@ impl AgentBuilder {
...
+    /// let result = agent.get("http://httpbin.org/headers").call()?;
+    ///
+    /// // Overrides user-agent set on the agent
+    /// let result = agent.get("http://httpbin.org/headers")
+    ///     .set("User-Agent", "super-ferris/2.0")
+    ///     .call()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn user_agent(mut self, user_agent: String) -> Self {
```
algesten, line 460: I'd like this to be `user_agent: &str` to be analogous with the rest of the API, see `Request::set` for instance.

algesten (changes requested):
This looks good! Thank you!

> 1. Should there be any error checking, e.g. that the string is non-empty?
>
> 2. Should there be a way to disable the User-Agent header? I opted not to add a no_user_agent() method since it doesn't seem necessary (and could always be added later) but wanted to confirm.

Not having a user agent is a very rare case so I agree we shouldn't clutter the API with a `no_user_agent()`. We could potentially make it so that setting an empty string effectively unsets it. This would ideally be implemented in `send_prelude`, so that user agent can be disabled either on agent or request level.

> 3. What test coverage would you like me to add, if any? I tried to add a test to simple.rs however since the User-Agent and Accept headers are written directly to the stream they can't be inspected in the same way. Removing lines 381-390 doesn't cause any tests to fail presently.

I'm happy covering it with only doc tests. You need to add some inspection of the result in addition to what you have.

src/unit.rs
dimo414: Just cleaning up (what seemed like) noise. If this is intentional I'll restore it.

src/agent.rs
dimo414: SGTM

dimo414:
> You need to add some inspection of the result in addition to what you have.

Added some assertions using `feature=json` to parse the httpbin output. I can swap to a string contains check if you don't want this gated on the feature.

src/unit.rs
algesten: it's my quirk. no biggie ;)

algesten (approved):
LGTM! Let's ship it.

I'll try to get two more PR into a point release. #281 and #310 … just waiting to see if @jsha got a mo'.

jsha (approved):

dimo414:
@algesten just checking on the timeline of a new point release that will include this PR?

algesten:
@dimo414 sorry for the silence. On it now.

dimo414:
No worries, thanks!

Issue #312, opened 2021-02-04
Timeout not honored (program hang?) on some websites


Shnatsel:
Hello! I'm re-running the "download top million websites" test with various HTTP clients.

The only issue I've found is in 7 cases out of a million, ureq process is killed on a 60-second timeout by an external tool even though it has connect timeout of 10 seconds and a timeout for the entire download set to 30 seconds.

I'm using a modified version of the [in-tree test](https://github.com/algesten/ureq/blob/main/examples/smoke-test/main.rs), which can be found [here](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/ureq-smoke-test/src/main.rs). I'm using the [Tranco list](https://tranco-list.eu/), just like the in-tree test.

I'm attaching an archive with the output for all 7 problematic websites: [ureq_hangs.tar.gz](https://github.com/algesten/ureq/files/5927656/ureq_hangs.tar.gz)

jsha:
Thanks for the report! Are you able to reproduce reliably for those hostnames? I tried a couple with an unmodified copy of the `smoke-test` in-tree example, and was not able to reproduce.

```
$ time ./target/debug/examples/smoke-test <(echo 1,shuyang.gov.cn)
[2021-02-04T18:55:46Z ERROR smoke_test] ⚠️ https://www.shuyang.gov.cn/ https://www.shuyang.gov.cn/: Network Error: timed out reading response
[2021-02-04T18:55:46Z ERROR smoke_test] ⚠️ https://shuyang.gov.cn/ https://shuyang.gov.cn/: Connection Failed: Connect error: connection timed out

[2021-02-04T18:56:04Z ERROR smoke_test] ⚠️ http://shuyang.gov.cn/ http://shuyang.gov.cn/: Network Error: timed out reading response

real    0m23.271s
user    0m0.032s
sys     0m0.024s
```

jsha:
Here's output from a run using your modified tool:

```
$ RUST_LOG=trace time ./target/debug/examples/shnatsel-smoke-test shuyang.gov.cn
Fetching http://shuyang.gov.cn

Did not hang! Error: http://shuyang.gov.cn/: Too Many Redirects
Command exited with non-zero status 1
0.01user 0.00system 1:29.31elapsed 0%CPU (0avgtext+0avgdata 5232maxresident)k
0inputs+0outputs (0major+275minor)pagefaults 0swaps
```

[...]

PR #313, opened 2021-02-06
Apply deadline across redirects.

jsha:
Previously, each redirect could take `timeout` time, so a series of slow
redirects could run for longer than expected, or indefinitely.

Related #312 /cc @Shnatsel


algesten (approved):
looks good!

PR #324, opened 2021-02-15
Allow either LF or CRLF line ends for header fields

### LINKED ISSUE #321: "Header field didn't end with \r" error on some websites
On some websites, e.g. etihad.com, ureq fails with the following error:

> Header field didn't end with \r

Firefox and curl work fine.

There's 295 such websites in the top million (I'm using [Tranco list generated on the 3rd of February](https://tranco-list.eu/list/3G6L)).

Archive with all occurrences: [ureq-header-did-not-end-well.tar.gz](https://github.com/algesten/ureq/files/5983467/ureq-header-did-not-end-well.tar.gz)

nico-abram:


jsha (changes requested):
Welcome to the repo, @nico-abram! Thanks for finding the reference on this. Would you mind adding a test case?

nico-abram:
Added a test case

jsha (approved):
Thanks!

algesten:
Being lenient seems good. Thanks!

PR #327, opened 2021-02-21
Allow status lines with missing reason phrase

### LINKED ISSUE #316: "Bad Status: Wrong number of tokens in status line" error on some websites
On some websites, e.g. http://hack.pl, ureq fails with the following error: 

> Bad Status: Wrong number of tokens in status line

However, curl and Firefox work fine.

There's 296 such websites in the top million (I'm using [Tranco list generated on the 3rd of February](https://tranco-list.eu/list/3G6L)).

Archive with all occurrences: [ureq-bad-status.tar.gz](https://github.com/algesten/ureq/files/5978826/ureq-bad-status.tar.gz)

algesten:
The spec says the reason phrase at least must be a space. However in
the wild, there are sites that just ends after the status code.
To be more compatible, this commit relaxes ureq's parsing.

Close #316

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -471,7 +471,7 @@ fn parse_status_line(line: &str) -> Result<(ResponseStatusIndex, u16), Error> {
     //      status-line = HTTP-version SP status-code SP reason-phrase CRLF
     let mut split: Vec<&str> = line.splitn(3, ' ').collect();
     if split.len() == 2 {
-        // As a special case, we are lenient parsing lines without reason phrase.
+        // As a special case, we are lenient parsing lines without a space after the code.
         // This is technically against spec. "HTTP/1.1 200\r\n"
         split.push("");
     }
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -469,7 +469,12 @@ fn parse_status_line(line: &str) -> Result<(ResponseStatusIndex, u16), Error> {
     }
     // https://tools.ietf.org/html/rfc7230#section-3.1.2
     //      status-line = HTTP-version SP status-code SP reason-phrase CRLF
-    let split: Vec<&str> = line.splitn(3, ' ').collect();
+    let mut split: Vec<&str> = line.splitn(3, ' ').collect();
+    if split.len() == 2 {
+        // As a special case, we are lenient parsing lines without reason phrase.
```
jsha, line 474: Minor tweak:

```suggestion
        // As a special case, we are lenient parsing lines without a space after the code.
```

(because the reason-phrase itself is indeed optional)

jsha (approved):

PR #329, opened 2021-02-21
Provide context for errors when reading status line and headers

### LINKED ISSUE #320: "Network Error: stream did not contain valid UTF-8" error, even with 'charset' enabled
On some websites, e.g. paneliste.com, ureq fails with the following error:

> Network Error: stream did not contain valid UTF-8

despite the `charset` feature being enabled.

Firefox and curl work fine.

There's 96 such websites in the top million (I'm using [Tranco list generated on the 3rd of February](https://tranco-list.eu/list/3G6L)).

Archive with all occurrences: [ureq-did-not-contain-valid-utf8.tar.gz](https://github.com/algesten/ureq/files/5983425/ureq-did-not-contain-valid-utf8.tar.gz)

algesten:
The HTTP spec allows for non-ascii values both in the status line and
in headers. ureq does not handle that, we can however provide better
context for when it happens.

Close #320

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -8,9 +8,9 @@ use url::Url;
 use crate::error::{Error, ErrorKind::BadStatus};
 use crate::header::Header;
 use crate::pool::PoolReturnRead;
-use crate::stream;
 use crate::stream::{DeadlineStream, Stream};
 use crate::unit::Unit;
+use crate::{stream, ErrorKind};
 
 #[cfg(feature = "json")]
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -539,9 +539,30 @@ impl FromStr for Response {
...
+    if let Err(e) = result {
+        // Provide context to errors encountered while reading the line.
+        // ureq does not currently handle non-ascii status lines and
+        // header values. For historical reasons, the HTTP spec does
+        // allow for characters in the range 0x80-0xff, but these are
+        // very rarely encountered in the wild.
+        // See https://github.com/algesten/ureq/issues/320
+        let reason = if e.kind() == io::ErrorKind::InvalidData {
+            format!(
+                "Invalid data in {} (see issue #320): {}",
```
jsha, line 555: People are likely to see this error in contexts where it's not clear it comes from ureq, so `issue #320` will be confusing. I would recommend omitting it; if you'd like to include it, let's use a full URL.

src/response.rs
```diff
@@ -539,9 +539,30 @@ impl FromStr for Response {
...
+                context,
+                e.to_string()
```
jsha, line 557: I think if we add `.src(e)` to the error, we don't need to interpolate it as a string here.

jsha (changes requested):
Looks good!  A couple of tweaks below.

src/response.rs
algesten: Ok!

src/response.rs
algesten: Ok!

algesten:
@jsha This should be ok now. If you agree I squash, rebase and merge.

jsha (approved):

PR #335, opened 2021-02-26
Introduce Request::timeout to override agent's config

### LINKED ISSUE #334: 2.0: Restore per request timeouts or allow Agent to re-configure timeouts?
Subj.
I would like to have option to dynamically change timeouts in my application.
With current approach I would need to re-create the whole agent from a scratch each time timeout is changed. (it is not desirable since `Agent` is field in my case)
I would prefer to have option to change it also in Agent or override in requests?

Let me know which is works for you @algesten

DoumanAsh:
Fixes #334

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -40,6 +40,7 @@ pub struct Request {
     error_on_non_2xx: bool,
     headers: Vec<Header>,
     query_params: Vec<(String, String)>,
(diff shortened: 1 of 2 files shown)


algesten:
@DoumanAsh Looks good!

Let's merge!

PR #340, opened 2021-03-13
Url access functions for Request

algesten:
Land #310 first

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -323,14 +323,6 @@ impl From<ParseError> for Error {
     }
 }
 
-impl From<&ParseError> for Error {
-    fn from(err: &ParseError) -> Self {
-        ErrorKind::InvalidUrl
-            .msg(&format!("failed to parse URL: {:?}", err))
-            .src(err.clone())
-    }
-}
-
 impl fmt::Display for ErrorKind {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
--- a/src/error.rs
+++ b/src/error.rs
@@ -323,6 +323,14 @@ impl From<ParseError> for Error {
     }
 }
 
+impl From<&ParseError> for Error {
+    fn from(err: &ParseError) -> Self {
+        ErrorKind::InvalidUrl
+            .msg(&format!("failed to parse URL: {:?}", err))
+            .src(err.clone())
+    }
+}
+
 impl fmt::Display for ErrorKind {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
--- a/src/request.rs
+++ b/src/request.rs
@@ -21,7 +21,10 @@ pub struct ParsedUrl {
     /// Url parse result kept as Result since we want to defer the surfacing of it
     /// until we do call()/send()/etc.
     result: std::result::Result<Url, url::ParseError>,
+    /// The original &str url, or None if constructed via url::Url.
     origin: Option<Arc<String>>,
+    /// Query pairs copied out from Cow<str> in url::Url when using Request::parsed_url().
+    query_pairs: Option<Vec<(String, String)>>,
 }
 
 impl fmt::Display for ParsedUrl {
@@ -59,7 +62,27 @@ impl ParsedUrl {
 
         let origin = origin.map(Arc::new);
 
-        ParsedUrl { result, origin }
+        ParsedUrl {
+            result,
+            origin,
+            query_pairs: None,
+        }
+    }
+
+    /// Prepares the ParsedUrl to return an owned copy from `Request::parsed_url()`.
+    fn cloned_with_query_pairs(&self) -> Self {
+        let mut clone = self.clone();
+        let mut query_pairs = vec![];
+
+        if let Ok(url) = &self.result {
+            for (k, v) in url.query_pairs() {
+                query_pairs.push((k.into(), v.into()));
+            }
+        }
+
+        clone.query_pairs = Some(query_pairs);
+
+        clone
     }
 
     // NOTICE:
@@ -110,17 +133,19 @@ impl ParsedUrl {
     ///     .query("foo", "43");
     ///
     /// assert_eq!(req.parsed_url().unwrap().query_pairs(), vec![
-    ///     ("foo".to_string(), "42".to_string()),
-    ///     ("foo".to_string(), "43".to_string())
+    ///     ("foo", "42"),
+    ///     ("foo", "43")
     /// ]);
     /// # Ok(())
     /// # }
     /// ```
-    pub fn query_pairs(&self) -> Vec<(String, String)> {
+    pub fn query_pairs(&self) -> Vec<(&str, &str)> {
         let mut ret = vec![];
 
-        for (k, v) in self.as_url().query_pairs() {
-            ret.push((k.into(), v.into()));
+        if let Some(v) = &self.query_pairs {
+            for (k, v) in v {
+                ret.push((k.as_str(), v.as_str()));
+            }
         }
 
         ret
@@ -475,7 +500,7 @@ impl Request {
     /// # }
     /// ```
     pub fn parsed_url(&self) -> Result<ParsedUrl> {
-        let p = self.parsed_url.clone();
+        let p = self.parsed_url.cloned_with_query_pairs();
 
         // If there is a parse error, surface it now.
         if let Err(e) = p.result {
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,7 +1,7 @@
 use std::io::Read;
 use std::{fmt, time};
 
-use url::{form_urlencoded, Url};
+use url::{form_urlencoded, ParseError, Url};
 
 use crate::body::Payload;
 use crate::header::{self, Header};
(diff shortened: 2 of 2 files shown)


algesten:
@jsha Here are the url accessor functions. They sort of use the ideas from the original PR, but in a cleaner way. And there's a `parsed_url()` to get the `Url` variant.

algesten:
@jsha this on is ready to land now that #310 is out of the way. Any comments?

algesten:
@jsha gentle reping on this one. i like to land it and release a new version.

jsha:
Will take a look. Thanks for the ping!

jsha (commented):
I like this in general! One small issue I have: `Request` is a builder, so most methods like `.foo()` will set some field called `foo` on it. But this introduces the method `.url()` which doesn't set `url` but gets it. The same is true of `.method()` and `.query_params()`, which we've already merged but haven't yet release. What do you think of prefixing getters with `get_`?

Also, I have an idea to make a wrapper class in ureq, something like `ParsedUrl`, that would wrap `Url` and provide the nice types we want. If I can get that to you by this evening PDT, do you mind waiting and considering it as an alternate?

algesten:
@jsha I've pushed an update to this PR that tries to reconcile what you did in #348 with the work I've been doing here.

> One small issue I have: Request is a builder, so most methods like .foo() will set some field called foo on it. But this introduces the method .url() which doesn't set url but gets it.

I hear you, however I feel the boat has already sailed on this one because of `.set()` and `.header()` which are already public API. I think we can revisit this for a future 3.0 breaking change.

In order to not introduce _yet more_ naming schemes (get_xxx), I'm zooming in on the already present `.header()` and think `.method()` and `.url()` should be congruent with that.

> Also, I have an idea to make a wrapper class in ureq, something like ParsedUrl, that would wrap Url and provide the nice types we want.

I'm open to `RequestUrl` (or `ParsedUrl`) being part of the public API and maybe also moving `query_pairs()` (which is a better name than `query_params`), so I did that also in this PR. However while `ParsedUrl` adds some nice inspections, it doesn't solve `.method()`.

On some level I _really_ want to avoid the inevitable `Result` that comes with `ParsedUrl`, so i find it unfortunate that `query_pairs()` gets "gated" behind it, but I can see the logic of why it would.

I think `Request::url()` as I have it is good and also `Request::method()`, so ideally I'd like to keep them in addition to this `Request::parsed_url()`.

The changes in this commit tries to find a middle ground between going all i on `ParsedUrl` and keeping some inspection stuff in `Request`.

src/request.rs
```diff
@@ -14,28 +14,113 @@ use super::SerdeValue;
...
+    ///     .query("foo", "43");
+    ///
+    /// assert_eq!(req.parsed_url().unwrap().query_pairs(), vec![
+    ///     ("foo".to_string(), "42".to_string()),
+    ///     ("foo".to_string(), "43".to_string())
+    /// ]);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn query_pairs(&self) -> Vec<(String, String)> {
```
jsha, line 116: I propose to make this return `Vec<(&str, &str)>`, because the caller doesn't generally need owned strings. `&str` is more consistent with the rest of our API, and leaves the door open to a more optimized implementation later. In #348 I did this by storing a copy of the query pairs in the Url object on construction.

src/request.rs
```diff
@@ -368,32 +453,90 @@ impl Request {
...
-    ///     ("foo".to_string(), "42".to_string()),
-    ///     ("foo".to_string(), "43".to_string())
-    /// ]);
+    /// assert_eq!(req.parsed_url().unwrap().host(), "httpbin.org");
     /// # Ok(())
     /// # }
     /// ```
-    pub fn query_params(&self) -> Vec<(String, String)> {
-        let mut ret = vec![];
+    pub fn parsed_url(&self) -> Result<&ParsedUrl> {
```
jsha, line 474: On the other hand, here, I think it makes sense to return an owned `Result<ParsedUrl>`.

src/request.rs
```diff
@@ -14,28 +14,113 @@ use super::SerdeValue;
...
 pub type Result<T> = std::result::Result<T, Error>;
 
+/// Parsed result of a request url with handy inspection methods.
 #[derive(Clone)]
-struct ParsedUrl(std::result::Result<Url, url::ParseError>);
+pub struct ParsedUrl {
+    /// Url parse result kept as Result since we want to defer the surfacing of it
+    /// until we do call()/send()/etc.
+    result: std::result::Result<Url, url::ParseError>,
+    origin: Option<String>,
```
jsha, line 23: I think we don't need to carry around the original input string. It looks like the only place we currently use it is in the `.url()` accessor, so we can return the original in case it failed to parse. But I can't think of a case where it's useful to get back the original malformed URL in that case, since the request is doomed to fail anyhow. Why not define `.url()` to return "" if the input URL failed to parse?

src/request.rs
```diff
@@ -368,32 +453,90 @@ impl Request {
...
+                // This should not happen since we either created Request using a valid Url
+                // object, in which case we have Ok(url) above, or we used a String, in
+                // which case we definitely have a value in self.parsed_url.1
+                .expect("Illegal URL without original")
         }
+    }
 
-        ret
+    /// Test whether the url for this request is valid. If this test is
+    /// `false`, the resulting use of this Request will fail with an error.
```
jsha, line 527: This feels like an awkward function to use. What's the expected use case?

src/error.rs
```diff
@@ -323,6 +323,14 @@ impl From<ParseError> for Error {
     }
 }
 
+impl From<&ParseError> for Error {
```
jsha, line 326: Why is this From implemented on a ref instead of an owned error?

src/request.rs
```diff
@@ -14,28 +14,113 @@ use super::SerdeValue;
...
 
 impl From<Url> for ParsedUrl {
     fn from(url: Url) -> Self {
-        ParsedUrl(Ok(url))
+        ParsedUrl::new(Ok(url), None)
+    }
+}
+
+impl ParsedUrl {
+    fn new(mut result: std::result::Result<Url, ParseError>, origin: Option<String>) -> Self {
```
jsha, line 51: I think `mut` is unnecessary here.

jsha (changes requested):
Thanks for the update! A few thoughts below:

src/request.rs
algesten: Ideally I want `Vec<(&str, &str)>`, but, as you point out, that forces us to do the String allocation on construction, effectively penalizing every use of `Request`, also those that don't want this.

However, an escape hatch could be to only do this when someone calls `parsed_url()`. I'll look into it.

src/request.rs
algesten: However that forces us to clone. Is that really to prefer?

src/request.rs
algesten: I just really want that original url. I feel quite strongly about it. Parsing, technically, should only happen once we do `call`, `send` etc. The only reason it surfaces sooner would be on `parsed_url()`.

Any parsed (correct?) variant is extra topping for me. Not the actual base API.

src/request.rs
algesten: Let's remove it.

src/error.rs
algesten: It's implemented for both because `parsed_url` returns a ref. We can remove it if we decide to clone to an owned ParsedUrl.

src/request.rs
algesten: Given the below changing of `result` if host is empty, it is required.

src/request.rs
algesten: Done

src/request.rs
algesten: Gone.

src/error.rs
algesten: Gone

src/request.rs
jsha: Fair enough. In that case, what do you think of making this struct an `enum { Str(String), Url(Url), Error(url::ParseError) }`, since it's guaranteed to contain exactly one of those?

Also, I think the term `origin` should be called `original` because "origin" has a technical meaning on the web (the scheme + authority).

src/request.rs
algesten: Done

src/request.rs
algesten: > Fair enough. In that case, what do you think of making this struct an enum { Str(String), Url(Url), Error(url::ParseError) }, since it's guaranteed to contain exactly one of those?

Ok. I'll fix!

> Also, I think the term origin should be called original because "origin" has a technical meaning on the web (the scheme + authority).

Good point! Helps me to have a native speaker checking my language. I fix!

jsha: > what do you think of making this struct an enum { Str(String), Url(Url), Error(url::ParseError) }, since it's guaranteed to contain exactly one of those?

I realized I was wrong about this! I this is guaranteed to contain one of:

1) a String _and_ a ParseError
2) a Url

So maybe `enum { Str(String, ParseError), Url(Url) }`?

jsha (commented):
Another thing that occurred to me: We could store the canonical version of the request URL as String. That would mean when someone calls `request_url`, we would serialize to string and then re-parse as URL later. I was reluctant to do that originally because rounds up serializing and re-parsing tend to lead to bugs if things don't round-trip fully. But a version of round-tripping is specified by the URL Standard: https://url.spec.whatwg.org/#goals

> Ensure the combination of parser, serializer, and API guarantee idempotence. For example, a non-failure result of a parse-then-serialize operation will not change with any further parse-then-serialize operations applied to it. Similarly, manipulating a non-failure result through the API will not change from applying any number of serialize-then-parse operations to it. 

And the `url` crate is documented to implement the URL Standard: https://docs.rs/url/2.2.1/url/:

> rust-url is an implementation of the URL Standard for the Rust programming language.

src/request.rs
algesten: @jsha 

> So maybe `enum { Str(String, ParseError), Url(Url) }`?

I'm trying to go down this path, but tbh, it seems … awkward.

Here's what I try to do:

```rust
/// Encapsulation of the various state a url can be in.
#[derive(Clone)]
enum UrlState {
    /// String has not yet been parsed to a Url yet.
    Unparsed(Arc<String>),
    /// A successfully parsed url.
    Parsed(Url),
    /// Special state when using `Request::parsed_url().
    /// Query pairs copied out from Cow<str> in url::Url.
    ParsedAndQueryPairs(ParsedAndQueryPairs),
    /// A url that failed to parse. The string is the original we tried to parse.
    ParseError(Arc<String>, ParseError),
}

#[derive(Clone)]
struct ParsedAndQueryPairs {
    url: Url,
    query_pairs: Vec<(String, String)>,
}
```

And then try to encapsulate all state transitions neatly in a `impl UrlState`. But it's total code explosion. I've pushed it, but I think it makes things more complicated.

algesten:
> We could store the canonical version of the request URL as String. That would mean when someone calls request_url, we would serialize to string and then re-parse as URL later.

One good thing I think this PR does, is to make do with the `query_pairs` local var in `Request`. Whenever we do `Request::query()` to add another one, we push that all the way down into `Url`. The benefits are that we get one single view of the url also with these modifications.

Having a `String` internally would undo that, or force us to parse/serialize on every `query`, but maybe could be worth it?

src/request.rs
jsha: Yeah, I agree this makes things way more complicated, and I preferred the previous version. Thanks for trying it out.

jsha:
> One good thing I think this PR does, is to make do with the query_pairs local var in Request. Whenever we do Request::query() to add another one, we push that all the way down into Url. The benefits are that we get one single view of the url also with these modifications.

I think the version of this on `main` is OK, and defines a single view of the URL. It takes a couple of steps to produce that view (parse; add query_params), but it's consistent and well defined.

> Having a String internally would undo that, or force us to parse/serialize on every query, but maybe could be worth it?

My first preference is to keep a separate query_params Vec (i.e. what's in main) so we don't have to parse/serialize on each call to `.query()`. But I could be convinced to go another way. I think this is mainly a tradeoff on internal code complexity vs performance, and doesn't have implications for the external API, so we can pick something and modify later based on performance testing or changes in our implementation preferences.

algesten:
> My first preference is to keep a separate query_params Vec (i.e. what's in main) so we don't have to parse/serialize on each call to .query(). But I could be convinced to go another way.

This is what I mean with "One good thing I think this PR does, is to make do with the query_pairs local var in Request". The way `main` is right now, we have a double storage of query parameters. Some that are passed in via the request url, and others that are added via `Request::query()`. The good thing this PR does is removes this doubleness and just uses the underlying `Url`.

The advantage of this PR is that any call to `Request::url()` and `Request::parsed_url()` directly reflects the changes made via `Request::query()` without any further work.

I especially dislike that `main` uses _both_ Url _and_ a separate query_params Vec _at the same time_ – since that does result in inconsistencies. For example, `query_pairs()` should return the parameters from both. I can sort of get behind using _only_ a separate query_params Vec (which is where this PR started). Though I think this PR got a hella lot better when we ditched the separate query params vec.

algesten:
If you want I can amend this PR to use a single `String` storage that is temporarily parsed to `Url` when needed (on `.query()`

jsha:
> I especially dislike that main uses both Url and a separate query_params Vec at the same time – since that does result in inconsistencies. For example, query_pairs() should return the parameters from both.

I definitely agree that `query_pairs()` must return the parameters from both the original URL and any ones the user added with `.query()`. I think it's possible to do that consistently, even with the double storage of query parameters. It just means we need to parse the URL and merge the query params into it (which yield a possible error we need to deal with).

>  I can sort of get behind using only a separate query_params Vec (which is where this PR started).

So `Request` would have a `url: String` and a `query_params: Vec`? That sounds good to me. Then we don't need to parse/serialize on every call to `.query("foo", "bar")`.

BTW I realize it's getting late for you. If you want to schedule some time to work through this live tomorrow, I'm happy to do that; I know this is something you're planning to use for your work soon so I want to get it squared away in a timely way.

jsha:
I'm also happy to prepare a branch that incorporates the ideas I'm talking about above. By the way:

> On some level I really want to avoid the inevitable Result that comes with ParsedUrl, so i find it unfortunate that query_pairs() gets "gated" behind it, but I can see the logic of why it would.

In the current branch, there is `pub fn parsed_url(&self) -> Result<ParsedUrl>`. Should I take that to mean you're okay with the `Result<ParsedUrl>`, or are you still looking for ways to eliminate the `Result` and just return `ParsedUrl`?

algesten:
> BTW I realize it's getting late for you. If you want to schedule some time to work through this live tomorrow, I'm happy to do that;

Thanks! I'm actually on holiday this week, hence I got time to look at this. 🙈
However I am fully expecting to be out hiking all day tomorrow, but thanks! I might pick you up on it next week when I'm back to work.

> Should I take that to mean you're okay with the `Result<ParsedUrl>`

Yep. I'm ok with that. However #350 nicely sidesteps the entire issue. :P

algesten:
Closing in favor of #350

PR #342, opened 2021-03-14
Add -k option to cureq example

jsha:
By analogy with curl, this turns off certificate verification. Requires
enabling the "dangerous_configuration" feature in the rustls dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ cookie = { version = "0.14", features = ["percent-encode"], optional = true}
 once_cell = "1"
 url = "2"
 socks = { version = "0.3.2", optional = true }
-rustls = { version = "0.19", optional = true, features = [] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -30,7 +30,7 @@ cookie = { version = "0.14", features = ["percent-encode"], optional = true}
 once_cell = "1"
 url = "2"
 socks = { version = "0.3.2", optional = true }
-rustls = { version = "0.19", optional = true, features = [] }
+rustls = { version = "0.19", optional = true, features = ["dangerous_configuration"] }
```
algesten, line 33: Rather than doing this by default, I think you can add another `rustls` import in the `dev-dependencies` that pulls in the extra feature.

algesten (changes requested):

algesten (approved):
LGTM!

PR #343, opened 2021-03-14
Improve handling of response body sizes

jsha:
Remove consideration of `Connection: close` in determining response body
size. A response can have Content-Length in addition to Connection:
close.

Remove consideration of HTTP version. An HTTP/1.0 response can have a
Content-Length.

Add consideration of Transfer-Encoding, which is specified to take
precedence over Content-Length.

No-body response codes take precedence over any Transfer-Encoding that
might be present, and enforce a body size of 0.

Explicitly reject a Content-Length header on no-body response codes.

I noticed this when testing against a server that sent `HTTP/1.0`, `Connection: close`, and a `Content-Length`, but didn't actually close the connection; this caused cureq to wait indefinitely.

Draft because I need to write unit tests.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 use std::{fmt, io::BufRead};
 
 use chunked_transfer::Decoder as ChunkDecoder;
+use log::trace;
 use url::Url;
 
 use crate::error::{Error, ErrorKind::BadStatus};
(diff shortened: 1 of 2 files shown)


src/response.rs
```diff
@@ -269,11 +287,22 @@ impl Response {
         let stream = DeadlineStream::new(stream, deadline);
 
         match (use_chunked, limit_bytes) {
-            (true, _) => Box::new(PoolReturnRead::new(unit, ChunkDecoder::new(stream))),
+            (_, Some(0)) => {
+                trace!("returning zero-length response body stream");
+                Box::new(PoolReturnRead::new(unit, LimitedRead::new(stream, 0)))
+            }
+            (true, _) => {
+                trace!("building ChunkDecoder");
```
algesten, line 295: maybe all these trace! logs can be on the form "Using ChunkDecoder", "Using LimitRead with limit {}", etc.

algesten (commented):
I like where this is going.

jsha:
Obsoleted by #625

PR #344, opened 2021-03-14
Ensure Error and Response implement Send+Sync

algesten:
This is to ensure we don't accidentally introduce breaking changes
with respects to Send and Sync.


algesten:
@jsha Some backwards compat checking.

jsha (approved):

PR #345, opened 2021-03-14
Move unit tests inside conditionally compiled mod tests { } blocks

algesten:
Idiomatic rust organizes unit tests into `mod tests { }` blocks
using conditional compilation `[cfg(test)]` to decide whether to
compile the code in that block.

This commit moves "bare" test functions into such blocks, and puts
the block at the bottom of respective file.


jsha (approved):
👍🏻

PR #347, opened 2021-03-14
Handle non-utf8 status and headers

### LINKED ISSUE #341: stream did not contain valid UTF-8 on fastcompany URL
In https://github.com/algesten/ureq/issues/320#issuecomment-795423427, @kl reports:
> I'm running in to this as well with this: https://www.fastcompany.com/90601037/american-surviellance-state-facial-recognition-hypocrisy

> I have the following:
>
```
let agent = AgentBuilder::new().timeout(Duration::from_secs(20)).build();
let resp = agent.get(link).call().context("failed to get article")?;
let doc = Document::from_read(resp.into_reader())?;
```

> but the result from call() is Err so I'm getting this error:

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Transport(Transport { kind: Io, message: None, url: Some(Url { scheme: "https", username: "", password: None, host: Some(Domain("www.fastcompany.com")), port: None, path: "/90601037/american-surviellance-state-facial-recognition-hypocrisy", query: None, fragment: None }), source: Some(Custom { kind: InvalidData, error: "stream did not contain valid UTF-8" }), response: None })', src/resolve.rs:47:39

> So this is on the line before resp.into_reader().

@kl, I tried to reproduce this with cureq (https://github.com/algesten/ureq/blob/main/examples/cureq/main.rs), with this command line:

    cargo build --example cureq --features="charset cookies socks-proxy"
    ./target/debug/examples/cureq https://www.fastcompany.com/90601037/american-surviellance-state-facial-recognition-hypocrisy

But it succeeded for me. Can you tell me what version of ureq you're on, and if it's not the latest, can you try upgrading?

Also, I'm not able to try your example code because I don't have the `Document` struct. Would you mind sharing a full self-contained example? Thanks!

algesten:
For header values, there is a Response::header_raw() function to obtain the &[u8]
bytes. For status_text, we currently don't provide a method of obtaining
the raw bytes.

Close #341

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -1,6 +1,66 @@
 use crate::error::{Error, ErrorKind};
 use std::fmt;
-use std::str::FromStr;
+use std::str::{from_utf8, FromStr};
+
+/// Since a status line or header can contain non-utf8 characters the
+/// backing store is a `Vec<u8>`
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct HeaderLine(Vec<u8>);
+
+impl From<String> for HeaderLine {
+    fn from(s: String) -> Self {
+        HeaderLine(s.into_bytes())
+    }
+}
+
+impl From<Vec<u8>> for HeaderLine {
+    fn from(b: Vec<u8>) -> Self {
+        HeaderLine(b)
+    }
+}
+
+impl HeaderLine {
+    pub fn into_string_lossy(self) -> String {
+        // Try to avoid an extra allcation.
+        String::from_utf8(self.0)
+            .unwrap_or_else(|e| String::from_utf8_lossy(&e.into_bytes()).to_string())
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    pub fn into_header(self) -> Result<Header, Error> {
+        // The header name should always be ascii, we can read anything up to the
+        // ':' delimiter byte-by-byte.
+        let mut index = 0;
+
+        for c in self.as_bytes() {
+            if *c == b':' {
+                break;
+            }
+            if !is_tchar(c) {
+                return Err(Error::new(
+                    ErrorKind::BadHeader,
+                    Some(format!("Invalid char ({:0x?}) while looking for ':'", *c)),
+                ));
+            }
+            index += 1;
+        }
+
+        Ok(Header { line: self, index })
+    }
+}
+
+impl fmt::Display for HeaderLine {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", String::from_utf8_lossy(&self.0))
+    }
+}
 
 #[derive(Clone, PartialEq)]
 /// Wrapper type for a header field.
@@ -8,7 +68,7 @@ use std::str::FromStr;
 pub struct Header {
     // Line contains the unmodified bytes of single header field.
     // It does not contain the final CRLF.
-    line: String,
+    line: HeaderLine,
     // Index is the position of the colon within the header field.
     // Invariant: index > 0
     // Invariant: index + 1 < line.len()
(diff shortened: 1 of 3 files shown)


jsha:
Thanks for working on this change! I definitely want to fix this, and the linked issue is a good demonstration why.

I'm uneasy at the use of lossy conversion. I think it will always be the wrong thing for non-UTF-8 text, and additionally it introduces the prospect of someone maliciously taking advantage of that lossy conversion. Also it requires us to do a lot of allocations: something that is in general not our first priority to optimize, but I think for this particular case it might be starting to get a little too expensive.

Let's consider the API we want. We definitely want to return `&str` for header fields, because in the 99% use case, they are valid `&str`. We definitely want to return non-error if a header field contains non-ASCII.

Do we want to allow users to get a `&[u8]` for header fields that are allowed to have non-ASCII bytes? Maybe. Do we want to process `Set-Cookie` headers that contain non-ASCII? Probably.

Here's what RFC 7230 says (https://tools.ietf.org/html/rfc7230#page-26):

>    Historically, HTTP has allowed field content with text in the
>   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
>   through use of [RFC2047] encoding.  In practice, most HTTP header
>   field values use only a subset of the US-ASCII charset [USASCII].
>   Newly defined header fields SHOULD limit their field values to
>   US-ASCII octets.  A recipient SHOULD treat other octets in field
>   content (obs-text) as opaque data.

So allowance of non-ASCII is on a per-header basis. Set-Cookie in particular allows it (https://tools.ietf.org/html/rfc6265#page-26):

>    NOTE: Despite its name, the cookie-string is actually a sequence of
>   octets, not a sequence of characters.  To convert the cookie-string
>   (or components thereof) into a sequence of characters (e.g., for
>   presentation to the user), the user agent might wish to try using the
>   UTF-8 character encoding [RFC3629] to decode the octet sequence.
>   This decoding might fail, however, because not every sequence of
>   octets is valid UTF-8.

Back to 7230, we have this useful text:

>   Messages are parsed using a generic algorithm, independent of the
>   individual header field names.  The contents within a given field
>   value are not parsed until a later stage of message interpretation
>   (usually after the message's entire header section has been
>   processed)

Here's what I think we should do: When processing the header lines, store the values in `Vec<u8>`. When a user calls, e.g. `.header("foo")`, we call [`str::from_utf8`](https://doc.rust-lang.org/std/str/fn.from_utf8.html), and if the vec is not valid UTF-8, return `None` (i.e. pretend the header doesn't exist, because it failed validation).

For headers that we know are allowed to contain non-UTF-8, we can either handle them internally (like Set-Cookie) or provide a header-specific accessor that returns either `&[u8]` or some header-specific representation.

algesten:
@jsha 

> I'm uneasy at the use of lossy conversion. I think it will always be the wrong thing for non-UTF-8 text, and additionally it introduces the prospect of someone maliciously taking advantage of that lossy conversion. 

I hear you. I don't quite get how it could be used, but I'm not going to push for having the lossy conversion.

> Also it requires us to do a lot of allocations: something that is in general not our first priority to optimize, but I think for this particular case it might be starting to get a little too expensive.

It actually not much more (I was taking care here).

1. Reading the underlying stream to a `Vec<u8>` is same as reading to a `String`, but doesn't check utf-8.
2. Doing `String::from_utf8(Vec<u8>)` checks utf-8, then reuse the same buffer for the String. No extra allocation.
3. Finding the index of `:` is as cheap over a `Vec<u8>` as `String`.
4. `String::as_str()` is mainly used now (as before) to access the `String`.

There is one more allocation, which is for Raw headers – `from_lossy_utf8` for the (rare) non-utf8 case.

> Here's what I think we should do: When processing the header lines, store the values in Vec<u8>. When a user calls, e.g. .header("foo"), we call str::from_utf8, and if the vec is not valid UTF-8, return None (i.e. pretend the header doesn't exist, because it failed validation).

I can amend this PR to return `None` for non-utf8 headers. That would remove the need for the extra `from_lossy_utf8` allocation.

algesten:
@jsha I pushed a commit that:

1. Uses a single `Vec<u8>` backing store.
2. Returns `None` for headers that cannot be represented as utf-8.

jsha (approved):
This looks good!

>> I'm uneasy at the use of lossy conversion.
> I don't quite get how it could be used, but I'm not going to push for having the lossy conversion.

I was mistaken here - I thought that the lossy conversion would remove invalid bytes, but on checking the documentation I see that it replaces them. So, yes, I'm not sure it could be exploited. But I do like this version somewhat better if you are happy with it.

I would be inclined to not add `header_raw` to the public API until we know we need it and have a concrete use case. The current bug we are fixing just requires that we not error the request due to non-ASCII in a Cookie header (and, optionally, we can improve Cookie handling to handle that case correctly).

The main use case I imagine for `header_raw` is if someone is using `ureq` in a context like a proxy where they want to pass along all headers unchanged. In that context, we would need something slightly different, like an iterator that goes over all headers, in order, and provides the raw content.

That said, I am not too concerned over this addition to the public API. If you feel it's needed I'm happy to defer to your judgement here.

algesten:
@jsha 

> I would be inclined to not add `header_raw` to the public API until we know we need it and have a concrete use case. The current bug we are fixing just requires that we not error the request due to non-ASCII in a Cookie header (and, optionally, we can improve Cookie handling to handle that case correctly).

Happy to remove it. However, if we make no way of accessing these headers, do you think `headers_names` should return the name for them? It would be quite expensive to check validity in `header_names`.

jsha:
>  if we make no way of accessing these headers, do you think headers_names should return the name for them? It would be quite expensive to check validity in header_names.

I think it's okay for `headers_names` to return a header name where `.header(name)` returns None. That conveys a specific semantic: there is a header here, but we don't know what charset to interpret it as.

By the way, we should actually return None for non-ASCII field values, rather than non-UTF-8 field values. There are ISO-8859-1 sequences that can be incorrectly interpreted as UTF-8. For instance: C2 A5 is ¥ in UTF-8 and Â¥ in ISO-8859-1.

Another possibility that I just realized: We could take an unconditional dependency on `encoding_rs`, and always interpret non-ASCII bytes as ISO-8859-1. I think that's less preferable than the current approach, but putting it out there for consideration.

algesten:
> I think it's okay for headers_names to return a header name where .header(name) returns None. That conveys a specific semantic: there is a header here, but we don't know what charset to interpret it as.

Fair enough.

> By the way, we should actually return None for non-ASCII field values, rather than non-UTF-8 field values. There are ISO-8859-1 sequences that can be incorrectly interpreted as UTF-8. For instance: C2 A5 is ¥ in UTF-8 and Â¥ in ISO-8859-1.

Hm. Good point.

PR #348, opened 2021-03-18
Add RequestUrl

jsha:
This provides a way to access the URL that will be used for a
currently-being-built Request. This is particularly useful for building
Requests that need signing, since you can access the parsed values of
query parameters.

@algesten this is my proposed alternative to #340. My reasoning is
that a user who wants access to the query pairs is likely to also want
the scheme, host, path, etc. By putting them on one object we can keep
the cognitive burden of Request small, while still making accessors that
return the types we want (&str).


jsha:
Closing in preference for #340.

PR #349, opened 2021-03-23
Remove specialized error message for complete_io.

jsha:
This was triggering for errors like InvalidCertificate and giving the
message "Sometimes this means the host doesn't support any of the same
ciphersuites as rustls, or doesn't support TLS 1.2 and above," which is
confusing in that situation.

For now, offer no specialized error. I'd like to come back when I find
more time and restore this, but only for the error cases where it's
useful.


algesten (approved):
👍

PR #350, opened 2021-03-23
Url access functions for Request (simpler)

algesten:
@jsha Here's a take from scratch with keeping an internal `url: String` and no parsed state.

It keeps `query_pairs` in sync with `url`, and there is parsing on every `query()` call, but I actually quite like this one a lot because it's simple.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,7 +1,7 @@
 use std::io::Read;
 use std::{fmt, time};
 
-use url::{form_urlencoded, Url};
+use url::{form_urlencoded, ParseError, Url};
 
 use crate::body::Payload;
 use crate::header::{self, Header};
@@ -13,31 +13,6 @@ use crate::{agent::Agent, error::Error};
 use super::SerdeValue;
 
 pub type Result<T> = std::result::Result<T, Error>;
-
-#[derive(Clone)]
-struct ParsedUrl(std::result::Result<Url, url::ParseError>);
-
-impl fmt::Display for ParsedUrl {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if let Ok(url) = &self.0 {
-            write!(f, "{}", url.as_str())
-        } else {
-            write!(f, "{:?}", self.0)
-        }
-    }
-}
-
-impl From<String> for ParsedUrl {
-    fn from(s: String) -> Self {
-        ParsedUrl(s.parse())
-    }
-}
-
-impl From<Url> for ParsedUrl {
-    fn from(url: Url) -> Self {
-        ParsedUrl(Ok(url))
-    }
-}
 
 /// Request instances are builders that creates a request.
 ///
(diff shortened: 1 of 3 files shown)


src/request.rs
```diff
@@ -65,25 +41,38 @@ impl fmt::Debug for Request {
...
+            let mut vec = vec![];
+
+            for (k, v) in url.query_pairs() {
+                vec.push((k.into(), v.into()));
+            }
+
+            vec
+        } else {
+            vec![]
+        };
```
jsha, line 69: I think this can be simplified like:

```
let query_pairs: Vec<(String, String)> = match url.parse::<Url>() {
  Ok(u) => u.query_pairs().collect(),
  Err(e) => vec![],
}
```

src/request.rs
```diff
@@ -377,23 +376,52 @@ impl Request {
...
 
-        ret
+    /// Get the url str that will be used for this request.
+    ///
+    /// The url might differ from that originally provided when constructing the
+    /// request if additional query parameters have been added using [`Request::query()`].
+    ///
+    /// In case the original url provided to build the request is not possible to
+    /// parse to a Url, this function returns the original, and it will error once the
+    /// Request object is used.
```
jsha, line 400:
```suggestion
    /// In case the original url provided to build the request is not possible to
    /// parse to a Url, this function returns the original, and the Request will
    /// produce an error when `call` or `send_*` is called.
```

jsha (commented):
I like this much better than where we landed with #350.

I still think query_pairs belongs on a separate `struct ParsedUrl` or similar. My reasoning is that we are asking `Request` to represent to many things: It's acting as a builder object _and_ an almost-finished request to be read from _and_ as a partial implementation of Url.

It's that "partial implementation of Url" that I'm worried about in particular. Right now we are just implementing the `.query_pairs()` accessor, but by the same logic we will soon want `.path()`, `.scheme()`, `.host()` etc. Since those are notionally part of a URL rather than a request, I think we will wind up with a nicer API if we represent them as a separate object.

Also, one of the obstacles we keep stubbing our toes on is trying to avoid a `Result` object when reading the query params. I think this is related to treating Request as both a builder object and an almost-finished request to be read from, combined with the fact that we take URLs as `&str` input. Those URLs have to be parsed at some point. In the builder pattern, we can defer the error involved in parsing until we actually build the object. But when we try to read fields before the object is built, we can't defer the error until build anymore. We have to either expose it with a `Result` return type, or as you do here, return the un-parseable original input.

algesten:
> Also, one of the obstacles we keep stubbing our toes on is trying to avoid a Result object when reading the query params. I think this is related to treating Request as both a builder object and an almost-finished request to be read from, combined with the fact that we take URLs as &str input.

Ok. I'll bring back the ParsedUrl wrapper.

src/request.rs
algesten: done

algesten:
@jsha Hopefully this is the final take 😬 🤞

jsha (approved):
Looks good to me! Thanks for your patience through all the revision requests.

algesten:
> Thanks for your patience through all the revision requests.

No thank you for having the patiences to keep feeding back!

I feel much happier with this end result than where we were in the middle of it.

Issue #267, opened 2020-12-03
Document and handle memory exhaustion issues


jsha:
It's possible for a malicious server to send very large responses and run a ureq user out of memory. For instance, the server could send a never-ending stream of headers, or a very large body.

For the response body, we should document that into_string, etc., are subject to maliciously large payloads, and if you are making requests to untrusted servers, you should use `into_reader()` and place an appropriate limit on how much you will read. Perhaps also for into_string, we should allocate the full capacity of the string up front, based on Content-Length? That would make it much faster for a malicious response to crash a ureq process, since the attacker wouldn't have to actually send all the bytes. But it might be better than slowly using up memory and crashing later? The ideal would be if it were possible to check if there's enough memory to allocate the corresponding string, and return an error instead of crashing. But I don't know of a way to do that in Rust.

For the response headers, since we can't delegate input limits to the user, I think we should add some reasonable limits. For instance, `hyper` has a max of 100 header fields, and `curl` has a max of 100k bytes for the whole header.

algesten:
I think we could put some reasonable default limits in. Like. No string bigger than 2GB. If the user needs to handle that, they can always go via Reader. Sadly I don't think std Rust has fine grained control over allocation for String, Vec etc. It simply panics if it can't allocate.

I think a kb limit for header fields makes more sense than a fixed number. Let's do that too!

Shnatsel:
`surf` sets the limit to 8kb and it's a bit too aggressive (241 websites out of the top million fail with surf, meanwhile curl and Firefox work fine), but there definitely needs to be _some_ limit. I've just DoS'd `ureq` with this one-liner in Linux shell:

`( echo -e "HTTP/1.1 200 OK\r"; echo -n "Huge-header: "; yes A | tr -d '\n' ) | nc -l localhost 8080`

jsha:
#351 addresses this for header fields, number of header fields, and size of body for `into_string()`. It documents the issue for `into_reader()`. I'm undecided yet what to do about `into_json()`. I suppose everything that parses JSON strings is subject to reading a huge string into memory? Probably we should set a size limit for the number of bytes read by `into_json()` like we will for `into_string()`. But now I'm wondering if serde_json itself should have a warning about unbounded input.

algesten:
For `into_json()`, the same limit as for `into_string()`, 10MB should be ok.

[...]

PR #351, opened 2021-03-24
Add size limits on responses

jsha:
This prevents a malicious server from running the client out of memory
by returning a very large body.

Limit into_string() to 1 megabyte.
Limit individual header fields to 100 kilobytes.
Limit number of header fields to 100.

Part of #267

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -21,6 +21,10 @@ use encoding_rs::Encoding;
 pub const DEFAULT_CONTENT_TYPE: &str = "text/plain";
 pub const DEFAULT_CHARACTER_SET: &str = "utf-8";
 const INTO_STRING_LIMIT: usize = 1_024 * 1_024;
+// Follow the example of curl and limit a single header to 100kB:
+// https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html
+const MAX_HEADER_SIZE: usize = 100 * 1_024;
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -21,6 +21,10 @@ use encoding_rs::Encoding;
 pub const DEFAULT_CONTENT_TYPE: &str = "text/plain";
 pub const DEFAULT_CHARACTER_SET: &str = "utf-8";
 const INTO_STRING_LIMIT: usize = 1_024 * 1_024;
```
algesten, line 23: 1MB sounds a bit too conservative. Could we maybe have 10 or 20?

My thinking is that ureq will is likely to be used mostly in situations that aren't super constrained. And if memory consumption is a concern, the user should consider using a streaming read anyway.

algesten (approved):
Looks good! One small thing.

src/response.rs
jsha: Yep, makes sense! I originally thought even as big as 1GB... But when writing this PR I thought "things that are UTF-8 are usually textual, and it's rare to have textual content that big." But it's easy to have a >1MB JSON file for instance.

algesten: Yes that's my thinking. I have that exact thing at work. We're running AWS transcription jobs on hour long interviews. The resulting JSON files we store on S3 are big!

algesten:
I think this is ready to merge. Looks good!

PR #338, opened 2021-02-27
Remove percent-encoding of cookies

lukasschlueter:
Fixes #315 

NOTE: This might be breaking for some people that might have relied on additional client-side percent-encoding, but I'd classify this as a bug-fix

PR #353, opened 2021-03-24
Stop percent encoding cookies and enforce RFC

### LINKED ISSUE #315: Cookies are getting URL encoded twice
Hi there,

I've found a bug in the cookie handling. If the server sets a cookie containing special chars that would require URL encoding, you end up with wrong cookie data.

In Firefox, navigating to https://httpbin.org/cookies/set?name=http%3A%2F%2Flocalhost.asdf.asdf.com returns the following (same behaviour also found in MS Edge or `isahc` for example):
```json
{
  "cookies": {
    "name": "http://localhost.asdf.asdf.com"
  }
}
```

ureq however returns:
```json
{
  "cookies": {
    "name": "http%3A%2F%2Flocalhost.asdf.asdf.com"
  }
}
```

I'm using ureq version `2.0.1` with `cookies` and `charset`(not sure that one matters) features

<details>
<summary>Sample test suite</summary>

```Rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_ureq() {
        let agent = AgentBuilder::new().build();

        agent
            .get("https://httpbin.org/cookies/set?name=http%3A%2F%2Flocalhost.asdf.asdf.com")
            // .query("name", "http://localhost.asdf.asdf.com")
            .call()
            .unwrap();
        let r = agent
            .get("https://httpbin.org/cookies/set?name2=https%253A%252F%252Flogin.local.host%252F")
            // .query("name2", "https%3A%2F%2Flogin.local.host%2F")
            .call()
            .unwrap();
        let response = agent
            .get("https://httpbin.org/cookies")
            .call()
            .unwrap()
            .into_string();

        assert_eq!(
            response.unwrap(),
            r#"{
  "cookies": {
    "name": "http://localhost.asdf.asdf.com", 
    "name2": "https%3A%2F%2Flogin.local.host%2F"
  }
}
"#
        )
    }

    #[test]
    fn test_isahc() {
        use isahc::http::{Request, Uri};
        use isahc::ReadResponseExt;

        let client = isahc::HttpClientBuilder::new().cookies().build().unwrap();

        client
            .get("https://httpbin.org/cookies/set?name=http%3A%2F%2Flocalhost.asdf.asdf.com")
            .unwrap();
        client
            .get("https://httpbin.org/cookies/set?name2=https%253A%252F%252Flogin.local.host%252F")
            .unwrap();
        let response = client.get("https://httpbin.org/cookies").unwrap().text();

        assert_eq!(
            response.unwrap(),
            r#"{
  "cookies": {
    "name": "http://localhost.asdf.asdf.com", 
    "name2": "https%3A%2F%2Flogin.local.host%2F"
  }
}
"#
        )
    }
}
```

</details>

algesten:
Close #315
Superseds #338 

Stop ureq from percent-encoding cookie name/values.

Enforce RFC 6265 and 2616 for cookie name/values.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -301,6 +301,15 @@ fn extract_cookies(agent: &Agent, url: &Url) -> Option<Header> {
         .cookie_tin
         .get_request_cookies(url)
         .iter()
+        // This guards against sending rfc non-compliant cookies, even if the user has
+        // "prepped" their local cookie store with such cookies.
+        .filter(|c| {
+            let is_ok = is_cookie_rfc_compliant(c);
+            if !is_ok {
+                debug!("do not send non compliant cookie: {:?}", c);
+            }
+            is_ok
+        })
         .map(|c| c.to_string())
         .collect::<Vec<_>>()
         .join(";");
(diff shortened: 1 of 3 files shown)


algesten:
@jsha I have a minor worry about this PR. It makes ureq RFC compliant, but probably at the expense of some interoperability. It's not clear to me why all those characters are disallowed from cookie names. Maybe some old school idea that these could be stored as individual files in a filesystem?

I guess I'm willing to "try it and see" if we get any user reports.

src/unit.rs
```diff
@@ -428,9 +445,94 @@ fn save_cookies(unit: &Unit, resp: &Response) {
...
+    // CHAR           = <any US-ASCII character (octets 0 - 127)>
+    // ...
+    //        CTL            = <any US-ASCII control character
+    //                         (octets 0 - 31) and DEL (127)>
+    // ...
+    //        token          = 1*<any CHAR except CTLs or separators>
+    //        separators     = "(" | ")" | "<" | ">" | "@"
+    //                       | "," | ";" | ":" | "\" | <">
+    //                       | "/" | "[" | "]" | "?" | "="
+    //                       | "{" | "}" | SP | HT
```
jsha, line 472: I think this is the same as token / tchar from RFC 7230: https://tools.ietf.org/html/rfc7230#page-26

```
   Most HTTP header field values are defined using common syntax
   components (token, quoted-string, and comment) separated by
   whitespace or specific delimiting characters.  Delimiters are chosen
   from the set of US-ASCII visual characters not allowed in a token
   (DQUOTE and "(),/:;<=>?@[\]{}").

     token          = 1*tchar

     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```

Given that, we already have is_tchar in header.rs: https://github.com/algesten/ureq/blob/9ec4e7192aa80d38d968e5989742ba10fbbbe575/src/header.rs#L192-L201. We should probably factor that out and use it both here and for headers.

jsha (approved):
Looks good! One suggested tweak.

src/unit.rs
algesten: done

PR #354, opened 2021-03-24
Fix clippy warnings and run as lint in CI

algesten:
Fixes all clippy warnings by either addressing the problem or disabling the lint. 

Also start running clippy as part of CI build.


algesten:
@jsha This does what #306 does and enables clippy as part of CI build.

jsha (approved):
Nice!

Issue #296, opened 2021-01-07
ureq::Error is massive


Nemo157:
```rust
[src/main.rs:17] std::mem::size_of::<ureq::Error>() = 576
```

It would be nice if this wasn't so large so that users don't need to `Box` it to avoid copying around a lot of unused memory in cases like `Result<(), ureq::Error>::Ok(())`.

Unfortunately, the size seems to mostly come from `Response`, and it's publicly visible that `ureq::Error::Http` contains a `Response`-by-value, so I don't know any way to non-breakingly fix this other than boxing `Response`'s data internally, which would affect performance on the success path too.

```
[src/main.rs:18] std::mem::size_of::<ureq::Response>() = 432
[src/main.rs:19] std::mem::size_of::<ureq::Transport>() = 568
```

jsha:
Good point, thanks for raising it. That's a bummer that we're somewhat painted into a corner by the design of `ureq::Error::Status` (BTW I'm assuming you're talking about the `Status` variant; there's no `Http` variant in 2.0.0). Of course, we could always take another major revision and go to 3.0.0. Doing so shortly after the release would minimize the pain of people upgrading twice.

I'd be interested in digging into `Response` to see if we can reduce its size in general.

Do you happen to have any benchmarks that could show us the performance penalty of the large Error object?

Nemo157:
https://github.com/serde-rs/json/issues/184 is the only benchmarking I've seen, supposedly a 5% performance improvement dropping `serde_json::Error` from 56 to 8 bytes.

jyn514:
Another alternative is to make `Response` smaller instead of boxing it. The only things I see that could be big are `Option<Url>`, `ResponseStatusIndex`, `Option<Unit>`, or `Stream`. It would be nice to narrow down which are making it so big.

algesten:
Some investigation. This is compiling on an M1 mac which I suspect is the reason my numbers are different.

Since `Transport` contains `Response`, the two enum error variants are quite similar.

```
ureq::Error 984
ureq::Transport 976
ureq::Response 840
```

In `Response` these are the top offenders.

```
Response::url 88
Response::stream 496
Response::unit 176
```

`Stream` is another enum (for TLS vs plain):

```
TcpStream 4
rustls::StreamOwned 456
```

So 46% of `Error` size is from rustls. This reminds me that I had seen this size difference before (clippy warning about big enum size difference), but for fear of performance, opted for keeping as is rather than tucking it away on the heap.

[...]

PR #356, opened 2021-03-24
Move Response::unit and stream to the heap

algesten:
This tries to go some way to fixing the problem described in #296 

On my machine, Error goes from 1000 to 344 (debug build), which maybe doesn't entirely solve it, but goes some way.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -225,13 +225,13 @@ impl PoolKey {
 /// *Internal API*
 pub(crate) struct PoolReturnRead<R: Read + Sized + Into<Stream>> {
     // unit that contains the agent where we want to return the reader.
-    unit: Option<Unit>,
+    unit: Option<Box<Unit>>,
     // wrapped reader around the same stream
     reader: Option<R>,
 }
 
 impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
-    pub fn new(unit: Option<Unit>, reader: R) -> Self {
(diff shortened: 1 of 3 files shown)


src/pool.rs
```diff
@@ -225,13 +225,13 @@ impl PoolKey {
 /// *Internal API*
 pub(crate) struct PoolReturnRead<R: Read + Sized + Into<Stream>> {
     // unit that contains the agent where we want to return the reader.
-    unit: Option<Unit>,
+    unit: Box<Option<Unit>>,
```
jsha, line 228: I think we want this the other way around:

```suggestion
    unit: Option<Box<Unit>>,
```

Because of the null pointer optimization. I wasn't able to find documentation dedicated to the NPO specifically but it's mentioned several places, including here: https://doc.rust-lang.org/nomicon/repr-rust.html

> The classic case of this is Rust's "null pointer optimization": an enum consisting of a single outer unit variant (e.g. None) and a (potentially nested) non- nullable pointer variant (e.g. Some(&T)) makes the tag unnecessary. A null pointer can safely be interpreted as the unit (None) variant. The net result is that, for example, size_of::<Option<&T>>() == size_of::<&T>().

jsha (changes requested):

src/pool.rs
algesten: done

jsha (approved):

Issue #339, opened 2021-02-27
Change per-request timeout to so it can only shorten Agent's timeout


jsha:
In https://github.com/algesten/ureq/pull/335 we introduced a per-request timeout that overrides the Agent's timeout. I'd like to propose a slight tweak to the semantics: The per-request config should only be able to make the timeout shorter, not longer. For instance, if I set `Agent.timeout` to 60s, and `Request.timeout` to 120s, the actual timeout would be 60s. But if I set `Request.timeout` to 30s, the actual timeout would be 30s.

My reasoning is that people will often want to construct a single Agent for the whole program, with default settings like max timeout. If would be confusing to find that some requests take longer than the globally configured Agent is supposed to allow.

Related: #335. /cc @DoumanAsh

DoumanAsh:
Hm, I do not believe this would be a right approach.
The problem is that current agent is not very convenient when you need to dynamically adjust timeout.
Not to mention it will be probably error prone for users themself, because it is highly likely that they are going to miss or forget about this complication.

Well, for my use case, as we cannot re-configure client conveniently, we're not going to set agent's timeout so I do not have strong feeling as we can upgrade to 2.x anyway.
But I feel this tweak is not really beneficial. 

> If would be confusing to find that some requests take longer than the globally configured Agent is supposed to allow.

I don't really think so.
The intention is to provide override after all.
It would be weird to be allowed override only up to agent's timeout.
In that case agent's config would be better named as `max_timeout`

jsha:
Can you tell me more about when you need to dynamically adjust timeout? It sounds as if, for your use case, it's better to not set any timeout on the Agent (effectively, an infinite timeout), and only set per-Request timeouts. Does that work for you?

DoumanAsh:
Well yeah, that's probably what I'm going to do.
We have timeout as part of configuration, and I don't want it to be left without ability to be changed at runtime

algesten:
@jsha I'm leaning towards letting the `Request` both extend or shorten the timeout. We can only speculate about how this will be used, is it more common to do as you say, set a max in the agent, or could it be as common to set a default in the agent that is sometimes extended?

Clamping a max on both agent and request level is easy for ureq's users to code, however if we disallow extending, the users are left stranded if they want to extend beyond the agent's max. Hence, I'm leaning towards not interfering, let the users do what they want.

jsha:
[...]

PR #357, opened 2021-03-25
Prepare release 2.1.0

algesten:


algesten:
@jsha #356 is referenced in the changelog since I believe we can merge it before releasing.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 2.0.3
+ * Bugfix: allow status lines without a reason phrase (#316)
+ * Example: "cureq" to easier make ad-hoc command line tests (#330)
```
jsha, line 3:
```suggestion
 * Example: "cureq" to make ad-hoc command line tests easier (#330)
```

jsha (commented):
This reminds me: what do you think of #339? Should the per-request method be able to set a longer timeout than the agent's timeout? We should answer this before cutting the new release, because it would be an API break to change it later.

jsha (approved):

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ureq"
-version = "2.0.2"
+version = "2.0.3"
```
jsha, line 3: Oh, one thing: this should actually take us to `2.1.0`. https://semver.org/

> Given a version number MAJOR.MINOR.PATCH, increment the:
>    MAJOR version when you make incompatible API changes,
>    MINOR version when you add functionality in a backwards compatible manner, and
>    PATCH version when you make backwards compatible bug fixes.

algesten: good catch!

CHANGELOG.md
jsha: I think this tweak got lost in the rebase; otherwise everything looks good.

PR #362, opened 2021-04-05
Flush sending headers and body always

### LINKED ISSUE #361: Sending prelude has to perfom flush?
This is a bit of bug/question report

It seems in our code we have situations where writing prelude gets buffered without actually sending data.
For some reason it leads to sudden connection abort

Here is line in question
https://github.com/algesten/ureq/blob/main/src/unit.rs#L410

Should we always flush it? My colleague told me that flush fixes the issue for us as server no longer aborts connection

P.s. problem seems to be with re-used connections though

DoumanAsh:
As per issue I added flush calls to ensure that data is sent over to server
Fixes #361

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -182,5 +182,5 @@ pub(crate) fn send_body(
         copy(&mut body.reader, stream)?;
     };
 
-    Ok(())
+    stream.flush()
 }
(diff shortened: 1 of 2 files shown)


jsha:
Fixed by #372. The reason `flush()` was also fixing the problem is that it calls `complete_io()`, which does a read and potentially returns an error (or early EOF). That error then causes ureq to drop the connection early enough that the request can still be retried, even if it's a POST.

PR #364, opened 2021-04-06
Fix typo, disable default features of cookie

AlyoshaVasilieva:
`warning: unused manifest key: dependencies.cookie.deafult-features`


algesten (approved):
Thanks

PR #369, opened 2021-04-13
Avoid extra cloning and parsing of url

Berrysoft:
If one requests with `Url`, the struct is passed by reference, but actually it is cloned, then stored as a `String`, and finally parsed again when requesting. This pull request aims to avoid extra cloning and parsing. If the url is passed by reference, it'll be only cloned once; if it is passed by value, there won't be any cloning or parsing.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -4,7 +4,7 @@ use url::Url;
 
 use crate::pool::ConnectionPool;
 use crate::proxy::Proxy;
-use crate::request::Request;
+use crate::request::{LazyUrl, Request};
 use crate::resolve::{ArcResolver, StdResolver};
 use std::time::Duration;
 
@@ -116,16 +116,6 @@ impl Agent {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn request(&self, method: &str, path: &str) -> Request {
-        Request::new(self.clone(), method.into(), path.into())
-    }
-
-    /// Make a request using an already-parsed [Url].
-    ///
-    /// This is useful if you've got a parsed Url from some other source, or if
-    /// you want to parse the URL and then modify it before making the request.
-    /// If you'd just like to pass a String or a `&str`, try [request][Agent::request].
-    ///
     /// ```
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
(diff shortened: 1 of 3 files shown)


algesten:
Hi @Berrysoft 

Thanks for looking and contributing to ureq. I'm afraid I'm not interested in the direction this PR is going, and I'll explain why. ureq has two primary design goals. It's synchronous and it got a ergonomic/"user friendly", obvious API. The first point is obvious, but I'll elaborate on the second.

I want ureq to be extremely approachable to people who are new to Rust. On a philosophical level, it saddens me that the current trend in the community is async. Rust is such a nice language and I want more people to discover it, however for a beginner async is not easy. It doesn't take much async code before one needs to grapple with conflicting runtimes, `Unpin` and cancellation trickiness (don't get me wrong, I love Rust async story, but I don't think every API needs to be async).

Specifically I prefer ergonomic/simple over efficiency and correctness. It doesn't mean I don't care about performance or correctness, but faced with a situation where we must chose, I'll go for ergonomic.

Given the above, I want the ureq API to have as little "noise" and scary type bounds as possible. For example, I want `&str` to be the "native" input/output type for stuff like URLs, query parameters, cookies, header values, etc. There are certainly other types that would be possible, and might have provided some convenience and/or efficiency in certain situations (`Into<String>`, `FromStr<Url>` etc), but the ideal for me is `ureq::get(url: &str)`.

So about your PR. Given above motivation, I don't want to alter the API – I want to stick with `&str`. The problem you're trying to address (of not unnecessarily parsing the Url), is one that very recently was decided to be the least bad of weighing a number of options together.

* Request follows a builder pattern. A url parsing error should surface as a `Result` first on `.call()` (or `.send()` etc). Not before that.
* The "native" input to ureq is `&str and we want to expose query parameters, host, path etc all as `&str`.
* Specifically we do not want to have `Cow` in our surface API, since this is a rather awkward type for a beginner.
* `Url` is a great crate, but it prioritizes correctness and efficiency, which means exposing the user to some more advanced types and Rust concepts.
* We're ok with having `Url` in some parts of the ureq API, but it is secondary to `&str`.
* It's nice to rely on `Url` internally for correctness of stuff like query parameter encoding.

The above was explored at lengths in these two PRs (where you can read more about the motivations). You will also find that we actually had a more lazy approach to the Url parsing earlier, but deliberately moved away from it due to above reasoning. It's a point in case where we had to chose ergonomic/simple over efficiency.

https://github.com/algesten/ureq/pull/310
https://github.com/algesten/ureq/pull/350

Which doesn't mean I'm particularly happy with the current situation, but I think it's acceptable. For an advanced user where efficiency is paramount, they can always manipulate a `Url` beforehand and use that through `agent::request_url()`.

I'm also happy to consider PRs that improves on the current situation, but not that alters the surface API. One way forward would be to lobby the `Url` crate maintainers to have `&str` instead of `Cow<str>` for query parameters. Another would be to try and isolate the `Url` crate's handling of query parameters (correct encoding etc), so that we can reuse that internally in ureq.

Berrysoft:
OK, @algesten . I respect your opinion. But at least the extra clone could be avoided without changing public APIs.

https://github.com/algesten/ureq/blob/50d9ccff8ccc167a49faba7ae1585db4bcdf114d/src/agent.rs#L143-L145
https://github.com/algesten/ureq/blob/50d9ccff8ccc167a49faba7ae1585db4bcdf114d/src/request.rs#L53-L55

Just a `to_string` call is OK. Do you need a pull request for that?

algesten:
Actually a PR would be good, cause I don't quite get what we're talking about :)

PR #370, opened 2021-04-17
Rename types to follow Rust conventions

algesten:
This is a breaking API change.

Uncertain if we want to land it, but  wan to open for the discussion. I noticed the change in a crate we depend on, and was thinking maybe we should follow along? https://github.com/briansmith/webpki/commit/5b67fb92dacf216d11e26f99f4b972e24deabd4c

The breaking change is that `ErrorKind::HTTP` is now `ErrorKind::Http`

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -38,7 +38,7 @@ pub(crate) struct AgentConfig {
     pub redirects: u32,
     pub user_agent: String,
     #[cfg(feature = "tls")]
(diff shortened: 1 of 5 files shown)


jsha:
I think we should hang onto this change until the next time we need to do an API break for functional reasons. Renaming isn't enough reason for an API break by itself, IMO.

algesten:
Fair enough! Thanks!

Issue #124, opened 2020-07-31
Late TLS errors break agent pooling


algesten:
When using an agent pooled connection, we try to validate the connection before use, but some TLS errors might surface as late as when we start reading the response. 

https://github.com/algesten/ureq/blob/master/src/unit.rs#L172

```rust
body::send_body(body, unit.is_chunked, &mut stream)?;

// start reading the response to process cookies and redirects.
let mut stream = stream::DeadlineStream::new(stream, unit.deadline);let mut resp = Response::from_read(&mut stream);
```

In practice the error could probably surface already on `send_body`, however it's hard to detect the error once the body is sent.

The solution might be to buffer something like 2kb of the body, send it, try to read response headers but look out for `WouldBlock` – a TLS error would probably surface here.

The mechanic of reading response headers before sending entire body is necessary anyway to properly handle 307 `Expect: 100-Continue`, so this bug and that feature should be thought of in tandem.

See background here #122

jsha:
> The solution might be to buffer something like 2kb of the body, send it, try to read response headers but look out for WouldBlock – a TLS error would probably surface here.

How about another approach - we can make `unit.rs` aware of Payload. So e.g. unit::connect would take a Payload argument instead of a Reader. Most Payloads (Empty, Text, JSON, Bytes) are replayable; only Reader is not. So we could have more generic logic that says "if we get a connection-related error during a request-response cycle, and HTTP semantics say the request is retryable, we will retry it." With the exception that Payload::Reader is never retryable once it's started being read. Users who want retryability would have to use one of the replayable body types.

algesten:
@jsha 

> Most Payloads (Empty, Text, JSON, Bytes) are replayable; only Reader is not.

Yeah, that's way simpler. I'm just thinking of my own use case for work with ureq. I stream large bodies (video files). It's unfortunate if it behaves differently depending on type of body.

PR #372, opened 2021-04-17
Don't reuse conns with bytes pending from server

jsha:
Related to #361 and #124.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -147,7 +147,14 @@ impl Stream {
 
     // Check if the server has closed a stream by performing a one-byte
     // non-blocking read. If this returns EOF, the server has closed the
-    // connection: return true. If this returns WouldBlock (aka EAGAIN),
+    // connection: return true. If this returns a successful read, there are
+    // some bytes on the connection even though there was no inflight request.
+    // For plain HTTP streams, that might mean an HTTP 408 was pushed; it
+    // could also mean a buggy server. For HTTPS streams, that might mean a
+    // close_notify alert, which is the proper way to shut down an idle stream.
+    // Either way, bytes available on the stream before we've made a request
+    // means the stream is not usable, so we should discard it.
+    // If this returns WouldBlock (aka EAGAIN),
     // that means the connection is still open: return false. Otherwise
     // return an error.
     fn serverclosed_stream(stream: &std::net::TcpStream) -> io::Result<bool> {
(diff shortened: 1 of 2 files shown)


algesten (approved):
This looks exactly right.

algesten (approved):
Let's land a publish a new version. Thanks!

algesten:
@jsha I was thinking a bit more about this, and wonder if we have covered all the cases. While this check will cover a lot of cases, I wonder if there isn't a race condition here also. Consider:

1. ureq does this check, all looks good, connection still open.
2. remote closes the connection.
3. ureq sends next request.

Given distributed systems, we can't really know whether a connection can be reused until we sent the next request successfully.

jsha:
There's definitely a race condition, and the [HTTP RFC talks about it](https://tools.ietf.org/html/rfc7230#section-6.3.1):

>   Connections can be closed at any time, with or without intention.
>   Implementations ought to anticipate the need to recover from
>   asynchronous close events.

>    When an inbound connection is closed prematurely, a client MAY open a
   new connection and automatically retransmit an aborted sequence of
   requests if all of those requests have idempotent methods (Section
   4.2.2 of [RFC7231]).  A proxy MUST NOT automatically retry
   non-idempotent requests.

>   A user agent MUST NOT automatically retry a request with a non-
   idempotent method unless it has some means to know that the request
   semantics are actually idempotent, regardless of the method, or some
   means to detect that the original request was never applied.
   
So: connections can be closed at any time, and the remedy is retrying. But retrying doesn't work if you've sent a POST request, because they are not idempotent. So it's best to try and reduce the likelihood of sending on a closed connection as much as possible.

One thing that's not clear to me is whether it's permissible to retry an incompletely sent POST. For instance, if you sent only the headers, or sent (Content-Length - 1) bytes, then discovered the connection was closed, could you assume the POST was not applied and retry automatically?

algesten:
@jsha 

> One thing that's not clear to me is whether it's permissible to retry an incompletely sent POST.

I think this line in the spec:

>> or some means to detect that the original request was never applied.

means it's not permissible. We can't know the reason for the connection closing. Could be the server, or some middle box. The server might have applied  the non-idempotent request.

DoumanAsh:
The only extra you can do is to flush after sending headers, making sure server receives it triggering whatever other reason which would lead server to close connection I think (which also helped us to mitigate this situation)

PR #374, opened 2021-04-20
1.x: Don't reuse conns with bytes pending from server (#372)

### LINKED ISSUE #124: Late TLS errors break agent pooling

### LINKED ISSUE #361: Sending prelude has to perfom flush?

DoumanAsh:
This makes us less likely to try and reuse a closed connection, which
produces problems in particular for requests that can't be retried.

Fixes #361
Fixes #124

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -68,6 +68,7 @@ impl AgentState {
     fn new() -> Self {
         Self::default()
     }
+
     pub fn pool(&mut self) -> &mut ConnectionPool {
         &mut self.pool
     }
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -116,7 +116,15 @@ impl fmt::Debug for Stream {
 impl Stream {
     // Check if the server has closed a stream by performing a one-byte
     // non-blocking read. If this returns EOF, the server has closed the
-    // connection: return true. If this returns WouldBlock (aka EAGAIN),
+    // connection: return true. If this returns a successful read, there are
+    // some bytes on the connection even though there was no inflight request.
+    // For plain HTTP streams, that might mean an HTTP 408 was pushed; it
+    // could also mean a buggy server that sent more bytes than a response's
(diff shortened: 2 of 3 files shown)


DoumanAsh:
Will adapt tests as separate commit

DoumanAsh:
@jsha ping?

jsha (approved):

DoumanAsh:
Thank you, do you want to take some time to verify this fix before releasing new version?
We would like to use it in our code and enable back connection pooling.

jsha:
I have enough confidence in the tests to release now. I've sent out a pair of PRs for CHANGELOGs, and will do a new release on the 1.x and 2.x series soon.

PR #377, opened 2021-04-26
Update CHANGELOG and Cargo.toml for 1.5.5

jsha:


algesten (approved):
we're still doing 1.x?

jsha:
I figured this is a serious enough bug to backport, and @DoumanAsh was kind enough to help us find it, and is still on 1.x.

DoumanAsh:
Yes, we currently using 1.5.x in our product and we are being careful with upgrading our dependencies, so I would appreciate if we could still use 1.5.

PR #380, opened 2021-04-27
[1.x] Clone state when building agent

LeoComandini:
Before this change, Agent.build() always set the default
AgentState.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -65,10 +65,6 @@ pub(crate) struct AgentState {
 }
 
 impl AgentState {
-    fn new() -> Self {
-        Self::default()
-    }
-
     pub fn pool(&mut self) -> &mut ConnectionPool {
         &mut self.pool
     }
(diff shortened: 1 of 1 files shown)


jsha:
Hi @LeoComandini! We're only backporting major fixes to 1.x. Is this a problem on 2.1.1? Can you add more detail about the problem you're trying to solve?

LeoComandini:
This is not a problem for 2.1.1, as it use the new agent builder.

Following the docs https://docs.rs/ureq/1.5.5/ureq/struct.Agent.html#method.set_proxy, the proxy is set with set_proxy but unset by calling build.
Perhaps you prefer just to change the docs to:
```
let proxy = ureq::Proxy::new("user:password@cool.proxy:9090").unwrap();
let mut agent = ureq::agent()
agent.set_proxy(proxy)
```

jsha:
I think we're not likely to backport this fix to 1.x, sorry.

PR #389, opened 2021-05-22
Restore support for native-tls

### LINKED ISSUE #319: Restore support for native-tls
I see that `ureq` used to support `native-tls`, but that was dropped in version 2.0.

This is a bit of a problem for [my use case](https://github.com/rust-secure-code/cargo-supply-chain) which prioritizes portability over confidentiality of the data.

As much as I love `rustls`, it relies on `ring`, which has the drawback of only working on x86 and ARM. This excludes e.g. POWER9, which is currently the only option for a high-end desktop or server with fully open-source firmware.

On the other hand, I'm querying publicly available information over HTTPS, so confidentiality of the data is not a high priority.

I understand that native-tls support was removed due to the added maintenance burden; however, I see that most of the other HTTP clients support both (e.g. `attohttpc`, `minreq`, `reqwest`, `surf/http-client`, `mio_httpc`, and others). Perhaps they have discovered a good way to abstract over both libraries?

Lack of native-tls support is particularly unfortunate since `ureq` is the most feature-rich blocking client, and with this, #29 and #267 squared away, the ecosystem could simply converge on it as opposed to the current fragmented state.

jsha:
This allows any of: rustls, native-tls, both, or neither. It replaces the `stream::Inner` enum with a trait, which will hopefully be more pleasant to work with.

Fixes #319


src/stream.rs
```diff
@@ -22,15 +18,100 @@ use crate::error::ErrorKind;
...
-    #[cfg(feature = "tls")]
-    Https(rustls::StreamOwned<rustls::ClientSession, TcpStream>),
-    Test(Box<dyn Read + Send + Sync>, Vec<u8>),
+trait Inner: Read + Write {
+    fn is_poolable(&self) -> bool;
+    fn socket(&self) -> Option<&TcpStream>;
+    fn as_write_vec(&self) -> &[u8] {
+        panic!("as_write_vec on non Test stream");
+    }
+}
```
algesten, line 30: Nice!

algesten (approved):
This looks great to me. The Inner solution was what I was stuck on in my attempt.

👍 

Thanks for doing this!

dralley:
Is anything blocking this from merging?  @jsha @algesten

jsha:
Mainly just indecision about whether to merge this vs the more general
version at https://github.com/algesten/ureq/pull/391. Also looks like I
need to fix up some tests.

src/stream.rs
```diff
@@ -341,37 +374,67 @@ fn configure_certs(config: &mut rustls::ClientConfig) {
         .add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
 }
 
-#[cfg(feature = "tls")]
+#[cfg(all(not(feature = "tls"), not(feature = "native-tls")))]
+pub(crate) fn connect_https(unit: &Unit, _hostname: &str) -> Result<Stream, Error> {
+    Err(ErrorKind::UnknownScheme
+        .msg("URL has 'https:' scheme but ureq was built without HTTP support")
```
dontlaugh, line 380:
```suggestion
        .msg("URL has 'https:' scheme but ureq was built without TLS support")
```

dontlaugh (commented):
I do see how the feature gating in `connect_https` is a bit noisy, especially if you were to add another implementation. I have also read through #391 and it seems like a nice solution.

But I still lean towards this approach that doesn't introduce more types (yet).

There is nothing in this PR that precludes us from adding something like #391 later, right?

algesten:
Thanks @dontlaugh, that is good to hear.

Jacob is driving these changes, and he's currently on vacation, however before he left, the last thing we discussed was that he's leaning towards this solution too.

You weighing in gives us more confidence it is the right way forward. Thanks!

jsha:
Thinking some more about this in the context of feature unification:

https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

> A consequence of this is that features should be *additive*. That is,
enabling a feature should not disable functionality, and it should usually
be safe to enable any combination of features.

This PR fails the test above: the native-tls feature isn't additive. It
disables use of rustls. So if you have a big dependency tree, one
dependency might be expecting to use rustls and another might be expecting
to use native-tls. Due to feature unification, native-tls would be in force
for both dependencies.

I think the ideal would be for the top-level crate to be able to pick a TLS
backend that would be in force for all its dependencies, so we would inly
need to build one of rustls or native-tls for any given binary. I don't
think it's possible to achieve that with the cargo feature system.

A second best way to handle the situation of two deps wanting rustls and
native-tls would be: they both get compiled, but only one gets used at
runtime. We could do this the way the `log` crate does, with global state.
For instance, we would have a global default TlsConnector, which the
application could modify under a lock to use whichever backend it prefers.

There's also the question of what happens when you're building on a
platform that is unsupported by rustls, and one of your deps uses ureq with
--no-default-features --features native-tls feature, and another dep uses
the default features. I think we should use a platform config gate for the
unsupported platforms, and default to native-tls in that case.

So, current thinking: implement the platform gate first, since it doesn't
commit us to a new customization API, and then figure out jf we want
further customization and how it should work with diamond dependencies.

algesten:
I will write more tomorrow, but I think there's a problem with intention here.

_Why_ would one dep select native-tls instead of relying on defaults?

The only reason I believe would be that it consumes an API on TLS 1.0 or some bad cipher. Ie the function of the lib requires that backwards compat.

Making a global singleton wouldn't solve that, since we effectively could get libraries "fighting" over setting the singleton.

Maybe it would be better if ureq preferred native-tls if it is enabled, but there's an override switch on the agent builder?

dontlaugh:
It's not just old ciphers. WebPKI, a hard dep of rustls, doesn't support IP SANs. https://github.com/briansmith/webpki/issues/54

Honestly avoiding anything ring/ruslts/webpki makes everyone's life easier.

algesten:
> Honestly avoiding anything ring/ruslts/webpki makes everyone's life easier.

I don't agree, but that's beside the point. I understand there are cases where people can't use rustls, and do want to support that. My comment is more about how we achieve that if we get diamond dependencies on ureq with different sets of feature flags.

PR #391, opened 2021-05-28
Use native-tls on platforms that need it, and remove "tls" feature

jsha:
Per #319, we want to add back native-tls support. However, as discussed in https://github.com/algesten/ureq/issues/319#issuecomment-930795723, using Cargo features to select a TLS backend (as we did in 1.x) is a bad fit, due to diamond dependencies.

Instead, ureq takes an opinionated stance and allows runtime configuration to change it. On all platforms that support rustls, we use rustls. On other platforms, we use native-tls. In either case, you can override the default: supply your own TLS implementation by calling `AgentBuilder::tls_connector`. Since this is configured per-Agent, there's no risk that crate A can accidentally change the backend used by crate B.

We now have a couple of new traits, `TlsConnector` and `HttpsStream`, that abstract over TLS configuration and encrypted streams. There are impls of `TlsConnector` for both `rustls::ClientConfig` and `native_tls::TlsConnector`. To enable the native_tls impl on platforms that default to rustls, build with the native-tls-adapter feature (and call `AgentBuilder::tls_connector`).

To mange the complexity introduced by having platform-based configs, this removes the "tls" feature. TLS is now always enabled, on all platforms. For applications that want to omit TLS support for binary-size reasons, we may in the future offer a linker-friendly AgentBuilder method that disables TLS. That should allow the linker to remove the unused TLS-related symbols and build time.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
-    env: 
+    env:
       RUSTFLAGS: -D warnings
     steps:
     - uses: actions/checkout@v2
@@ -39,7 +39,7 @@ jobs:
         with:
           command: doc
           # Keep in sync with Cargo.toml's [package.metadata.docs.rs]
-          args: --no-default-features --no-deps --features "tls json charset cookies socks-proxy"
+          args: --no-default-features --no-deps --features "tls native-tls json charset cookies socks-proxy"
   build_and_test:
     name: Test
     runs-on: ubuntu-latest
@@ -47,13 +47,15 @@ jobs:
       matrix:
         tls:
           - ""
-          - tls
+          - "tls"
+          - "native-tls"
         feature:
           - ""
           - json
           - charset
           - cookies
           - socks-proxy
+          - native-tls-adapter
     env:
       RUST_BACKTRACE: "1"
       RUSTFLAGS: "-D dead_code -D unused-variables -D unused"
@@ -68,4 +70,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --no-default-features --features "${{ matrix.tls }} ${{ matrix.feature }}"
+          args: --no-default-features --features "${{ matrix.tls }}" "${{ matrix.feature }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,6 @@ jobs:
           - charset
           - cookies
           - socks-proxy
-          - native-tls-adapter
     env:
       RUST_BACKTRACE: "1"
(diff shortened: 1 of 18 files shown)


dontlaugh (commented):
Two teaspoons of bikeshed paint.

jsha:
Alright, this mostly reflects my latest thinking in https://github.com/algesten/ureq/issues/319:

 - Rustls stays the default.
 - Dependencies can't change that default.
 - If you want to use native-tls (or another backend, like mbedtls), you need to configure that on an Agent.
 - If a library crate wants to let you configure alternate backends, it needs to let you pass in an Agent (we should document that).
 - You can remove the rustls dependency entirely (in favor of native-tls or some other backend) with `--no-default-features`.

jsha:
This is now complete and ready for a review. I think the result is pretty good! I've rewritten the PR description to reflect the latest state.

jsha:
>  Compiling without any TLS support is in similar territory – an odd thing to want, but is it right that we say it shouldn't be possible?
> Saving compile time is another reason to keep deps slim. If you absolutely don't need TLS, it might be a bit irritating to see those crates compile…

I agree it's good to keep compile times fast, but maintaining the `tls` feature has a significant maintenance burden, especially now that it will be interacting with per-platform conditional compilation and the native-tls-adapter feature (which I'm happy to rename native-tls). When we dropping native-tls support in the 1.x -> 2.x migration, it was mainly because of maintenance burden - but now I'm thinking that a lot of that burden was due to the `tls` feature itself.

As a way to quantify the maintenance burden, check out the diffs to test.sh and test.yml. Previously we had to test [tls, no-tls] x [every feature]. Now we just have to test [every feature]. Of course the tests are automated, but each combination in that matrix is a potential breakage when modifying code that the authors (us) have to fix.

And that maintenance burden is laid out against a fairly small benefit: keeping compile times fast for the small number of users that might explicitly disable TLS. I would be interested in hearing from such users if this turns out to cause significant burden for them, though.

algesten:
Hi @ThomasLamprecht thanks for weighing in!

This is exactly the kind of viewpoint we want to hear. I say the decision here is still up in the air.

algesten:
@jsha I pushed a commit reintroducing `tls` and `native-tls` as independent options.

I think the rustls arch-selection is elegant, but I think Cargo is lacking what we need to take it all the way (https://github.com/rust-lang/cargo/issues/1197).

I wanted to see this maintenance burden for myself, so I decided to give it a go and I think you have already solved 99% of the problems we used to have in ureq 1.0 by your refactoring. Specifically:

1. The recognition of `https` is not feature specific. I.e. internally we always have a concept of "https stream" regardless of feature set.
2. The `crate::default_tls_config()` you made on the top, isolates the default impl very neatly.

jsha:
I pushed some small changes with my review fixes. I think this is pretty much ready to go - what do you think @algesten ?

algesten:
This continues in #449

PR #394, opened 2021-06-18
Clear Content-Length header from redirected requests

guspower:
This PR removes the `Content-Length` header from subsequent redirect requests if set. A test verifies the new behaviour.

This issue causes `hyper` to attempt to read a non-existent body from the redirect GET request, as seen when trying to use `wiremock-rs` to simulate redirect-on-post behaviour where the POST contains a payload. The issue has be replicated in [an example project](https://github.com/guspower/redirect-on-post).

PR #395, opened 2021-06-18
 Clear Content-Length header from redirected requests

jsha:
This PR removes the Content-Length header from subsequent redirect requests if set. A test verifies the new behaviour.

This is a copy of @guspower's #394, tweaked to pass our lint checks.

### DIFF
--- a/src/test/redirect.rs
+++ b/src/test/redirect.rs
@@ -157,7 +157,9 @@ fn redirect_post_with_data() {
         assert_eq!(unit.method, "GET");
         test::make_response(200, "OK", vec![], vec![])
     });
-    let resp = post("test://host/redirect_post1").send_string("data").unwrap();
+    let resp = post("test://host/redirect_post1")
+        .send_string("data")
+        .unwrap();
     assert_eq!(resp.status(), 200);
 }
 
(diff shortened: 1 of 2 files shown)


algesten (approved):
This looks good. 

I'm wondering wheter there are more headers to clear out. 

Off the top of my head `ETag` and `Content-Type` could be interesting. Or maybe even more Contebt-xxx headers? 🤔 

Anything relating to the request body really.

guspower:
Yes good point. Most probably the following would need to be scrubbed:

```
Content-Encoding
Content-Length
Content-MD5
Content-Type
If-Match
If-Modified-Since
If-None-Match
If-Range
If-Unmodified-Since
Range
Transfer-Encoding 
```

`Cache-Control` and `Pragma` might also need to be removed as they would probably be irrelevant or incorrect for the subsequent redirect request. 

@algesten Am happy to pull together a change to reflect this behaviour if that helps.

jsha:
An issue or another PR would be great. I'd be curious to see what the behavior of some other user-agents is with respect to those headers. I think `Range` in particular should stay, since it's talking about a range of bytes in the requested object. Perhaps you were thinking of Content-Range, which is about the range of bytes in a POSTed / PUT object?

jsha:
Oh and by the way closing this out; we wound up merging #394.

PR #405, opened 2021-08-11
Fix clippy warnings

nmurali94:
1. Fix linter warning from clippy about unnecessary borrows - "This expression borrows a reference ... that is immediately dereferenced by the compiler"
2. Replace explicit loop with iterator operations in Header to avoid unnecessary mutation

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -134,7 +134,7 @@ fn copy_chunked<R: Read, W: Write>(reader: &mut R, writer: &mut W) -> io::Result
         let header = header_str.as_bytes();
         assert!(header.len() <= CHUNK_HEADER_MAX_SIZE);
         let start_index = CHUNK_HEADER_MAX_SIZE - header.len();
-        (&mut chunk[start_index..]).write_all(&header).unwrap();
+        (&mut chunk[start_index..]).write_all(header).unwrap();
 
         // And add the footer
         chunk.extend_from_slice(b"\r\n");
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -113,7 +113,7 @@ impl ConnectionPool {
 
                 // Remove the newest matching PoolKey from self.lru. That
                 // corresponds to the stream we just removed from `recycle`.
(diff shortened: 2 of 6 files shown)


algesten:
Hi! Welcome to ureq!

Thank you for the PR, but I think we need to split the work up. Fixing the clippy lint warnings is good, maybe we can reshape this PR to do only that?

Replacing the for-loops with iterators, I'm not so sure about. Typically I find for loops more readable than long iterator chains, and in this specific instance, I think you will end up with worse performance.

The for loop both checks for `:` and illegal chars at the same time.

```rust
        let bytes = self.as_bytes();
        let index = bytes.iter().position(|x| *x == b':').unwrap_or(bytes.len());

        let name_bytes = &bytes[..index];
        let bad = name_bytes.partition_point(is_tchar);
```

This however uses two iterators (`position()` and `partition_point()`) to do the same thing. Reading the docs, I also don't think the use of `partition_point` is correct here.

So to sum it up:

* If you change this PR to addressing the clippy lints only, I'm happy to land that.
* If you want to explore iterators further, please open a separate PR, but beware that I have a mild preference against them, and that I don't want to accept anything that makes performance worse.

nmurali94:
> Fixing the clippy lint warnings is good, maybe we can reshape this PR to do only that?

Will do!

> This however uses two iterators (position() and partition_point()) to do the same thing

My view here was that while we have two iterations here, we're doing half the work in each step so we shouldn't be doing much more work. Plus without a branch check in each step, my intuition says the two iterators can be better optimized by the compiler. But since I don't have anything concrete to show here, I'll defer to your judgement

algesten:
> My view here was that while we have two iterations here, we're doing half the work in each step so we shouldn't be doing much more work. Plus without a branch check in each step, my intuition says the two iterators can be better optimized by the compiler. But since I don't have anything concrete to show here, I'll defer to your judgement

You could be right. I can be convinced by science here ;)

algesten (approved):
LGTM!

PR #409, opened 2021-08-13
Use debug logs when making requests

pingiun:
I think logging of opening a connection constitutes as debug information

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -1,7 +1,7 @@
 use std::io::{self, Write};
 use std::time;
 
-use log::{debug, info};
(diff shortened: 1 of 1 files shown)


algesten:
Agree!

Would you mind inspecting the other on `info` level to ensure we don't have any more that should be downgraded?

pingiun:
> Would you mind inspecting the other on `info` level to ensure we don't have any more that should be downgraded?

These are the only ones!

algesten (commented):
Let's merge

algesten (approved):

PR #410, opened 2021-08-16
Add socks4 support 

Airkek:
image

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -4,6 +4,8 @@ use crate::error::{Error, ErrorKind};
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum Proto {
     HTTPConnect,
+    SOCKS4,
+    SOCKS4A,
     SOCKS5,
 }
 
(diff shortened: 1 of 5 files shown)


src/proxy.rs
```diff
@@ -192,8 +198,8 @@ mod tests {
 
     #[cfg(feature = "socks-proxy")]
     #[test]
-    fn parse_proxy_socks_user_pass_server_port() {
-        let proxy = Proxy::new("socks://user:p@ssw0rd@localhost:9999").unwrap();
+    fn parse_proxy_socks4_user_pass_server_port() {
+        let proxy = Proxy::new("socks4://user:p@ssw0rd@localhost:9999").unwrap();
```
algesten, line 202: I think we need to duplicate this test to have a `socks://` ensure SOCKS5 and a `socks4://` doing SOCKS4.

To maintain backwards compatibility and not have to major version, we need to make it so `socks://` means SOCKS5. This should also be highlighted in the docs.

algesten (changes requested):
Hi! Welcome to ureq. This overall looks good. Thanks!

I'm not very good with proxies, cause I never use them. I think there's one little thing we need to address with the PR. See comment.

src/proxy.rs
Airkek: > To maintain backwards compatibility and not have to major version, we need to make it so socks:// means SOCKS5.

`socks` still means SOCKS5

algesten: Cool! thanks!

algesten (approved):

algesten:
Perfect! Let's land this. Thanks!

PR #415, opened 2021-09-01
Redact sensitive headers when logging prelude

### LINKED ISSUE #414: Avoid logging sensitive headers
Currently the entire HTTP prelude is printed at debug level, including all headers

https://github.com/algesten/ureq/blob/a1d7009e24185d05e3e411acdcf2b55103858b5e/src/unit.rs#L410

It would be nice to avoid printing sensitive headers like `Authorization`

andrewhickman:
Closes #414

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -1,3 +1,4 @@
+use std::fmt::Display;
 use std::io::{self, Write};
 use std::time;
 
@@ -355,15 +356,12 @@ fn connect_socket(unit: &Unit, hostname: &str, use_pooled: bool) -> Result<(Stre
 #[allow(clippy::write_with_newline)]
 fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> io::Result<()> {
     // build into a buffer and send in one go.
-    let mut prelude: Vec<u8> = vec![];
+    let mut prelude = PreludeBuilder::new();
 
     // request line
-    write!(
-        prelude,
-        "{} {}{}{} HTTP/1.1\r\n",
-        unit.method,
+    prelude.write_request_line(
+        &unit.method,
         unit.url.path(),
-        if unit.url.query().is_some() { "?" } else { "" },
         unit.url.query().unwrap_or_default(),
     )?;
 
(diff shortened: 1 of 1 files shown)


src/unit.rs
```diff
@@ -378,42 +377,117 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> io::Result<()>
...
     // other headers
     for header in &unit.headers {
         if !redir || !header.is_name("Authorization") {
             if let Some(v) = header.value() {
-                write!(prelude, "{}: {}\r\n", header.name(), v)?;
+                if is_header_sensitive(header) {
+                    prelude.write_sensitive_header(header.name(), v)?;
+                } else {
+                    prelude.write_header(header.name(), v)?;
+                }
```
algesten, line 405: Wondering about the design choice here of having the `is_header_sensitive` external to `PreludeBuilder` at the same time as as the builder has `write_sensitive_header`. Not saying it's wrong, just wondering if there was a thought there?

src/unit.rs
```diff
@@ -378,42 +377,117 @@ fn send_prelude(unit: &Unit, stream: &mut Stream, redir: bool) -> io::Result<()>
...
             }
         }
     }
 
     // finish
-    write!(prelude, "\r\n")?;
+    prelude.finish()?;
 
-    debug!("writing prelude: {}", String::from_utf8_lossy(&prelude));
+    debug!("writing prelude: {:?}", prelude);
```
algesten, line 413: Hm hm… I think we maybe should use the `Display` trait here and let the `Debug` trait just dump the internal state of the `PreludeBuilder`.

algesten (changes requested):
Hi and thanks for the PR!

Looks good! I had a couple of small questions/things.

src/unit.rs
andrewhickman: Changed to use `Display` instead

src/unit.rs
andrewhickman: I think the reason I didn't include that logic inside PreludeBuilder is I wanted to re-use the `Header::is_name` function, but not all the headers we add come from an actual `Header` struct

jsha (approved):
This generally looks good to me. @algesten do you want to take a second look?

algesten (approved):
I think this is good to go

PR #35, opened 2020-01-25
Add flate2 and automatic decompression

adrianN:
I want to work on issue #29, mostly to learn some Rust. I added literally ten lines that, to me, look like they should add support for automatic decompression. I open this PR before writing any tests to get a discussion started whether I'm even working on the right file. I only have little time per week on this, so if you want to get it done quickly, somebody else should work on this.

PR #421, opened 2021-09-26
Add support for gzip and brotli

AlyoshaVasilieva:
I wanted compression, found #35, and used it as starting point.

Automatically sends the Accept-Encoding header on requests when appropriate, so just enabling the features gets compression working.

I didn't add support for `deflate` as its implementation on some servers used to be buggy and it seems like most everyone is using gzip/br now anyway.

I tested this on ~1500 websites and it appears to be working.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -18,6 +18,12 @@ use serde::de::DeserializeOwned;
 #[cfg(feature = "charset")]
 use encoding_rs::Encoding;
 
+#[cfg(feature = "gzip")]
+use flate2::read::GzDecoder;
+
+#[cfg(feature = "brotli")]
+use brotli_decompressor::Decompressor;
+
 pub const DEFAULT_CONTENT_TYPE: &str = "text/plain";
 pub const DEFAULT_CHARACTER_SET: &str = "utf-8";
 const INTO_STRING_LIMIT: usize = 10 * 1_024 * 1_024;
@@ -281,6 +287,11 @@ impl Response {
                 .and_then(|l| l.parse::<usize>().ok())
         };
 
+        let compression = self
+            .header("content-encoding")
+            .map(Compression::from_header_value)
+            .flatten();
+
         let stream = self.stream;
         let unit = self.unit;
         if let Some(unit) = &unit {
@@ -292,12 +303,17 @@ impl Response {
         let deadline = unit.as_ref().and_then(|u| u.deadline);
         let stream = DeadlineStream::new(*stream, deadline);
 
-        match (use_chunked, limit_bytes) {
+        let body_reader: Box<dyn Read + Send> = match (use_chunked, limit_bytes) {
             (true, _) => Box::new(PoolReturnRead::new(unit, ChunkDecoder::new(stream))),
             (false, Some(len)) => {
                 Box::new(PoolReturnRead::new(unit, LimitedRead::new(stream, len)))
             }
             (false, None) => Box::new(stream),
+        };
+
+        match compression {
+            None => body_reader,
+            Some(c) => c.wrap_reader(body_reader),
         }
     }
 
@@ -505,6 +521,38 @@ impl Response {
         let previous_url = previous.get_url().to_string();
         self.history = previous.history;
         self.history.push(previous_url);
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+enum Compression {
+    #[cfg(feature = "brotli")]
+    Brotli,
+    #[cfg(feature = "gzip")]
+    Gzip,
+}
+
+impl Compression {
+    /// Convert a string like "br" to an enum value
+    fn from_header_value(value: &str) -> Option<Compression> {
+        match value {
+            #[cfg(feature = "brotli")]
+            "br" => Some(Compression::Brotli),
+            #[cfg(feature = "gzip")]
+            "gzip" | "x-gzip" => Some(Compression::Gzip),
+            _ => None,
+        }
+    }
+
+    /// Wrap the raw reader with a decompressing reader
+    #[allow(unused_variables)] // when no features enabled, reader is unused (unreachable)
+    fn wrap_reader(self, reader: Box<dyn Read + Send>) -> Box<dyn Read + Send> {
+        match self {
+            #[cfg(feature = "brotli")]
+            Compression::Brotli => Box::new(Decompressor::new(reader, 4096)),
+            #[cfg(feature = "gzip")]
+            Compression::Gzip => Box::new(GzDecoder::new(reader)),
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


jsha:
Thanks for the contribution! Just letting you know we've seen this and will review it soon.

src/response.rs
```diff
@@ -18,6 +18,12 @@ use serde::de::DeserializeOwned;
 #[cfg(feature = "charset")]
 use encoding_rs::Encoding;
 
+#[cfg(feature = "gzip")]
+use flate2::read::GzDecoder;
+
+#[cfg(feature = "brotli")]
+use brotli_decompressor::Decompressor;
```
algesten, line 25: Minor: If we import this as `use brotli_decompressor::Decompressor as BrotliDecoder` so we get symmetry between `GzDecoder` and `BrotliDecoder`.

algesten (commented):
Hey thanks for this PR!

This looks great to me.

The PR doesn't address sending bodies with compression, and whilst I that's fine to leave for  future PR, I think we should consider the feature flags names to future proof them.

Currently the feature flags are `gzip` and `brotli`. If we implement sending with compression,  are we just extending the meaning of those?

Is it a meaningful granularity to be able to turn off `brotli` and keep `gzip`? Maybe we should go the other way and just have `compression` and `decompression` as the flags?

I don't have an answer here, but I wanted to get the discussion going.

jsha:
For the record, here are the transitive dependencies brought in. Nice and small, as expected.

```
├── brotli-decompressor v2.3.2
│   ├── alloc-no-stdlib v2.0.3
│   └── alloc-stdlib v0.2.1
│       └── alloc-no-stdlib v2.0.3
├── flate2 v1.0.22
│   ├── cfg-if v1.0.0
│   ├── crc32fast v1.2.1
│   │   └── cfg-if v1.0.0
│   ├── libc v0.2.94
│   └── miniz_oxide v0.4.4
│       └── adler v1.0.2
│       [build-dependencies]
│       └── autocfg v1.0.1
```

> The PR doesn't address sending bodies with compression, and whilst I that's fine to leave for future PR, I think we should consider the feature flags names to future proof them.

I think sending bodies with compression should be a per-Request option rather than a feature. My understanding is that it's quite rare for servers to handle compressed bodies, and so this would only be something you'd apply when you know you're sending to such a server.

> Is it a meaningful granularity to be able to turn off brotli and keep gzip?

I'm on the fence here. The main reason I can think of would be to reduce code size or extra dependencies. Also, what about platform support? Are gzip and brotli buildable on all the same platforms?

Ultimately I'd like at least gzip to be one of our default features since it's such a widespread part of the web platform and is a trivial way to improve performance in a lot of cases. Maybe also brotli, though brotli is much less widely used so far.

src/response.rs
AlyoshaVasilieva: Makes sense to me, done.

AlyoshaVasilieva:
In the past I've specifically enabled one compression method on a client when I knew it was the only one supported by the only server the client was going to talk to; no reason to waste time compiling the code of multiple compression methods when the only thing the server will use is gzip.

`flate2` (with its default features) and `brotli-decompressor` are both pure Rust, so no platform concerns that I know of.

jsha (approved):
This looks good to me. @algesten what do you think?

@AlyoshaVasilieva are you up for writing some unittests?

Thinking about the question of making gzip enabled by default: It's a bit of a breaking change because someone could be parsing the Encoding header and doing gunzip themselves. If we start decompressing automatically before the data reaches the user, we could break them. Note that this also runs into some of the same "spooky action at a distance" that can result from features: if you have an application that decompresses gzip, and you add a dependency that turns on the gzip feature in ureq, you'll start getting already-decompressed data out of ureq.

algesten:
I think this is almost ready to be merged. The two things are:

[ ] Some unti test
[ ] Update documentation of feature flags

AlyoshaVasilieva:
Added some documentation of features.

I'm not confident in my ability to write a proper unit test for this.

Re: breaking, I think someone manually adding the Encoding header and decompressing responses themself would be quite weird to be doing. Not that that means nobody is doing it, of course, such as [this code](https://github.com/maximumstock/poe-stash-indexer/blob/f80a8396c00b26e9677558d7be732c642f2932e1/stash-api/src/sync/fetcher.rs#L149-L197) among others. I'd expect the vast majority of users to benefit from automatic gzip, though. And doing decompression yourself seems knowably risky to me, since automatic decompression is very common in HTTP clients.

jsha:
Thanks for the test @AlyoshaVasilieva !

A thought about compat-breaking: maybe when we do automatic decompression we should remove the content-encoding header? That way it would look to any downstream code as if the server sent it non-compressed. Downsides: Content-Length would also be a lie; it would become impossible to check whether a server sent a content-encoding header or not.

AlyoshaVasilieva:
The first usage of manual decompression I found on GitHub doesn't bother checking for the Content-Encoding header, it sets the Accept-Encoding header to gzip and then unconditionally gunzip's the response. No way to avoid breaking that, AFAIK (without bumping ureq to v3).

I *think* removing the Content-Encoding header (and Content-Length?) is proper, but I'm not actually sure how to write that - anyone checking them is doing so before `into_reader`, which is where most of the code I added is. I can try reworking the impl, though it won't make this a fully non-breaking change.

AlyoshaVasilieva:
Modified the implementation. It now determines compression during `Response::do_from_stream` instead of in `into_reader`; two fields have been added to the Response to allow this. When the client will be automatically decompressing the body, the Content-Encoding and Content-Length headers are removed. This should lower breakage, but anyone unconditionally decompressing a body will still have broken code. (They shouldn't really be doing that, though.)

Let me know if this seems like a better way to do things.

algesten:
@jsha i'm on the fence about modifying the response headers. Spontaneously i feel like i wouldn't expect that to happen.

I think we might rather bring out in documentation that reading the body might not exactly be `content-length` if compression features are on.

jsha:
@algesten I agree it feels "weird," but FWIW that's what Go's net/http does, and I've found it works great: https://pkg.go.dev/net/http

> 	// Uncompressed reports whether the response was sent compressed but
>	// was decompressed by the http package. When true, reading from
>	// Body yields the uncompressed content instead of the compressed
>	// content actually set from the server, ContentLength is set to -1,
>	// and the "Content-Length" and "Content-Encoding" fields are deleted
>	// from the responseHeader. To get the original response from
>	// the server, set Transport.DisableCompression to true.
>	Uncompressed bool

>	// DisableCompression, if true, prevents the Transport from
>	// requesting compression with an "Accept-Encoding: gzip"
>	// request header when the Request contains no existing
>	// Accept-Encoding value. If the Transport requests gzip on
>	// its own and gets a gzipped response, it's transparently
>	// decoded in the Response.Body. However, if the user
>	// explicitly requested gzip it is not automatically
>	// uncompressed.
>	DisableCompression bool

And I think this approach would cause less breakage of otherwise-valid programs that try to use the content-encoding header.

algesten:
@jsha alright. i buy that.

algesten:
Landing this is #453. Thanks!

PR #423, opened 2021-10-01
Remove trailing newline from testing response

jonhoo:
The newline becomes part of the response body, even though it was not passed in
as `body`, which makes `Response::new` difficult to use with anything that
checksums data contents for example. Arguably there should also be a mechanism
for passing in `[u8]` bodies, but that's for a separate feature request.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -109,7 +109,7 @@ impl Response {
     /// # }
     /// ```
     pub fn new(status: u16, status_text: &str, body: &str) -> Result<Response, Error> {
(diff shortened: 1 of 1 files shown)


algesten (approved):
Yeah, I agree.

This is a breaking change, but hopefully not too bad since it's not part of `ureq` "core functions", but more a helper for unit tests.

@jsha you think we can merge it?

jsha:
Yes, I think this is good to merge. I would consider it a bug fix for the intended functionality of Response::new.

jonhoo:
Amazing, thank you! Do you have a rough ETA for when you're planning the next release?

PR #424, opened 2021-10-01
get rid of warnings when building

mcr:
While I started this branch testing --no-default-features, I later realized that TLS had nothing to do with these warnings.
But, I wanted a clean slate for work to add mbedtls support.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -94,7 +94,8 @@ pub struct Transport {
     message: Option<String>,
     url: Option<Url>,
     source: Option<Box<dyn error::Error + Send + Sync + 'static>>,
-    response: Option<Response>,
+    #[allow(dead_code)]
+    response: Option<Response>,  // when building without tls, only tests read this.
 }
 
 /// Extension to [`Result<Response, Error>`] for handling all status codes as [`Response`].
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -94,7 +94,8 @@ pub struct Transport {
     message: Option<String>,
     url: Option<Url>,
     source: Option<Box<dyn error::Error + Send + Sync + 'static>>,
-    response: Option<Response>,
+    #[allow(dead_code)]
+    response: Option<Response>,  // when building without tls, only tests read this.
```
jsha, line 98: This is surprising! We should make sure we have test coverage of this code path in the non-TLS case, too.

jsha (commented):
Cool that you are working on mbedtls support. You'll be interested in https://github.com/algesten/ureq/pull/391 and https://github.com/algesten/ureq/issues/319 then.

src/error.rs
mcr: I realized that actually it happens even for the default, at least it does with my rustc 1.57.0-nightly (5ecc8ad84 2021-09-19)

jsha: Aha, you are correct! Thanks for pointing that out. Now that I look at it again the reason is clear: a Response is only available on a Status error; if we got a Transport error it doesn't actually make sense to have a Response, because the error occurred before we could read a response.

jsha:
Merged in #425.

PR #426, opened 2021-10-02
Add support for arbitrary TLS backends (rebased, some fixes)

mcr:
This rebases #391 on origin/main, and adds some work to make doctests and --no-default-features work.
The pattern of Arc<> may not be as intended, as I had to update some doctests.
Also the way I omitted the "tls" doctests is probably not the best solution.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           command: doc
           # Keep in sync with Cargo.toml's [package.metadata.docs.rs]
-          args: --no-default-features --no-deps --features "tls json charset cookies socks-proxy"
+          args: --no-default-features --no-deps --features "tls native-tls json charset cookies socks-proxy"
   build_and_test:
     name: Test
     runs-on: ubuntu-latest
@@ -48,6 +48,8 @@ jobs:
         tls:
           - ""
           - tls
+          - native-tls
+          - "tls native-tls"
         feature:
           - ""
           - json
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2018"
 
 [package.metadata.docs.rs]
 # Keep in sync with .github/workflows/test.yml
-features = [ "tls", "json", "charset", "cookies", "socks-proxy" ]
+features = [ "tls", "native-tls", "json", "charset", "cookies", "socks-proxy" ]
 
 [features]
 default = ["tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,8 @@ serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
 encoding_rs = { version = "0.8", optional = true }
 cookie_store = { version = "0.15", optional = true, default-features = false, features = ["preserve_order"] }
-log = "0.4"
+log = "0.4.11"
+native-tls = "0.2.7"
 
 [dev-dependencies]
 serde = { version = "1", features = ["derive"] }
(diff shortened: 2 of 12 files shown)


mcr:
Please review @jsha.

jsha:
Thanks for doing this work, @mcr! It looks like you missed my most recent commit on that PR, https://github.com/algesten/ureq/pull/391/commits/ba18080e7580fb0c8f5a6858c5d37f783e00bcc7. In that commit I significantly changed the approach. Now native-tls can only be configured by calling `.tls_connector()` on an Agent builder. The goal there is to avoid "spooky action at a distance," where adding a dependency to your tree could change which TLS backend is used by ureq (due to diamond dependencies).

mcr:
> Thanks for doing this work, @mcr! It looks like you missed my most recent commit on that PR, [ba18080](https://github.com/algesten/ureq/commit/ba18080e7580fb0c8f5a6858c5d37f783e00bcc7). 

I did miss it.  I've read it this morning.  
How can we collaborate better?

> In that commit I significantly changed the approach. Now native-tls can only be configured by calling `.tls_connector()` on an Agent builder. The goal there is to avoid "spooky action at a distance," where adding a dependency to your tree could change which TLS backend is used by ureq (due to diamond dependencies).

I think that I understand, but I won't be able to explain it to others.  I think that I need a picture/example :-)

algesten:
> I think that I understand, but I won't be able to explain it to others. I think that I need a picture/example :-)

I'll give it a go. Here's a diamond dependency.

```
       MyApp
       /  \
    LibA   LibB
       \  /
       ureq
```

* `LibA` and `LibB` are some higher level libraries that consumes some web-api by implicitly using `ureq`. Here  [acme-lib](https://crates.io/crates/acme-lib), and example of a convenience lib to talk to Let's Encrypt that uses `ureq`.
* `LibA` and `LibB` can both enable feature flags for `ureq`.
* Then `LibA` must for some reason must use the feature flag `native-tls` (imagine that the guys at Let's Encrypt don't know anything about security, so they require TLS 1.0 😁).
* Rust feature flag resolution is additive, so `MyApp` gets the sum of flags from both `LibA` and `LibB`.

If `ureq` automatically switched to `native-tls` when that feature flag is present, it means `MyApp` and `LibB` both would use `native-tls` for all ureq calls, also when those could use (the much more secure) `rustls`. This is quite devious, because enabling a feature flag might not be considered a breaking change for `LibA` (since it effectively provides an API), so even if the user carefully vetted that their app uses `rustls`, a simple patch update of some deps could suddenly kick the entire app into using `native-tls`.

By making `native-tls` an explicit config option on `Agent` there is no way that a deep down dependency could enable `native-tls` for everything.

mcr:
There are some other considerations, which the feature system can't really accomplish. For instance, I might need to use a particular backend: acceleration, code size (there shall be only one TLS), FIPS certification... so in the LibA/LibB situation, error might be the right answer.  The feature system can't cope with that, alas.
I'm happy with your solution, but the current cfg() options do not properly support that yet.

jsha:
> I did miss it. I've read it this morning.

No problem!

> How can we collaborate better?

Well, sometimes mistakes happen and it's hopefully not a big deal. One way to improve collaboration is to make sure to tell us when you're starting work on something. If I'd known you were building on top of that branch, I'd have made sure to let you know I pushed an update. Also, since it sounds like you're interested in ongoing ureq development, you could hit the "Watch" button at the top of https://github.com/algesten/ureq/. That way you'll see every conversation that happens on the repo. If that's too much, you can "Subscribe" to specific issues or PRs (the button is down along the right-hand side of the issue or PR page).

Great explanation @algesten , thanks!

PR #438, opened 2021-11-15
Add src to rustls error, bump to 0.20.1

AlyoshaVasilieva:


jsha (approved):
Thanks!

algesten:
Notice that we should not have patch version in the rustls dependency. `0.20` is correct since that will allow the end user to get .1 .2 etc on the patch level.

Never mind reverting this though since I fixed it in this branch landing next #449

jsha:
In general that is correct, but in this case we are depending on an API that was introduced in 0.20.1, so we do need to include the patch level. Perhaps that should technically have been a semver minor release (since it introduced an API).

algesten:
Ah got ya!

PR #439, opened 2021-11-21
Fix native-certs feature

amoskvin:
With ureq 2.3.0, the native-certs feature is not working due to incorrect `cfg` check, which prevents the code to load native certs from being built. This fixes the `cfg` check and updates the code to work with rustls-native-certs 0.6.


algesten:
This has been fixed in main.

amoskvin:
Thanks. Could you please publish it on crates.io?

PR #441, opened 2021-11-26
Fix two TLS things

lolgesten:


algesten:
@jsha does this look ok to you?

tests/https-agent.rs
```diff
@@ -23,6 +23,8 @@ fn agent_set_header() {
     assert_eq!("value", json.headers.get("Header").unwrap());
 }
 
+// From here https://badssl.com/download/
+// Decrypt key with: openssl rsa -in ./badssl.com-client.pem
```
jsha, line 27: Nit: technically this _parses_ the key. It doesn't decrypt it because the key is not encrypted.

jsha (approved):
Looks good!

tests/https-agent.rs
algesten: the key i downloaded from that url had a passphrase that is asked for when running this openssl command. 

is it just obfuscated and not encrypted?

PR #443, opened 2021-12-09
Preserve Authorization header in same host redirects

### LINKED ISSUE #442: Authentication header is not preserved in Same-Host Redirects
GIthub Rest api devolve to unauthenticated access if Authentication header isn't preserved in 301 redirects.  
cUrl -L do follow the redirect preserving the header.

llde:
Currently the Authorization header isn't preserved in redirect at all. This breaks the usage of github rest api  when trying to connect to an url that return 301 redirect. 

The change
If the Redirect is to the same host, the connection scheme is equal (no downgrade to http, no promotion to https), and the port is the same, the redirect preserve the Authorization header.
send_prelude now take the previous &[String] instead of the redir bool

Fix #442 
Also remove dangerous_configuration from rustls in dev-dependancies.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -14,6 +14,16 @@ use {
     cookie_store::CookieStore,
 };
 
+/// Specify the strategy for propagation of authorization headers during Redirects
+///
+/// Never is the default strategy and never send authorization headers in Redirects
+/// SameHost send the authorization header in Redirects only if the host of the Redirect is the same of the previous request, and both use the Https scheme
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RedirectAuthHeaders {
+    Never,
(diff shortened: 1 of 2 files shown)


algesten:
Thanks for taking a stab at this!

There are a couple of issues with this PR. First, it would be nice to focus it on only making the necessary changes. The Cargo.toml change removing `dangerous_configuration` is not right since the example `cureq` uses this feature. It would also be nice to not have whitespace changes (two blank rows removed). You can see how the tests fail since our lint pass runs `cargo fmt` to check the format of the code.

But before addressing those issues, we need to take a step back and look at the big picture. I'll put the discussion in the issue #442

algesten:
This work continue in #445

PR #445, opened 2021-12-11
Preserve Authorization header in same host redirects

### LINKED ISSUE #442: Authentication header is not preserved in Same-Host Redirects

algesten:
This PR builds on the work by @llde in #443. 

Currently the stripping of auth headers on redirect is done at the last stage, just when serializing the `Unit` to the socket stream.

This PR moves the logic for stripping auth header to the construction of the `Unit` for the redirect. That makes the logic testable (and tests are added).

Close #442

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -465,7 +465,7 @@ impl AgentBuilder {
     ///
     /// Defaults to [`RedirectAuthHeaders::Never`].
     ///
-    pub fn set_redirect_auth_headers(mut self, v: RedirectAuthHeaders) -> Self {
+    pub fn redirect_auth_headers(mut self, v: RedirectAuthHeaders) -> Self {
         self.config.redirect_auth_headers = v;
         self
     }
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -20,14 +20,16 @@ use {
 ///
 /// `Never` is the default strategy and never preserves `authorization` header in redirects.
 /// `SameHost` send the authorization header in redirects only if the host of the redirect is
-/// the same of the previous request, and both use the `https` scheme.
+/// the same of the previous request, and both use the same scheme (or switch to a more secure one, i.e
+/// we can redirect from `http` to `https`, but not the reverse).
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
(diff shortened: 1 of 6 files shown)


algesten:
@llde Thank you for your work on this! I was working on a review, but it snowballed in complexity (header stripping being done in the wrong place), so I opted for finishing it myself.

@jsha I believe this is ready to land, want to review?

algesten:
@jsha gentle reping for review. If you don't have time, no problem!

src/agent.rs
```diff
@@ -14,6 +14,21 @@ use {
...
+/// `Never` is the default strategy and never preserves `authorization` header in redirects.
+/// `SameHost` send the authorization header in redirects only if the host of the redirect is
+/// the same of the previous request, and both use the `https` scheme.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum RedirectAuthHeaders {
+    /// Never preserve the `authorization` header on redirect. This is the default.
+    Never,
+    /// Preserve the `authorization` header when the redirect is to the same host. Must
+    /// be under the `https` scheme (though port can differ).
```
jsha, line 28: I don't think the `https` scheme restriction makes sense here since we allow setting the authorization header on non-https. Instead, how about: the scheme must be the same. That way if you start on https you can't accidentally carry the authorization header on a redirect to http.

jsha (changes requested):
Generally looks great! I'm glad you've done a little tidying of how the headers are handled on redirects; that had been bugging me for a while.

Also I thought a useful thing to check in with would be how the Fetch standard governs authorization headers on redirects: https://fetch.spec.whatwg.org/#http-redirect-fetch. I couldn't figure it out from the doc. Might be nice to try a sample call to the `fetch()` in a browser to see what it does.

src/agent.rs
algesten: Or maybe we allow http -> https but not the reverse?

algesten: I went with "same or http -> https."

algesten:
@jsha I believe we're ready to land this one now.

PR #448, opened 2021-12-16
Middleware feature

### LINKED ISSUE #444: Setting headers in an AgentBuilder
I used pre-2.0 ureq before. I noticed that `Agent` now doesn't have a set method. The Changelog says this:

> Create AgentBuilder to separate the process of building an agent from using the resulting agent. Headers can be set on an AgentBuilder, not the resulting Agent.

But I don't see any way to set a header on the AgentBuilder either. Am I missing something?

algesten:
Implementation of middleware chains in Agent. 

Whether this is overkill to set some simple headers for every request going through an agent… maybe. But it could be very useful implementing oauth refresh token logic etc.

Close #444


algesten:
Reading the code with fresh eyes, I'm leaning towards dropping the `Extensions` part since it's not necessary and maybe confuses more than help.

algesten:
On the other hand – the extensions are so nice for making a quick middleware. The nice `MyCounter` examples just won't be as simple without it.

jsha:
I'd like to get rid of the Extension mechanism. I agree that it's not necessary, and it adds significant complexity. Also it allows middlewares to accidentally block the whole agent, causing bad performance.

Imagine you have two middlewares, each of which has its own internal state. Ideally they should each have their own internal mutexes so they don't need to contend with each other, and so they can choose their own guarantees.

For instance, the Counter example might look like this instead:

```rust
#[derive(Debug, Default)]
struct CounterState {
    request_count: AtomicUsize,
    total_bytes: AtomicUsize,
}

impl Middleware for CounterState {
    fn handle(&self, request: Request, next: MiddlewareNext) -> Result<Response, Error> {
        self.request_count.fetch_add(1, Relaxed);
        let response = next.handle(request)?;
        let len = response
            .header("Content-Length")
            .and_then(|s| s.parse::<usize>().ok())
            .unwrap();
        self.total_bytes.fetch_add(len, Relaxed);
        Ok(response)
    }
}
```

It seems like the signature for handle should actually be:

```rust
    fn handle(&self, request: Request, next: impl Middleware) -> Result<Response, Error> {
```

To abstract away the fact that we use MiddlewareNext internally. Then we can `impl Middleware for MiddlewareNext`, and unexport it.

algesten:
> I'd like to get rid of the Extension mechanism.

OK! It's going. 

> It seems like the signature for handle should actually be:
> `fn handle(&self, request: Request, next: impl Middleware) -> Result<Response, Error>`

I don't think that works. There are two problems.

1. What are we passing for `next` when continuing the chain?
2. By taking `self` we guarantee the continuation is called only once.

(the inspiration here is [nodejs express](http://expressjs.com/en/guide/using-middleware.html) where `next` also had a different type signature to the middleware call itself)

jsha:
Hm, you're right that `impl Middleware` doesn't work.

Another thought on the API: What about having three methods on `trait Middleware`:

```
fn before(&self, &mut Request) -> Result<(), Error> {}
fn after(&self, &mut Response) -> Result<(), Error> {}
fn handle(...)
```

All would have a default implementation. That way someone who only cared about before-handling, or only cared about after-handling, could implement just the method they care about. We could also have trait impls for those so you could pass in a plain function.

algesten:
@jsha I'm quite strongly biased towards this variant because it's how people with my background are used to middleware working.

I think it's really quite elegant that it's one function for pre- or post handling.

* https://expressjs.com/en/guide/using-middleware.html
* https://docs.rs/reqwest-middleware/0.1.3/reqwest_middleware/
* https://docs.rs/tide/0.16.0/tide/trait.Middleware.html

There are also examples of a pre-/post-model as well (rocket for instance).

Does that extra `MiddlewareNext` bother you that much?

jsha:
> Does that extra MiddlewareNext bother you that much?

Nope, once I realized the "consistent" variant wouldn't work, and the internals of MiddlewareNext are hidden, not a big deal.

> I think it's really quite elegant that it's one function for pre- or post handling.

Indeed! And I'm familiar with this style from gRPC: https://github.com/grpc-ecosystem/go-grpc-middleware#go-grpc-middleware

What I was thinking about above is not a replacement for the "single call" style of middleware but an addition to make it more convenient. I should have written out the implementation a little more to make it clear:

```rust
pub trait Middleware {
  fn before(&self, r: &mut Request) -> Result<(), Error> {}
  fn after(&self, r: &mut Response) -> Result<(), Error> {}
  fn handle(&self, r: mut Request, next: MiddlewareNext) -> Result<Response, Error> {
    self.before(&mut r)?;
    let resp = next.handle(r)?;
    self.after(&mut resp)?; 
  }
}
```

So the core is still `handle()`, which wraps a call to next. But because all the function have a default impl, you could choose to implement just `before`, or just `after`.

Another way to accomplish the same thing without default impls:

```
// same as current version
pub trait Middleware {
  fn handle(&self, r: Request, next: MiddlewareNext) -> Result<Response, Error>;
}

pub trait BeforeRequest {
  fn before(&self, r: &mut Request) -> Result<(), Error>;
}

impl Middleware for B: BeforeRequest {
  fn handle(&self, r: mut Request, next: MiddlewareNext) -> Result<Response, Error> {
    self.before(&mut r)?;
    next.handle(r)
  }
}

impl BeforeRequest for Fn(&mut Request) -> Result<(), Error> {
  fn before(&self, r: &mut Request) -> Result<(), Error> {
    self.call(r)?;
  }
}

// And imagine equivalents for AfterRequest
```

This would allow convenient very short impls like `|r: &mut Request| r.set("Accept-Encoding", "identity")` for cases that only care about before or after handling (and before is probably the much more common case), while still allowing the elegant wrapping case for things like metrics collection (that want to grab a timestamp before and after).

Anyhow, this is a convenience aspect and doesn't have to land in the first iteration. It may in fact be better to let it wait for a subsequent iteration so we get a little more experience.

algesten:
Oh. Cool. I didn't catch that. This is indeed very nice.

Both your proposed solutions are nice. Maybe we push this first iteration out and then discuss the relative merits in a followup?

jsha:
Sounds good! I'll dig further into the implementation. One early question: Why is the chain of middlewares implemented as a linked list rather than a Vec? Also, do we need them to be `Arc` or can they be `Box` (since the Agent itself is already Arc, and we only need shared references to the middlewares).

src/request.rs
```diff
@@ -125,16 +128,38 @@ impl Request {
...
+                req.headers,
+                &reader,
+                deadline,
+            );
+
+            unit::connect(unit, true, reader).map_err(|e| e.url(url.clone()))
+        };
+
+        // This clone is quite cheap since either we are cloning the Optional::None or a Vec<Arc<dyn Middleware>>.
+        let maybe_middleware = agent.state.middleware.clone();
```
jsha, line 146: Can MiddlewareNext instead store an iterator over the `middleware` Vec?

algesten:
I was pondering whether it would be possible to construct the chain already in agent. It would require some way of getting the end actual request to the last step.

src/agent.rs
```diff
@@ -112,6 +111,7 @@ pub(crate) struct AgentState {
     #[cfg(feature = "cookies")]
     pub(crate) cookie_tin: CookieTin,
     pub(crate) resolver: ArcResolver,
+    pub(crate) middleware: Option<Vec<Arc<dyn Middleware + Send + Sync>>>,
```
jsha, line 114: I think we can make this just `Vec<Arc<dyn Middleware + Send + Sync>>`, since the "no middleware" case is adequately represented by an empty vec.

https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees

> if you construct a Vec with capacity 0 via Vec::new, vec![], Vec::with_capacity(0), or by calling shrink_to_fit on an empty Vec, it will not allocate memory.

Also I think we can get rid of the `Arc` if we do: `Vec<Box<dyn Middleware + Send + Sync + 'static>>`.

src/agent.rs
```diff
@@ -594,6 +596,56 @@ impl AgentBuilder {
...
+
+    /// Add middleware handler to this agent.
+    ///
+    /// All requests made by the agent will use this middleware. Middleware is invoked
+    /// in the order they are added to the builder.
+    pub fn middleware(mut self, m: impl Middleware + Send + Sync + 'static) -> Self {
+        if self.middleware.is_none() {
+            self.middleware = Some(vec![]);
+        }
+        self.middleware.as_mut().unwrap().push(Arc::new(m));
```
jsha, line 608: Changing middleware from `Option` to `Vec` as I propose above lets us remove this unwrap.

src/middleware.rs
```diff
@@ -0,0 +1,169 @@
...
+/// # fn main() -> Result<(), ureq::Error> {
+/// # ureq::is_test(true);
+/// use ureq::{Request, Response, Middleware, MiddlewareNext, Error};
+/// use std::sync::atomic::{AtomicU64, Ordering};
+/// use std::sync::Arc;
+///
+/// // Middleware that stores a counter state. This example uses an AtomicU64
+/// // since the middleware is potentially shared by multiple threads running
+/// // requests at the same time.
+/// struct MyCounter(Arc<AtomicU64>);
```
jsha, line 104: Probably this makes a bit more sense as:

```
struct MyCounter(AtomicU64);
```

And then down below we construct `Arc::new(MyCounter::new())`, and give one close to the Agent and keep one clone for ourselves. In other words, `Arc` isn't part of the internal guarantees of MyCounter, it's just there so we can share ownership with Agent (which requires a 'static lifetime bound).

jsha (changes requested):

src/request.rs
algesten: Maybe. I'll take a stab at it.

src/agent.rs
algesten: Wanted to save on some unnecessary heap allocation since also an empty array would allocate capacity.

algesten: Arc is there because we don't want to force Middleware impl to also do clone.

The concurrency in Request do_call currently builds on cloning the middleware chain, hence Arc instead of Box. 

Maybe that can change if we get the iterator in MiddleWare next working.

src/middleware.rs
algesten: Alright. I was trying to mirror how the example is doing it where an arc/mutex protects an inner state.

Not sure what is less confusing to a beginner? 🤔

src/agent.rs
jsha: > since also an empty array would allocate capacity.

But the docs I linked for Vec (https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees) guarantee that an empty array does not allocate.

jsha: > Arc is there because we don't want to force Middleware impl to also do clone.

I think the most common case is that the Middleware impl is fully given to use - e.g. a `fn`.

There's another common case, like metrics, where the user wants an `Arc`. But in that case it's because they want to retain a handle to the middleware so they can collect data, like the example you have with Counter. But in that case, the user needs to make the Arc themselves and pass us a clone of it. So for instance in the MyCounter example we wind up with two Arcs: one built inside the Agent, and one built inside the MyCounter. It would be nicer if the user makes an `Arc<MyCounter>` and passes us a clone; that way there's only one Arc.

Maybe the best thing is if we can make a param that accepts a T or an Arc. If we get a T, we Arc::new() it; if we get an Arc, we clone it. I'm not seeing immediately what that param would look like, but I bet it exists.

src/middleware.rs
jsha: Now that you mention it, I also think `MyMiddleware` should be `struct MyMiddleware(Mutex<MyState>)` and we should pass `Arc<MyMiddleware>` to the Agent.

src/agent.rs
algesten: > guarantee that an empty array does not allocate.

TIL <3

jsha: I learned it today too! My thought process was: "I bet this is to avoid allocations. But wait, does it really allocate when empty?" Very neat to learn about.

src/middleware.rs
algesten: I think this is not clear separation of concerns.

A common pattern for shared mutable state is `Arc<Mutex<T>>`. A middleware working with such a state is then naturally `MyMiddleware(Arc<Mutex<T>>)`.

That we want an additional `Arc<dyn Middleware>` to facilitate cloning of the middleware is a different issue. Sure. Technically these two could just be one single `Arc`, but I think it confuses two ideas.

src/agent.rs
algesten: > I think the most common case is that the Middleware impl is fully given to use - e.g. a fn.

I think I'm not making myself understood here. I will take a stab at the `iterator` approach in `MiddlewareNext` and see if we can simply side step this whole discussion :)

algesten:
@jsha I pushed a version based in `Iterator`.

I was wrangling back/forth with the `Arc<dyn Middleware>` and I struggle to get rid of it.

The problem is that we have two "conflicting" lifetimes. That of `Payload<'a>` and that of a borrowed iterator of the `&'b Agent.state.middleware`.

It might be possible to solve this with making `Middleware<'a, 'b>` and say that a must live as long as b (or reverse, i always struggle to tell).

However it's late and I my brain is a mush. Feel free to take a stab at it!

jsha:
Alright, I think I've got it, at least for the `MiddlewareNext` not having additional Arcs. I haven't taken a stab at the `Vec<Arc<dyn Middleware>>` yet. https://github.com/algesten/ureq/compare/main...jsha:middleware-ref?expand=1

https://github.com/algesten/ureq/commit/7aa36bca2fa848694643ee5f15f46800f8e14a6b

Also I suspect we can get rid of the `Box` on `request_fn` but haven't figure that out yet.

algesten:
Cloning agent eh? I was resisting that, but I guess it's no worse than cloning the chain.

algesten:
@jsha I think this might be ready now. Thanks for showing the way!

algesten:
I made one final change that I came up with last night. Since middleware always needs to be `Send + Sync + 'static` we can move that bound to the trait definition and thus get a simpler type where we use it.

Particularly I think this bit on `AgentBuilder` improves the API a lot for someone new to Rust and ureq:

```
-    pub fn middleware(mut self, m: impl Middleware + Send + Sync + 'static) -> Self {
+    pub fn middleware(mut self, m: impl Middleware) -> Self {
```

jsha:
Ooh great improvement! This is looking quite nice. I'll give it a thorough last review tomorrow.

src/agent.rs
```diff
@@ -594,6 +596,53 @@ impl AgentBuilder {
...
+#[cfg(feature = "tls")]
+#[derive(Clone)]
+pub(crate) struct TLSClientConfig(pub(crate) Arc<rustls::ClientConfig>);
+
+#[cfg(feature = "tls")]
+impl fmt::Debug for TLSClientConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("TLSClientConfig").finish()
+    }
+}
```
jsha, line 619: These seem like an unrelated change?

jsha (approved):
Looks good!

src/agent.rs
algesten: The debug stuff?

It's because deriving debug would require `Middleware` to also be debug, which felt unnecessary. 

The tls debug changes was because my preferred form is `impl fmt::Debug` and `use std::fmt` so I changed that one too so all are consistent in this file.

PR #449, opened 2021-12-17
Add support for alternate TLS implementations.

algesten:
This PR tidies up the last bits for releasing the `native-tls` feature where the majority of work was by @jsha in the #391 PR.


jsha (approved):
Awesome, thanks for picking this up! I think it is finally ready to go.

Once this is landed let's publish a release candidate and ask some of the folks who need the feature to test before we finalize it.

PR #450, opened 2021-12-18
Release 2.4.0

algesten:
PR for releasing 2.4.0.

To be decided whether we are including #448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,15 @@
+# 2.4.0
+
+ * Enable `gzip` feature by default (#455)
+ * `gzip` and `brotli` feature flags to enable decompression (#453, #421)
+ * Middleware function on agent (#448)
+ * Agent option to preserve `Authorization` header on redirects (#445)
+ * Deprecate re-exported serde types (#446)
+ * Widen type of `send_json` to `impl Serializable` (#446)
+ * `native-tls` feature provides an alternative TLS backend (#449, #391)
+
 # 2.3.2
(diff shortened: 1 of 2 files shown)


jsha:
This looks good! Before we release: What do you think of making `gzip` enabled by default? I think it would be a really good thing. It would save people lots of bytes with no trouble. As a concrete example, Go's net/http does gzip by default and transparently, and it's quite wonderful.

I'll try to send a patch/proposal today. I'd also like to take another pass at the interface for enabling gzip/brotli so we can make them more independent of feature flags and diamond dependency issues.

algesten:
Yeah let's enable gzip by default. Sounds good to me. I'll make a separate PR for that.

algesten:
@jsha I assume you're ok with releasing 2.4.0 now?

CHANGELOG.md
```diff
@@ -1,4 +1,15 @@
+# 2.4.0
+
+ * Enable `gzip` feature by default (#455)
+ * `gzip` and `brotli` feature flags to enable decompression (#453, #421)
+ * Middleware function on agent (#448)
+ * Agent option to preserve `Authorization` header on redirects (#445)
+ * Deprecate re-exported serde types (#446)
+ * Widen type of `send_json` to `impl Serializable` (#446)
+ * `native-tls` feature provides an alterantive TLS backend (#449, #391)
```
jsha, line 9:
```suggestion
 * `native-tls` feature provides an alternative TLS backend (#449, #391)
```

jsha (approved):

PR #454, opened 2021-12-19
Fix chunked encoding with broken ending

### LINKED ISSUE #325: "Error while decoding chunks" on some websites
On some websites, e.g. http://banxetoyota.vn, ureq fails with the following error: 

> Error while decoding chunks

However, curl and Firefox work fine.

There's 38 such websites in the top million (I'm using [Tranco list generated on the 3rd of February](https://tranco-list.eu/list/3G6L)).

Archive with all occurrences: [ureq-error-decoding-chunks.tar.gz](https://github.com/algesten/ureq/files/5984187/ureq-error-decoding-chunks.tar.gz)

algesten:
Some websites do not end chunked encoding correctly. The spec says the end should be:

```
0\r\n
\r\n
```

However some sites stop short after `0\r\n`. Curl handles this case fine.

Close #325

### DIFF
--- a/src/chunked/decoder.rs
+++ b/src/chunked/decoder.rs
@@ -125,6 +125,39 @@ where
             _ => Err(IoError::new(ErrorKind::InvalidInput, DecoderError)),
         }
     }
+
+    // Sometimes the last \r\n is missing.
+    fn read_end(&mut self) -> IoResult<()> {
+        fn expect_or_end(
+            bytes: &mut impl Iterator<Item = IoResult<u8>>,
+            expected: u8,
+        ) -> IoResult<()> {
+            match bytes.next() {
+                Some(Ok(c)) => {
+                    if c == expected {
+                        Ok(())
+                    } else {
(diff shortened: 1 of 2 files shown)


algesten:
I don't actually have a plan for how to fix this just yet.

I don't know whether the people behind the `chunked_transfer` crate would be open to a fix where the final `\r\n` is optional. Otherwise we need to handle it un ureq.

Shnatsel:
This needs to be handled with care, since it might allow [HTTP desync attacks](https://portswigger.net/research/http-desync-attacks-request-smuggling-reborn) and request smuggling.

@jhwgh1968 might know more about this particular case.

algesten:
@Shnatsel Good point! I provided a naive fix in upstream repo – maybe needs more thought. 🤔  https://github.com/frewsxcv/rust-chunked-transfer/pull/11

jsha:
I comment on https://github.com/algesten/ureq/issues/325#issuecomment-1344904282 - I think we should first implement trailers and see if we still have this problem. At least one of the affected sites that I spot-checked was actually sending trailers.

PR #457, opened 2021-12-19
Run cargo test with minimal versions

jsha:
This can catch breakages that are hidden by latest version selection.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,3 +71,21 @@ jobs:
         with:
           command: test
           args: --no-default-features --features "${{ matrix.tls }}" "${{ matrix.feature }}"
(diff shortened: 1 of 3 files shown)


algesten (approved):
So with this we will test both for max version in the normal tests and min version?

jsha:
Yep exactly!

algesten:
Merge away!

algesten (commented):
I think we should keep the versioning consistent.

For `rustls` we went with `>=0.20.1`, however I suspect that's not correct since I believe it will match `1.0.0`

Maybe we make the caret explicit? Or drop it in all places?

algesten:
Let's redo this against 3.x

PR #458, opened 2021-12-19
Use ? instead of unwrap in examples

jsha:
This is recommended by the Rust API Guidelines:

https://rust-lang.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark

One exception: When we need to unwrap an Option, the examples still use
.unwrap(). The alternative would be something like
`.ok_or(SomeErrorHere)?`, which feels like an awkward way to deal with
an Option. This might get better with NoneError:

https://docs.rs/rustc-std-workspace-std/1.0.1/std/option/struct.NoneError.html

I also rearranged some examples that used turbofish to use type
annotations. I think type annotations are more familiar to early Rust
users (and they use fewer characters and less punctuation, which is
always nice).

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -157,7 +157,7 @@ impl Agent {
     /// use {url::Url, ureq::Response};
     /// let agent = ureq::agent();
     ///
-    /// let mut url: Url = "http://example.com/some-page".parse().unwrap();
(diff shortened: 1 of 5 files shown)


algesten (approved):
👍✨

PR #462, opened 2021-12-23
Rename trait HttpsStream -> ReadWrite and make it public

### LINKED ISSUE #461: Can not implement TlsConnector::connect
Hi, 
thanks for adding `native-tls` support.  
I am trying to implement `TlsConnector::connect` to do some additional server certificate verification but get the compiler error `module stream is private`. Am I doing sth wrong here? Is there any other way to access the TlsStream or `TlsStream::peer_certificate`?

```rust
impl TlsConnector for CustomTlsConnector {
    fn connect(&self, dns_name: &str, tcp_stream: TcpStream) -> Result<Box<dyn ureq::stream::HttpsStream>, Error> {
         todo!()
    }
}
```

ureq version 2.4.0 with features `native-tls` and `json` on Rust 1.54.0 (Windows x64)

algesten:
Also provide an example of how to use it.

Close #461

### DIFF
--- a/examples/custom-tls.rs
+++ b/examples/custom-tls.rs
@@ -1,0 +1,81 @@
+use std::fmt;
+use std::io;
+use std::net::TcpStream;
+use std::sync::Arc;
+
+use ureq::{Error, ReadWrite, TlsConnector};
+
+pub fn main() -> Result<(), Error> {
+    let pass = PassThrough {
+        handshake_fail: false,
+    };
+
+    let agent = ureq::builder().tls_connector(Arc::new(pass)).build();
+
+    let _response = agent.get("https://httpbin.org/").call();
+
+    // Uncomment this if handshake_fail is set to true above.
+    // assert_eq!(
+    //     _response.unwrap_err().to_string(),
+    //     "https://httpbin.org/: Network Error: Tls handshake failed"
+    // );
+
+    Ok(())
+}
+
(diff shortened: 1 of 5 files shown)


algesten:
@jsha since there is nothing in using `HttpsStream` because we forgot to export it (oops), I'm suggesting a name change to `ReadWrite`.

"Stream" has a quite specific meaning internally in ureq and there is nothing particularly "Https" with it.

Before pushing this fix, I'd like to see if we can make a blanket impl of this trait for everything that implements `Read + Write + Send + Sync + 'static`.

algesten:
@jsha what do you think about this change?

To close the bug, we need to just export the trait. But the PR is two things:

1. Rename and export the trait `ReadWrite`.
2. Drop the `Inner` trait and just use `ReadWrite`.

mcr (commented):
Thanks for adding this example.

erkbb:
Hi, 
thanks for the example. I think that `ErrorKind::msg` and `ErrorKind::src` should also be public. Otherwise a custom `TlsConnector` would be unable to create errors. Maybe it would even make sense to add a dedicated `ErrorKind::TlsHandshakeFailed` ?

algesten:
@erkbb I've updated the example showing how to go via `io::Error` to capture handshake errors. For now, I prefer this to exposing more of ureq API.

PR #465, opened 2022-01-05
Add tcp no_delay option

### LINKED ISSUE #390: use no_delay on the TcpStream
Hi,

https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_nodelay

This is basically a default for most modern http clients since it reduces latency of sending bytes.

I was going through the stream.rs file but I'm not sure how to implement it and make a pull request so I decided to open this issue sorry about that.

daniel0611:
Closes #390.

Defaults to set no_delay but can be overwritten by calling `agent.no_delay(false)`.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -54,6 +54,7 @@ pub(crate) struct AgentConfig {
     pub timeout_read: Option<Duration>,
     pub timeout_write: Option<Duration>,
     pub timeout: Option<Duration>,
+    pub no_delay: Option<bool>,
     pub redirects: u32,
     pub redirect_auth_headers: RedirectAuthHeaders,
     pub user_agent: String,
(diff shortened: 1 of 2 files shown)


algesten:
Hi! Thanks for this PR – looks great. 

I guess it's somewhat dodgy to change the default value, but I can't see how it would break any existing code and we have precedence in all other http libraries. 

Let's merge!

Thanks!

algesten:
@jsha any comments on the API?

jsha:
API looks good. It's not clear to me why the internal value is an Option<bool> rather than a bool?

src/agent.rs
```diff
@@ -238,6 +239,7 @@ impl AgentBuilder {
                 timeout_read: None,
                 timeout_write: None,
                 timeout: None,
+                no_delay: None,
```
algesten, line 242: let's change to straight up bool.

algesten (changes requested):

daniel0611:
Sorry, can't remember why I went with an Option. Makes no sense to me now either. I changed it to `bool` now.

jsha (approved):

algesten (approved):

Issue #392, opened 2021-06-04
Authentication


mleonhard:
I wish to use ureq to call Twilio's API which [uses HTTP authentication](https://www.twilio.com/docs/usage/security).  Let's add some methods to support this:

```rust
impl Response {
    pub fn set_digest_auth(self, username: impl AsRef<str>, password: impl AsRef<str>) -> Self {}
    pub fn set_basic_auth(self, username: impl AsRef<str>, password: impl AsRef<str>) -> Self {}
}
```

Alternatively, we could add something to the docs explaining another way to do it.

mleonhard:
A workaround is to include the password in the Url:
```rust
let mut url = Url::parse("https://lookups.twilio.com/v1/PhoneNumbers?Type=carrier").unwrap();
url.path_segments_mut()
    .unwrap()
    .push(&tel.inner().to_string());
url.set_username(&twilio_sid).unwrap();
url.set_password(Some(&.twilio_token)).unwrap();
let response = ureq::request_url("GET", &url).call().unwrap();
if response.status() == 404 {
    return None;
}
...
```

Unfortunately, ureq logs url which now includes the password:
```
INFO sending request GET https://username:----PASSWORD----@lookups.twilio.com/v1/PhoneNumbers/10001112222
```
See "Redact password in logs" https://github.com/servo/rust-url/issues/714 .

fkusche:
You can do it by adding the authentication header manually
```rust
fn get_basic_auth_header( user: &str, pass: &str ) -> String {
    let usrpw = String::from( user ) + ":" + pass;
    String::from( "Basic " ) + &base64::encode( usrpw.as_bytes())
}
```

And then
```rust
let resp: SendResponse = ureq::post( &url )
    .set("Accept", "application/json")
    .set("Authorization", &get_basic_auth_header(&cfg.token_id, &cfg.token))
    ...
```

strohel:
HTTP basic auth is as easy as adding a correctly formatted header, though adding a helper to create such a header would be convenient.

HTTP digest auth is slightly more involved as it requires a round-trip to the server and adapting the second query to the nonce in server's first response.

I see ureq has had some support for HTTP authentication in the past, but it has been removed:
https://github.com/algesten/ureq/blob/3b68dce23a67ffbf887e2109e9a6c41e17803d07/CHANGELOG.md?plain=1#L75-L76

@algesten would such a feature fit into ureq (would a PR adding that be welcome)?

algesten:
[...]

PR #468, opened 2022-01-21
Implement digest authentication middleware behind a feature flag

PabloMansanet:
In #392 , @algesten mentioned that it could be possible to consider default middleware for common features like Digest authentication. Here's a simple implementation, in case it's general enough to add to the main crate.

Some thoughts:
* I realized the structure of the project is pretty nice and flat, so it felt wrong to create a subfolder for particular middlewares. I ended up adding this as a feature-flagged embedded module under `digest`. Is this a good approach?
* I'm not sure about using external services like httpbin for unit tests, but I saw other tests doing a similar thing so I went with what seemed to be the simplest approach. I'm happy to take directions on how to mock this instead to fully test it offline.
* The recursive solution helps transparently navigate the "back and forth" between server and client without needing any `Arc<Mutex<>>` state in the middleware itself, but this may be erring on the "too smart" side. I don't directly see any problems it might cause but I could definitely be missing some.

### DIFF
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -186,7 +186,7 @@ pub mod digest {
     /// let arbitrary_username = "MyUsername";
     /// let arbitrary_password = "MyPassword";
     /// let digest_auth_middleware =
-    ///     ureq::DigestAuthMiddleware::new(arbitrary_username.into(), arbitrary_password.into());
+    ///     ureq::DigestAuthMiddleware::new(arbitrary_username, arbitrary_password);
     /// # let url = String::new();
     ///
     /// let agent = ureq::AgentBuilder::new().middleware(digest_auth_middleware).build();
@@ -198,8 +198,14 @@ pub mod digest {
     }
 
     impl DigestAuthMiddleware {
-        pub fn new(username: Cow<'static, str>, password: Cow<'static, str>) -> Self {
-            Self { username, password }
+        pub fn new(
+            username: impl Into<Cow<'static, str>>,
+            password: impl Into<Cow<'static, str>>,
+        ) -> Self {
+            Self {
+                username: username.into(),
+                password: password.into(),
+            }
         }
 
         fn construct_answer_to_challenge(
@@ -208,7 +214,7 @@ pub mod digest {
             response: &Response,
         ) -> Option<String> {
             let challenge_string = response.header("www-authenticate")?;
-            let mut challenge = WwwAuthenticateHeader::from_str(&challenge_string).ok()?;
+            let mut challenge = WwwAuthenticateHeader::from_str(challenge_string).ok()?;
             let path = request.request_url().ok()?.path().to_string();
             let context = AuthContext::new(
                 self.username.as_ref(),
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -174,11 +174,13 @@ pub mod digest {
 
     /// Provides simple digest authentication powered by the `digest_auth` crate.
     ///
-    /// Any 401 response handled by this middleware is retried once with the
-    /// credentials provided on construction. If authentication fails or the middleware
-    /// is unable to generate an answer to the server challenge (such as a different authentication
-    /// scheme or a malformed challenge) the response is silently passed on to the rest
-    /// of the middleware chain.
+    /// Requests that receive a HTTP 401 response are retried once by this middleware with the
+    /// credentials provided on construction. The retry only happens under these conditions:
+    /// - there is no prior "authorization" header on the request set by the caller or other
+    ///   middleware, and
+    /// - the server provides HTTP Digest auth challenge in the "www-authenticate" header.
+    ///
+    /// In other cases, this middle ware acts as a no-op forwarder of requests and responses.
     ///
     /// ```
     /// let arbitrary_username = "MyUsername";
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -177,10 +177,10 @@ pub mod digest {
     /// Requests that receive a HTTP 401 response are retried once by this middleware with the
     /// credentials provided on construction. The retry only happens under these conditions:
     /// - there is no prior "authorization" header on the request set by the caller or other
-    ///   middleware, and
+    ///   middleware, and;
     /// - the server provides HTTP Digest auth challenge in the "www-authenticate" header.
     ///
-    /// In other cases, this middle ware acts as a no-op forwarder of requests and responses.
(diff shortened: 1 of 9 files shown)


strohel (approved):
For what it's worth, looks good to me!

Today I've learned that cases in `stc/test/*` [are called integration tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests), so calling external sites may be actually appropriate there.

src/middleware.rs
```diff
@@ -164,3 +164,87 @@ where
...
+    /// ```
+    pub struct DigestAuthMiddleware {
+        username: Cow<'static, str>,
+        password: Cow<'static, str>,
+    }
+
+    impl DigestAuthMiddleware {
+        pub fn new(
+            username: impl Into<Cow<'static, str>>,
+            password: impl Into<Cow<'static, str>>,
```
algesten, line 203: We have an informal policy of avoiding "complicated" type signatures, like those involving cows 🐮. In user facing PI, we prefer these to be just `&str`, even though it's a bit less efficient. It is however fine to have cows internally.

src/middleware.rs
```diff
@@ -164,3 +164,87 @@ where
...
+        }
+
+        fn construct_answer_to_challenge(
+            &self,
+            request: &Request,
+            response: &Response,
+        ) -> Option<String> {
+            let challenge_string = response.header("www-authenticate")?;
+            let mut challenge = WwwAuthenticateHeader::from_str(challenge_string).ok()?;
+            let path = request.request_url().ok()?.path().to_string();
```
algesten, line 218: Feels like we should be able to avoid `.to_string()` here given the below is a `Cow`

src/middleware.rs
```diff
@@ -164,3 +164,87 @@ where
...
+            username: impl Into<Cow<'static, str>>,
+            password: impl Into<Cow<'static, str>>,
+        ) -> Self {
+            Self {
+                username: username.into(),
+                password: password.into(),
+            }
+        }
+
+        fn construct_answer_to_challenge(
```
algesten, line 211: Maybe rename `respond_to_challenge`?

algesten (changes requested):
@PabloMansanet welcome to ureq!

This looks good, and I do welcome that we have some standard middleware in the main crate. I left a couple of notes in the code and here are some more.

* Please make changes `.github/workflows/test.yml` to ensure the new feature flag is being automatically tested.
* Similarly please update `test.sh` (this is what we use for local dev to roughly do what github actions do).
* We need to update `lib.rs` documentation for feature flags (and run `cargo readme` to generate a new README.md)

-----

Bigger picture discussion, maybe @jsha could input?

* Feature flags for middleware are good. No need to compile them if they are not wanted.
* These feature flags are different to the others we have because they require the user to both enable the feature flag _AND_ configure the agent with the middleware.
* Maybe we should have a separate section in README.md detailing how to use optional middleware?
* Since we're likely to have a couple more of these. I wonder if we should group them by some prefix. Like `mw-digest-auth`, `mw-basic-auth`

src/middleware.rs
```diff
@@ -164,3 +164,87 @@ where
         (self)(request, next)
     }
 }
+
+#[cfg(feature = "digest-auth")]
+pub mod digest {
```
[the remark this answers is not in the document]
PabloMansanet: Sounds good! Will get on that :)

src/middleware.rs
PabloMansanet: Originally these had `response` instead of `answer`, but there was a slightly annoying naming clash in the places where `ureq::Response` was being used, so I changed it to `answer` to remove the ambiguity. That said, I imagine `answer` is not a very correct or technical term... Maybe just renaming this function as you said (to `respond_to_challenge`) and leaving the `answer` variable untouched works?

PabloMansanet:
Thanks for the review @algesten. All the proposed changes make sense so I'll ping when I'm done with them :slightly_smiling_face:

src/middleware.rs
PabloMansanet: I'm not sure I can make Cows too useful with a &str input type as I need to satisfy the `'static` bound, but I'll replace Cows with Strings since this all seems like premature optimization anyway :)

PabloMansanet:
While testing this middleware on the field, we've noticed an issue that might be difficult to solve with the current Middleware API. 

Basically, while this works well for GET requests, anything with a payload is tricky because the middleware has no access to the payload from what I can see. Starting a request with `send_form` will send the payload without authorization, and then this middleware's recursive call to `send` will not have the payload available at all.

Is it conceivable to solve this kind of problem through a change in the `Middleware` API? I imagine the limitations around the current payload are due to `Payload::Reader`, but I wonder if, for all other `Payload` enum variants, it would be possible for middlewares to observe the payload.

There was a similar issue around 307/308 responses, and [the conclusion](https://github.com/algesten/ureq/issues/187) seems to be that PUT/POST/DELETE "retries" have to be handled at application level. Is this still the expectation, or can we possibly solve this at the middleware level for non-stream payloads?

algesten:
Hey! Thanks for fixing those things!

> There was a similar issue around 307/308 responses, and [the conclusion](https://github.com/algesten/ureq/issues/187) seems to be that PUT/POST/DELETE "retries" have to be handled at application level. Is this still the expectation, or can we possibly solve this at the middleware level for non-stream payloads?

As you correctly pointed out, this is very similar to handling 307/308 with body. I'm trying to think if there is some quick win here without handling streaming bodies. 

```rs
let response = next.handle(request.clone())?;
```

Here we would need something else, maybe something along the lines of:

```rs
// Buffer request bodies up to 10k
request.resend_buffer(10_000);

let response = next.handle(request)?;

...

if let Some((mut request, body)) = response.take_resend() {
    request.set("Authorization", "Bar");
    request.resend(body);
}
```

The idea is that `take_resend` will only return something if `resend_buffer` was set on the request and the request body is smaller than that buffer. `body` above would be some opaque type `ResendBody` that can only be used as an argument to `request.resend()`.

algesten:
Oh for the failed lint, do you mind running `cargo fmt`?

PabloMansanet:
Done! (running fmt, I mean). Thanks for the response @algesten . I see #473 is in the works, which will likely unblock me from working on this one, so I will wait for that to progress :)

algesten:

algesten:
Didn't mean to close this.

algesten:
Closing this since we're moving to 3.x

PR #469, opened 2022-01-24
example of using TLS provider to use mbedtls

mcr:
After many back and forth, I think that I've finally come up with a TlsConnector for MBEDTLS.
This was frustrated a bit because mbedtls is not Sync or Send, so it has to be wrapped in Mutex, and this presents some issues with the socket().   Even if I can figure out how to turn the context.io() return of Any back into TcpStream, it isn't clear to me how to return a reference to something inside the Mutex.
It may be that there are Arc and/or Mutex which is excessive.  Comments welcome.
I will be adding some code to dump the certificate chain in a new pull request.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ rustls-native-certs = { version = "0.6", optional = true }
 native-tls = { version = "0.2", optional = true }
 flate2 = { version = "1.0.22", optional = true }
 brotli-decompressor = { version = "2.3.2", optional = true }
+mbedtls = { version = "0.8.1", optional = true }
 
 [dev-dependencies]
 serde = { version = "1", features = ["derive"] }
@@ -53,6 +54,11 @@ rustls-pemfile = { version = "0.2" }
 
 [[example]]
 name = "smoke-test"
+
+[[example]]
+name = "mbedtls-req"
+required-features = ["mbedtls"]
+
 
 [[example]]
 name = "cureq"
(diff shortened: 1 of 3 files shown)


algesten (commented):
Hi @mcr!

Thanks for this example. Looks great!

I understand you need to wrap in a Mutex – I'm just looking into relaxing the `Sync` requirement on `TlsStream`, but that probably doesn't help you. We need `Send`, since ureq contains a connection pool, and we can't currently guarantee the connection will be reused from the pool on the same thread.

My only question is whether the `Oops` type is strictly necessary?

Also. You're getting a lint error due to formatting. Mind running `cargo fmt`?

Again thanks!

mcr:
> Hi @mcr!
> 
> Thanks for this example. Looks great!
> 
> I understand you need to wrap in a Mutex – I'm just looking into relaxing the `Sync` requirement on `TlsStream`, but that probably doesn't help you. We need `Send`, since ureq contains a connection pool, and we can't currently guarantee the connection will be reused from the pool on the same thread.

I think that we need to figure out how the lifetime of the things in the connection pool is dealt with.
Maybe I just need another level of Arc<>.    
I spent a bit of time trying to pass references into Response::do_from_stream(), (the todo comment in unit.rs), but that didn't work.

> My only question is whether the `Oops` type is strictly necessary?

Sorry, that's cargo culting https://www.urbandictionary.com/define.php?term=Cargo-Culting from smoke-test.

> Also. You're getting a lint error due to formatting. Mind running `cargo fmt`?

done.

algesten:
@mcr thanks for this! I made a follow-up PR with a quick cleanup https://github.com/algesten/ureq/pull/476

PR #471, opened 2022-01-29
Rename test function as_write_vec -> into_written_bytes

algesten:
Tests use `Response::as_write_vec` to inspect the outgoing HTTP/1.1
request line and headers. The current version has two problems:

1. Called `as_write_vec` when it actually returns a `&[u8]`.
2. Inspects/uses the `Response::stream` without consuming `Response`.

The first problem is trivial, but the second is subtle. Currently all
calls on `Response` that works with the internal `Response::stream`
consumes `self` (`into_string`, `into_reader`).

`Response` is by itself `Send + Sync`, and must be so because the
nested Stream is `Read + Write + Send + Sync`. However for
implementors of `TLSStream`, it would be nice to relax the `Sync`
requirement.

Assumption: If all fields in Response are `Sync` except
`Response::stream`, but any access to `stream` consumes `Response`, we
can consider the entire `Response` `Sync`.

This assumption can help us relax the `TlsStream` `Sync` requirement
in a later PR.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -520,8 +520,9 @@ impl Response {
     }
 
     #[cfg(test)]
-    pub fn as_write_vec(&self) -> &[u8] {
-        self.stream.as_write_vec()
+    pub fn into_written_bytes(self) -> Vec<u8> {
+        // Deliberately consume `self` so that any access to `self.stream` must be non-shared.
+        self.stream.written_bytes()
     }
 
     #[cfg(test)]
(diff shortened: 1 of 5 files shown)


jsha (approved):
Nice catch on this subtlety! Another thing I've been thinking: it's a shame this test-only method has to be on the TlsStream interface itself. It would be tidied if we could make some `TestStream` that impls TlsStream and also has an accessor that tests can use to get at the written data. Possible refinement for another time!

mcr (commented):
looks like the right thing.

PR #472, opened 2022-01-29
Fix rustls version spec

AlyoshaVasilieva:
`>=0.20.1` doesn't have an upper bound and will accept 0.21 (and higher) once they're available, but they'd presumably have breaking changes.
`0.20.1` means `^0.20.1`, which means `>=0.20.1, <0.21.0`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements


algesten:
Good point! Thanks!

PR #473, opened 2022-01-29
Resend body, 307/308 etc

algesten:


algesten:
@PabloMansanet I kinda hit a wall in here, and I'm not clear when I'll be able to resolve it.

The problem is that we don't want to take ownership of the body we are sending. This means we use this structure to handle different types of request body. 

```
pub(crate) enum Payload<'a> {
    Empty,
    Text(&'a str, String),
    Reader(Box<dyn Read + 'a>),
    Bytes(&'a [u8]),
}
```

The lifetime `'a` is only from calling `Request::send` (or another of the send methods) until we return the `Response`. To handle a resends, we need to have an `'a` that goes on for longer. We have conflicting requirements:

* Ability to inspect `Response` before doing a resend.
* Adding a lifetime to `Response` would be a breaking change (and be a quite clumsy API)

The solution is likely to not mix together `SizedReader<'a>` with `RewindReader`, but keeping them as separate things but used together. However I haven't quite thought out the path forward.

PabloMansanet:
Hey @algesten, thanks for the heads up! It's a pretty tricky problem...

I'm going to delve a bit deeper on the codebase to have a better view from where to propose a solution.

algesten:

PR #475, opened 2022-01-30
Remove Sync requirement of ReadWrite trait

### LINKED ISSUE #474: Relax Sync requirement of ReadWrite
Currently `ReadWrite` requires `Sync`.

```
pub trait ReadWrite: Read + Write + Send + Sync + 'static {
    fn socket(&self) -> Option<&TcpStream>;
}
```

Technically we don't need `Sync` because a socket connection will not be used outside a single thread while executing a request/response. The only reason it is `Sync` is because we want `Response`, `Error` etc to be `Sync`, and rust can't see that we're using `stream: Box<Stream>` in a "sync way".

This can be fixed to ease implementation of other TLS providers.

algesten:
3rd party TLS connections are currently required to provide a Sync
guarantee. Since ureq will only ever use the connection on a single
thread, this is not necessary.

The reason we end up with this bound is because we want Response and
Error to be Sync, in which case Rust's automatic inferral of Sync
fails.

This change "masks" the Stream in a wrapper making it Sync.

Close #474

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ socks = { version = "0.3", optional = true }
 serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
 encoding_rs = { version = "0.8", optional = true }
+sync_wrapper = { version = "0.1" } 
 cookie_store = { version = "0.15", optional = true, default-features = false, features = ["preserve_order"] }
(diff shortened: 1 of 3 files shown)


jsha (approved):
Looks good! What a neat solution.

Issue #478, opened 2022-02-04
webpki-roots is copy-left


emilk:
Thanks for a great library!

https://github.com/rustls/webpki-roots has a copy-left license (MPL-2.0) making it unsuitable for any closed-source use.

`webpki-roots` is enabled by the `tls` feature in `ureq (a default feature).

I suggest we make `rustls-native-certs` the default, and `webpki-roots` opt-in so that `ureq` does not create copy-left problems for its users unless they actively opt-in to using `webpki-roots`.

I can make a PR for this change.

---

PS: this was discovered using [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny)

jsha:
Thanks for spotting this!

I'm a little reluctant to change the default, in part because https://github.com/rustls/rustls-native-certs/issues/25 means that the platform verifier trusts some untrusted roots, which creates risk of accepting an inappropriate certificate.

Also, this would definitely merit a major version bump, because it could break connecting to some sites that would otherwise work. Still, overall the native-certs approach is the one we'd like to take long-term, assuming rustls adds the necessary extension points.

What do you think of waiting on this change until rustls has the necessary extension points to use the platform verifier _and_ platform roots, then doing it as a major version bump?

algesten:
I agree, @emilk we agree in principle to switch to `rustls-native-certs`, however now is not the time.

Would a way forward, for now, be to make a PR that documents the copyleft concerns?

emilk:
I understand this is a big change not to be taken lightly, and I trust your judgement on this.

I really enjoy the minimal dependencies of `ureq` (and not having to pull in an async executor), so I hope this can be resolved in the near future!

Until then I'll be putting this in my `Cargo.toml`:

``` toml
[patch.crates-io]
ureq = { git = "https://github.com/emilk/ureq/", branch = "opt-in-webpki-roots" }
```

Thanks for your attention and responses, and for a great crate!

emilk:
How about adding a new features to `ureq` in the meantime:

`tls_native_certs = ["webpki", "native-certs", "rustls"]`

This would allow users to use `tls` without any copy-left code being pulled in, by using

`ureq = { version = "2.4", default-features = false, features = ["gzip", "tls_native_certs"] }`

No major version bump required!

dralley:
[...]

PR #479, opened 2022-02-04
Make webpki-roots opt-in, and native-certs the default

emilk:
https://github.com/rustls/webpki-roots has a copy-left license (MPL-2.0) making it unsuitable for any closed-source use.

Therefore it is wiser if it is opt-in, rather than a default feature.

Closes https://github.com/algesten/ureq/issues/478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-# 2.4.0
+# UNRELEASED
+ * Make `native-certs` the default when using the `tls` feature, and make `webpki-roots` opt-in.
 
+# 2.4.0
  * Enable `gzip` feature by default (#455)
  * `gzip` and `brotli` feature flags to enable decompression (#453, #421)
  * Middleware function on agent (#448)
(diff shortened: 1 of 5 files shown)


jsha:
Thanks! Commented in #478.

dralley:
>https://github.com/rustls/webpki-roots has a copy-left license (MPL-2.0) making it unsuitable for any closed-source use.   It will not "infect" any larger work which includes MPLv2 licensed code.

This is not true.  MPLv2 is a "weak copyleft" license and it can be used in closed-source codebases without issue.  It will not "infect" any larger work which includes MPLv2 licensed code, which is different from GPL / LGPL style licenses.

See "Q11" from the FAQ [here](https://www.mozilla.org/en-US/MPL/2.0/FAQ/)

dralley:
Left a comment trying to provide some context [here](https://github.com/algesten/ureq/issues/478#issuecomment-1031029647)

emilk:
Closed in favor of the less-intrusive https://github.com/algesten/ureq/pull/482

PR #482, opened 2022-02-07
Add new feature "tls_native_certs"

emilk:
This adds the following feature:

`tls_native_certs = ["native-certs", "rustls"]`

This is an alternative to the `tls` feature, that uses `native-certs` instead of `webpki-roots`.

Closes https://github.com/algesten/ureq/issues/478

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ features = ["tls", "native-tls", "json", "charset", "cookies", "socks-proxy"]
 [features]
 default = ["tls", "gzip"]
 tls = ["webpki", "webpki-roots", "rustls"]
+tls_native_certs = ["native-certs", "rustls"]
 native-certs = ["rustls-native-certs"]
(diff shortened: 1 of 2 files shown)


algesten:
@emilk thanks for the PR!

Since I believe the copy-left problem in #478 is resolved. The question is – why do this? :)

Is it to reduce the number of dependencies?

emilk:
Well, I started on this before I realized #478 was resolved, so I thought I would finish the thought.

But yeah, it reduces dependencies for those users that want to use `native-certs` over `webpki` and `webpki-roots`.

Another alternative is to let users opt-in to `tls` and `webpki-roots` separately, which would require the same (minor) changes to the `.rs` files.

Anyway, since #478 is resolved I don't have a burning need for this, but I thought maybe others would :)

algesten:
Let's not complicate the feature flags at this point.

PR #490, opened 2022-03-15
Warn if `Request`s aren't sent

jyn514:
I added `ureq::get("http://example.com")` to a toy program and was very confused when it did nothing.
Change `Request` to give a warning if unsent:
```
warning: unused `ureq::Request` that must be used
  --> src/main.rs:48:5
   |
48 |     ureq::get("http://example.com");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: Requests do nothing until consumed by `call()`
```

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -24,6 +24,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// # }
 /// ```
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


jsha (approved):
Nice! Thanks for the fix.

Issue #481, opened 2022-02-06
Force HTTPS


insanitybit:
reqwest has an option to force HTTPS. This is particularly useful for redirects as it prevents HTTPS -> HTTP redirects.

https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.https_only

algesten:
I'd be open for such as configuration option on the agent level. @jsha what do you say?

algesten:
Closing since we're moving to ureq 3.x. This is solved in ureq 3.x

PR #492, opened 2022-04-04
implement https_only flag

pldubouilh:
hi - this should solve #481. I tested it against both `http` and `https -> http` redirect, and both cases are detected properly.
I'll add a test in a subsequent commit - as it needs to implement tls support from the test harness.
cheers !

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -336,6 +336,8 @@ pub enum ErrorKind {
     UnknownScheme,
     /// DNS lookup failed.
     Dns,
+    /// Insecure request attempted with https only set
+    InsecureRequestHttpsOnly,
     /// Connection to server failed.
     ConnectionFailed,
     /// Too many redirects.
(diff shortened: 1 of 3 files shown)


algesten (approved):
LGTM!

algesten:
If you fix the clippy lint, we can land this.

pldubouilh:
done - thank you :+1:

src/error.rs
```diff
@@ -336,6 +336,8 @@ pub enum ErrorKind {
     UnknownScheme,
     /// DNS lookup failed.
     Dns,
+    /// Insecure request attempted with https only set
+    InsecureRequestHttpsOnly,
```
upsuper, line 340: This is a breaking change according to https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#major-change-adding-new-variants so it should ideally have bumpped the major version based on semver. To prevent adding new variant becoming breaking change, the enum should be marked as [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).

upsuper: An example of downstream breakage: https://github.com/qiniu/rust-sdk/issues/10

PR #495, opened 2022-04-14
Add HTTPS proxy support

puffi:
This has a lot of interface changes, every change should be mentioned in the commit message.

Some things are still not clear to me, one is how to handle different TLS implementations when the connection to the proxy is made. We can't just reuse the TlsConnector for both the proxy connection as well as the connection to the target.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -267,7 +267,7 @@ impl Stream {
         Ok(())
     }
 
-    pub(crate) fn socket(&self) -> Option<&TcpStream> {
+    pub fn socket(&self) -> Option<&TcpStream> {
         self.inner.get_ref().socket()
     }
 
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -11,6 +11,8 @@ use chunked_transfer::Decoder as ChunkDecoder;
 #[cfg(feature = "socks-proxy")]
 use socks::{TargetAddr, ToTargetAddr};
 
+#[cfg(not(feature = "native-tls"))]
+use crate::default_tls_config;
 use crate::proxy::Proxy;
 use crate::{error::Error, proxy::Proto};
 
@@ -18,7 +20,7 @@ use crate::error::ErrorKind;
 use crate::unit::Unit;
 
 /// Trait for things implementing [std::io::Read] + [std::io::Write]. Used in [TlsConnector].
-pub trait ReadWrite: Read + Write + Send + 'static {
+pub trait ReadWrite: Read + Write + Send + Sync + 'static {
     fn socket(&self) -> Option<&TcpStream>;
 }
 
@@ -26,12 +28,12 @@ pub trait TlsConnector: Send + Sync {
     fn connect(
         &self,
         dns_name: &str,
-        tcp_stream: TcpStream,
+        tcp_stream: Stream,
     ) -> Result<Box<dyn ReadWrite>, crate::error::Error>;
 }
 
-pub(crate) struct Stream {
-    inner: BufReader<Box<dyn Inner + Send + 'static>>,
+pub struct Stream {
+    inner: BufReader<Box<dyn Inner + Send + Sync + 'static>>,
 }
 
 trait Inner: Read + Write {
@@ -188,7 +190,7 @@ impl fmt::Debug for Stream {
 }
 
 impl Stream {
-    fn new(t: impl Inner + Send + 'static) -> Stream {
+    fn new(t: impl Inner + Send + Sync + 'static) -> Stream {
         Stream::logged_create(Stream {
             inner: BufReader::new(Box::new(t)),
         })
@@ -328,7 +330,7 @@ pub(crate) fn connect_http(unit: &Unit, hostname: &str) -> Result<Stream, Error>
     //
     let port = unit.url.port().unwrap_or(80);
 
-    connect_host(unit, hostname, port).map(Stream::from_tcp_stream)
+    connect_host(unit, hostname, port)
 }
 
 pub(crate) fn connect_https(unit: &Unit, hostname: &str) -> Result<Stream, Error> {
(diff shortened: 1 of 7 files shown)


src/stream.rs
```diff
@@ -11,27 +11,29 @@ use chunked_transfer::Decoder as ChunkDecoder;
...
+use crate::default_tls_config;
 use crate::proxy::Proxy;
 use crate::{error::Error, proxy::Proto};
 
 use crate::error::ErrorKind;
 use crate::unit::Unit;
 
 /// Trait for things implementing [std::io::Read] + [std::io::Write]. Used in [TlsConnector].
-pub trait ReadWrite: Read + Write + Send + 'static {
+pub trait ReadWrite: Read + Write + Send + Sync + 'static {
```
algesten, line 23: This is a breaking change in our public API. Would required a major version bump.

src/stream.rs
```diff
@@ -11,27 +11,29 @@ use chunked_transfer::Decoder as ChunkDecoder;
...
     fn socket(&self) -> Option<&TcpStream>;
 }
 
 pub trait TlsConnector: Send + Sync {
     fn connect(
         &self,
         dns_name: &str,
-        tcp_stream: TcpStream,
+        tcp_stream: Stream,
```
algesten, line 31: This is also a breaking change in our public API.

algesten (commented):
Thanks for this PR!

Currently it makes changes that would break implementations for people using ureq. That means we would need to release these changes with a major version bump – and we're not keen to do that right now.

We can park this PR until we are ready to do a major bump. Alternatively if there's some way of preserving the current public API – that would be preferable.

src/stream.rs
puffi: This was required to get the .src() call working in case of an error.
std::io::Error requires 'Sync'.

Perhaps there's a different solution to this?

src/stream.rs
puffi: Yes, for HTTPS proxy support we require a TlsConnector (destination) wrapping a TlsConnector (proxy).
I couldn't think of a different solution that would be less intrusive and wouldn't require a public API change.

As you're more familiar with the code, does anything come to mind on how to solve this differently?

algesten:
@puffi I just had a realisation. Because the fix for #461 was never released, we are actually in a situation where no one can have implemented a TLS provider against 2.4.0. That means we can still change the API from TcpStream to Stream as you are proposing (it's obviously the better choice!).

I've pushed two commits to your PR. One that fixes some variable naming (tcp_stream  -> stream), and one that again removes the Sync trait bound for Stream.

src/stream.rs
```diff
@@ -426,8 +431,33 @@ pub(crate) fn connect_host(unit: &Unit, hostname: &str, port: u16) -> Result<Tcp
...
+            (Some(Proto::HTTPConnect), Some(ref proxy)) => {
+                Some((Stream::from_tcp_stream(stream), proxy))
+            }
+            _ => {
+                return Err(Error::new(
+                    ErrorKind::ProxyConnect,
+                    Some("No proxy defined, but proto set".into()),
+                ));
+            }
+        } {
```
algesten, line 460: To not lock us down to these two TLS implementations, this should use the TlsConnector configured in the Agent.

algesten: Done

algesten:
@jsha Do you have time to look at this?

HTTPS proxy support means we need to have a TLS stream inside a TLS stream. That means changing our public API so that `TlsConnector::connect` works over `Stream` instead of `TcpStream`. That's technically a breaking change.

However. Because of #461 no one has so far been able to implement a `TlsConnector` using our released version, which means I'm ok with changing that API. On a side note, I think it's more elegant `TlsConnector` works over the more generic `Stream`. What do you think?

This does mean `Stream` becomes part of our public API, and I'm not sure it's the good way to go. 

`TlsConnector::connect` trait fn requires a return of `Box<dyn ReadWrite>`, and it strikes me there's overlap between the `Stream` type and the `ReadWrite` trait. Why wouldn't we define the trait like this:

```rs
pub trait TlsConnector: Send + Sync {
    fn connect(
        &self,
        dns_name: &str,
        io: Box<dyn ReadWrite>,
    ) -> Result<Box<dyn ReadWrite>, crate::error::Error>;
}
```

I.e. keep `Stream` an internal concern, especially since `Stream::inner` has a `Box` anyway.

algesten:
#510 explores changing the `TlsConnector` trait in a way that would make this PR much simpler.

jsha:
I gave some feedback on #510, which seems like a good approach. Would you still like me to look at this implementation?

algesten:
@jsha I agree #510 is where it's at, and if we land that, this becomes a trivial implementation.

puffi:
Now that 2.5.0 is released and I got reminded about this pull request by a coworker, I'll work on this again.
Since lots of things have changed, would you recommend to close this one and open a new pull request?

algesten:
@puffi

Whatever is easier for you. I think we got all the history we need for the change either way.

puffi:
I'll open a new request once the code is ready.
But until then there's still one more question I have. Since the TlsConnector is part of the Agent and can only be used once, how can we know which TLS provider to use when creating the proxy connection? If we use the one from the agent, then it will fail in `connect_https`.
We at Proxmox use the `native-tls` feature, while the default one would be `rustls` (part of the default features). Then there's also the `mbedtls`.

puffi:
@algesten 

Or would it be an option to make the TlsConnector clonable to be able to get a new connector of the same provider? Currently that's not required as part of the trait, and as such not possible.

algesten:
@puffi not sure i follow here. I don't see how it "can only be used once". An agent can (per definition) issue many requests with the same connector.

The `TlsConnector` is stored away internally as this (notice `Clone`).

```rs
#[derive(Clone)]
pub(crate) struct TlsConfig(Arc<dyn TlsConnector>);
```

And `connect_https` doesn't even use the `Clone`, but accesses the connector via a shared (not even `mut`) reference:

```rs
    let tls_conf = &unit.agent.config.tls_config;
    let https_stream = tls_conf.connect(hostname, Box::new(sock))?;
```

Seems like the `TlsConnector` is highly reusable as it is. What am I missing?

puffi:
Closing this since I've opened a new one (#544).

PR #497, opened 2022-04-15
Fix Debug impl for AgentConfig

### LINKED ISSUE #496: Implement Debug for AgentConfig
`fmt::Debug` should be implemented properly for `AgentConfig`. Currently, there's just a `todo!()` inside the `fmt` function in the impl of `fmt::Debug` for `AgentConfig` around line 60-70 in `src/agent.rs`.

This caused a panic while printing the debug representation of `ureq::Agent` inside another panic for me. And panics inside panics aren't that very descriptive and should be avoided.

I know the problem is the field `tls_config: Arc<dyn TlsConnector>` which doesn't implement `Debug` and hence we cannot use `#[derive(Debug)]`. Perhaps the easiest solution is to just print `AgentConfig { ... }` as debug representation. It's at least better than a panic. Another alternative is to format the struct as usual and for the `tls_config` field we just write `tls_config: ...`.

jsha:
Wrap Arc<dyn TlsConnector> in a newtype with a stub Debug impl so we can
go back to `#[derive(Debug)]` on AgentConfig.

Fixes #496

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -711,16 +711,21 @@ mod tests {
     #[test]
     fn agent_config_debug() {
         let agent = AgentBuilder::new().build();
-        assert_eq!(
-            format!("{:?}", agent),
-            "Agent { config: AgentConfig { \
-            proxy: None, timeout_connect: Some(30s), timeout_read: None, \
-            timeout_write: None, timeout: None, https_only: false, \
-            no_delay: true, redirects: 5, redirect_auth_headers: Never, \
-            user_agent: \"ureq/2.4.0\", tls_config: TlsConfig \
-         }, state: AgentState { pool: ConnectionPool { \
-            max_idle: 100, max_idle_per_host: 1, connections: 0 \
-         }, resolver: ArcResolver(...), .. } }"
-        );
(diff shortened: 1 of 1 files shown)


src/agent.rs
```diff
@@ -695,4 +707,20 @@ mod tests {
...
+    #[test]
+    fn agent_config_debug() {
+        let agent = AgentBuilder::new().build();
+        assert_eq!(
+            format!("{:?}", agent),
+            "Agent { config: AgentConfig { \
+            proxy: None, timeout_connect: Some(30s), timeout_read: None, \
+            timeout_write: None, timeout: None, https_only: false, \
+            no_delay: true, redirects: 5, redirect_auth_headers: Never, \
+            user_agent: \"ureq/2.4.0\", tls_config: TlsConfig \
```
algesten, line 720: Will it not be tedious to lock down the version in a test like this?

algesten (approved):
LGTM! One question about the test.

src/agent.rs
jsha: Oops, good point. I will modify the test to not be so over-specified.

PR #500, opened 2022-04-21
Add Request::call_writer for writing request body using std::io::Write

ColonelThirtyTwo:
Request.call_writer() sends the request headers, and returns a RequestStream
object that can be used to write the request body. This is more convenient
than trying to create a Read object that does the same thing.

This also makes some major changes to the middleware trait, to support the
call_writer API. It's now split into `handle_request` and `handle_response`.

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -37,7 +37,7 @@ impl Default for Payload<'_> {
 /// The size of the body.
 ///
 /// *Internal API*
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub(crate) enum BodySize {
     Empty,
     Unknown,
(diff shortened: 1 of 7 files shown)


src/body.rs
```diff
@@ -97,6 +97,73 @@ const CHUNK_HEADER_MAX_SIZE: usize = 6; // four hex digits plus "\r\n"
...
+impl<W: Write> Write for ChunkWriter<W> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        debug_assert!(self.buffer.len() >= CHUNK_HEADER_MAX_SIZE);
+        let buf_remaining = CHUNK_MAX_SIZE.saturating_sub(self.buffer.len() + CHUNK_FOOTER_SIZE);
+
+        let num_write = buf.len().min(buf_remaining);
+        self.buffer.extend_from_slice(&buf[0..num_write]);
+
+        if self.buffer.len() >= CHUNK_MAX_SIZE - CHUNK_FOOTER_SIZE {
+            self.flush()?;
```
ColonelThirtyTwo, line 155: Should probably just be end_chunk, not a full flush

algesten:
Hello! Thank you for this PR!

Whilst in principle we're interested in a `std::io::Write` based API for sending the body content, this PR currently makes changes that we don't want to land.

* The Middleware changes are breaking, which would require a major version bump – this is something we don't want to do at this time.
* `call_writer` relies on a new `connect_inner_begin`, which effectively duplicates code from the read based code path. To ensure maintainability of the project, we need there to be only one code path for performing the requests regardless of how the body is provided.

Way forward:

Implementing this feature is hard. Personally I would take a step back and first try to refactor the `unit::connect` function so that it has only one code path regardless of whether it's a pull (read) or push (write) based call. The task is made harder by ensuring we (at some point) can implement [expect-continue](https://datatracker.ietf.org/doc/html/rfc2616#section-8.2.3) and 307/308 redirects that replay the body.

I think ideally the actual request should also be delayed until the first invocation of `write()` – this to avoid servers hanging up the socket between sending the headers and the body (many servers have timeouts to protect against [slowloris attacks](https://en.wikipedia.org/wiki/Slowloris_(computer_security))).

ColonelThirtyTwo:
I don't see how it would be possible to keep the current middleware API as it is while also having a write API. The middleware  API requires both the request and response information in the same call, but the Write API requires returning control to the caller to write the request.

I guess you could potentially run it in its own thread, and have the innermost handler block until the request is finished. Seems like a hack though - ureq would need its own thread pool.

The comment about `util::connect` is valid though. The big issue IIRC is supporting retries - it works for read bodies but won't work for write ones.

algesten:
I've been thinking about this the last few days, and as you say, there doesn't seem to be a way forward with the current middleware API.

We could _potentially_ consider some extension to the Middleware trait such as:

```rs
pub trait Middleware: Send + Sync + 'static {
    // required for pull based API (default)
    fn handle(&self, request: Request, next: MiddlewareNext) -> Result<Response, Error>;

    // required for extended push based API
    fn handle_push(&self, push_op: PushInput, next: MiddlewareNext) -> Result<PushOutput, Error> {
        panic!("Middleware::handle_push is not implemented, push based API doesn't work");
    }
}
```

This also hints at how I think it might be possible to rework the entire internal unit::connect.

```rs
// type already exists today
pub(crate) enum Payload<'a> {
    Empty,
    Text(&'a str, String),
    Reader(Box<dyn Read + 'a>),
    Bytes(&'a [u8]),
    Push(BodyChunk), // new thing
}

// Would be nice if it could have a slice, but unclear if we can make the lifetime work all the way.
struct BodyChunk(Vec<u8>);

enum PushInput {
    // Starts a request, Payload is either the entire read body, as per today, or the first BodyChunk for push.
    Begin(Request, Payload),
    // Additional writes after first.
    Chunk(BodyChunk),
    // No more chunks.
    End,
}

enum PushOutput {
    // When doing push based, indicates we need more writes.
    NeedMoreBody,
    // When push is complete, or we doing read based.
    Complete(Response),
}
```

algesten:

PR #503, opened 2022-04-24
Collapse "and and" in README

XrXr:


algesten:
Thanks! We use `cargo readme`, so I had to fix the lib.rs also.

PR #506, opened 2022-04-27
Use buffered reader when converting into json

RCasatta:
In my experience, with a 10Mb JSON, using `resp.into_json()` is taking more than twice the time than `serde_json::from_reader(BufReader::new(resp.into_reader()))` and I think there aren't big disadvantage to always buffer

UPDATE: performance degradation only happens when using timeout

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -436,7 +436,7 @@ impl Response {
         use crate::stream::io_err_timeout;
         use std::error::Error;
 
-        let reader = self.into_reader();
+        let reader = io::BufReader::new(self.into_reader());
         serde_json::from_reader(reader).map_err(|e| {
             // This is to unify TimedOut io::Error in the API.
             // We make a clone of the original error since serde_json::Error doesn't


jsha:
Interesting! I was slightly surprised this is such a performance improvement. We do use a `BufReader` internally:

```rust
pub(crate) struct Stream {
    inner: BufReader<Box<dyn Inner + Send + 'static>>,
}
```

However, we later wrap `Stream` in other readers, in particular `DeadlineStream`. If a deadline or timeout was set, `DeadlineStream` will perform a syscall (to set the read timeout) on every read _it_ receives. So if `serde_json` is doing a lot of small reads (which is totally reasonable for it to do), we would expect to see a lot of syscalls, which would slow things down.

Can you share your code that demonstrates the issue?

I'm not sure this pull request is the best way to implement a fix, assuming my diagnosis is correct. Instead, I think we probably want to change the `impl Read for DeadlineStream` so it checks `self.stream.buffer` first, and if there is anything available in the stream's buffer, it pulls as much as it can from that and returns. Then it would only do an actual read (with corresponding syscalls) when the stream's buffer is empty. https://github.com/algesten/ureq/blob/653cf03deed4ff2d5e723ce3916d9c4769e3f1df/src/stream.rs#L152-L165

RCasatta:
> Can you share your code that demonstrates the issue?

I confirm your diagnosis, it only happens with a timeout set.

This is a test that highlight the difference with the big json I have in my codebase
[index.zip](https://github.com/algesten/ureq/files/8580921/index.zip)

```

#[test]
#[cfg(feature = "json")]
fn big_body_as_json_no_buffer() {
    let now = std::time::Instant::now();

    test::set_handler("/big_body_as_json_no_buffer", |_unit| {
        test::make_response(
            200,
            "OK",
            vec![],
            include_str!("index.json").to_string().into_bytes(),
        )
    });
    let resp = get("test://host/big_body_as_json_no_buffer")
        .timeout(std::time::Duration::from_secs(30))
        .call()
        .unwrap();
    let _json: serde_json::Value = resp.into_json().unwrap();
    println!("big_body_as_json_no_buffer {:?}", now.elapsed());
}

#[test]
#[cfg(feature = "json")]
fn big_body_as_json_with_buffer() {
    use std::io::BufReader;
    let now = std::time::Instant::now();

    test::set_handler("/big_body_as_json_with_buffer", |_unit| {
        test::make_response(
            200,
            "OK",
            vec![],
            include_str!("index.json").to_string().into_bytes(),
        )
    });
    let resp = get("test://host/big_body_as_json_with_buffer")
        .timeout(std::time::Duration::from_secs(30))
        .call()
        .unwrap();
    let _json: serde_json::Value =
        serde_json::from_reader(BufReader::new(resp.into_reader())).unwrap();
    println!("big_body_as_json_with_buffer {:?}", now.elapsed());
}
```
> I'm not sure this pull request is the best way to implement a fix, assuming my diagnosis is correct. Instead, I think we probably want to change the impl Read for DeadlineStream so it checks self.stream.buffer first, and if there is anything available in the stream's buffer, it pulls as much as it can from that and returns. Then it would only do an actual read (with corresponding syscalls) when the stream's buffer is empty. 

I do agree but I am not familiar with the codebase enough to do it myself

jsha:
Closing in favor of #508

PR #508, opened 2022-04-28
DeadlineStream: read from Stream's buffer first

jsha:
An alternate fix for the problem reported in #506.


algesten (approved):
Nice!

lolgesten:
This looks great!

PR #510, opened 2022-04-30
Box<dyn ReadWrite> for TlsConnector::connect

algesten:
This PR explores the possibility of changing the signature of `TlsConnector::connect`.

Current situation

```rs
pub trait TlsConnector: Send + Sync {
    fn connect(
        &self,
        dns_name: &str,
        tcp_stream: TcpStream,
    ) -> Result<Box<dyn ReadWrite>, crate::error::Error>;
}
```

Proposed change in this PR:

```rs
pub trait TlsConnector: Send + Sync {
    fn connect(
        &self,
        dns_name: &str,
       io: Box<dyn ReadWrite>,
    ) -> Result<Box<dyn ReadWrite>, crate::error::Error>;
}
```

----------

## Motivation

#495 wants to add HTTPS proxy support (today we only have HTTP proxy). This means we would have one HTTPS connection to the proxy, and inside that we would tunnel the HTTPS traffic to the actual site being requested. That means we get a TLS stream inside a TLS stream.

The current `TlsConnector` works by wrapping `TcpStream` directly, which means we can at most have one layer of TLS. in #495 this is solved by publicly exposing the internal `Stream`.

This technically changes already released public API, however, because of bug #461 no one has been able to implement `TlsConnector` using our released versions, which is why I think we can allow ourselves to explore this space a bit more.

I'm keen on not exposing more types than we need, and `Stream` is not necessary. `Box<dyn ReadWrite>` is a more general alternative. The downside is that the current `TcpStream` is concrete, and so is `Stream` – a solution using `Box<dyn ReadWrite` directly, commits us to having an extra vtable lookup per read/write operation (`Stream` currently has a `Box<dyn>` internally, but we could potentially make that an enum).

### Future user provided transport

If we go down the path of `Box<dyn ReadWrite>`, one cool future possibility is to make ureq transport agnostic. We could let the user provide the transport in the agent. The API could look like this:

```rust
pub trait TransportProvider: Send + Sync {
    fn connect(
        &self,
        dns_name: &str,
    ) -> Result<Box<dyn ReadWrite>, crate::error::Error>;
}
```

`Agent::set_transport_provider(provider: impl TransportProvider)`

Our default implementation would provide a boxed up `TcpStream`, but a user implementation could let ureq work over any transport that is `Read + Write`.

### DIFF
--- a/examples/custom-tls.rs
+++ b/examples/custom-tls.rs
@@ -34,7 +34,7 @@ impl TlsConnector for PassThrough {
     fn connect(
         &self,
         _dns_name: &str,
-        io: Box<dyn ReadWrite + Sync>,
+        io: Box<dyn ReadWrite>,
     ) -> Result<Box<dyn ReadWrite>, Error> {
         if self.handshake_fail {
             let io_err = io::Error::new(io::ErrorKind::InvalidData, PassThroughError);
--- a/examples/custom-tls.rs
+++ b/examples/custom-tls.rs
@@ -31,21 +31,28 @@ struct PassThrough {
 }
 
 impl TlsConnector for PassThrough {
-    fn connect(&self, _dns_name: &str, tcp_stream: TcpStream) -> Result<Box<dyn ReadWrite>, Error> {
+    fn connect(
+        &self,
+        _dns_name: &str,
+        io: Box<dyn ReadWrite>,
+    ) -> Result<Box<dyn ReadWrite>, Error> {
         if self.handshake_fail {
             let io_err = io::Error::new(io::ErrorKind::InvalidData, PassThroughError);
             return Err(io_err.into());
         }
 
-        Ok(Box::new(CustomTlsStream(tcp_stream)))
+        Ok(Box::new(CustomTlsStream(io)))
     }
 }
 
-struct CustomTlsStream(TcpStream);
+struct CustomTlsStream(Box<dyn ReadWrite>);
 
 impl ReadWrite for CustomTlsStream {
     fn socket(&self) -> Option<&TcpStream> {
-        Some(&self.0)
+        self.0.socket()
+    }
+    fn is_poolable(&self) -> bool {
+        self.0.is_poolable()
     }
 }
 
--- a/examples/custom-tls.rs
+++ b/examples/custom-tls.rs
@@ -34,7 +34,7 @@ impl TlsConnector for PassThrough {
     fn connect(
         &self,
         _dns_name: &str,
-        io: Box<dyn ReadWrite>,
+        io: Box<dyn ReadWrite + Sync>,
     ) -> Result<Box<dyn ReadWrite>, Error> {
         if self.handshake_fail {
             let io_err = io::Error::new(io::ErrorKind::InvalidData, PassThroughError);
@@ -45,6 +45,7 @@ impl TlsConnector for PassThrough {
     }
 }
 
+#[derive(Debug)]
 struct CustomTlsStream(Box<dyn ReadWrite>);
 
 impl ReadWrite for CustomTlsStream {
--- a/examples/mbedtls/mbedtls_connector.rs
+++ b/examples/mbedtls/mbedtls_connector.rs
@@ -56,11 +56,10 @@ impl TlsConnector for MbedTlsConnector {
     fn connect(
         &self,
         _dns_name: &str,
-        io: Box<dyn ReadWrite>,
+        io: Box<dyn ReadWrite + Sync>,
     ) -> Result<Box<dyn ReadWrite>, Error> {
         let mut ctx = self.context.lock().unwrap();
-        let sync = SyncIo(Mutex::new(io));
-        match ctx.establish(sync, None) {
+        match ctx.establish(io, None) {
             Err(_) => {
                 let io_err = io::Error::new(io::ErrorKind::InvalidData, MbedTlsError);
                 return Err(io_err.into());
@@ -70,30 +69,14 @@ impl TlsConnector for MbedTlsConnector {
     }
 }
 
-/// Internal wrapper to make Box<dyn ReadWrite> implement Sync
-struct SyncIo(Mutex<Box<dyn ReadWrite>>);
-
-impl io::Read for SyncIo {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let mut lock = self.0.lock().unwrap();
-        lock.read(buf)
-    }
-}
-
-impl io::Write for SyncIo {
-    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        let mut lock = self.0.lock().unwrap();
-        lock.write(buf)
-    }
-
-    fn flush(&mut self) -> io::Result<()> {
-        let mut lock = self.0.lock().unwrap();
-        lock.flush()
-    }
-}
-
 struct MbedTlsStream {
     context: Arc<Mutex<Context>>, //tcp_stream: TcpStream,
+}
+
+impl fmt::Debug for MbedTlsStream {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MbedTlsStream").finish()
+    }
 }
 
 impl MbedTlsStream {
--- a/examples/mbedtls/mbedtls_connector.rs
+++ b/examples/mbedtls/mbedtls_connector.rs
@@ -53,15 +53,42 @@ impl MbedTlsConnector {
 }
 
 impl TlsConnector for MbedTlsConnector {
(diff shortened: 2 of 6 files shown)


jsha:
I like the idea, and I think the vtable lookup is totally fine. I also agree with your interpretation that we can break this API since the published version didn't actually work.

I see you're wrapping `ReadWrite` in a `Mutex` to make it `Sync`. But `TcpStream` is `Sync` itself. How about extending the bounds on `ReadWrite` to include `Sync`. so we don't need the `Mutex`? Then if someone wants to provide a non-`Sync` implementation they can wrap in a `Mutex` themself.

[Looking through the other TcpStream impls](https://doc.rust-lang.org/nightly/std/net/struct.TcpStream.html#trait-implementations), maybe we should also include `Debug`? It's nice to be able to log information about the TcpStream. Currently we do that via `to_socket()`. But if we're envisioning more abstraction, where the underlying transport might not be a `TcpStream`, it'd be useful to have a general `Debug` bound.

jsha:
Jon Gjengset's _Rust for Rustaceans_ has a great section on API design with some recommendations:

> #### Common Traits for Types
> Where possible, we should again avoid surprising the user and eagerly implement most of the standard traits even if we do not need them immediately.
...
> First among these standard traits is the Debug trait.
> Tied in close second are the Rust auto-traits Send and Sync (and, to a lesser extent, Unpin). If a type does not implement one of these traits, it should be for a very good reason. 
> The next set of nearly universal traits you should implement is Clone and Default.
> One step further down in the hierarchy of expected traits is the comparison traits: PartialEq, PartialOrd, Hash, Eq, and Ord.
> PartialOrd and Hash are more specialized, and may not apply quite as broadly, but where possible you will want to implement them too.
> Finally, for most types, it makes sense to implement the serde crate’s Serialize and Deserialize traits. 

Of those, we can't do anything after Clone. But perhaps we should include `Unpin`? It's not important to us, but someone may need to include one of these in some structure where `Unpin` is valuable. And we can't add trait bounds after release.

algesten:
That Mutex is in (your favorite) mbedtls example though ;) – I suspect they're not actually using the fact that it's Sync.

I was thinking that especially if we go down the user provided transport route, it's nice to not force sync of the transports the user provides. That TcpStream is Sync I find a bit amazing.

Totally agree on Debug trait.

Unpin is a bit well… are there any uses apart from async code?

jsha:
Unpin is only useful for async code, but sometimes it can be useful to use
ureq in async code. For instance, a colleague of mine had to use async
because he was using the AWS SDK package, but wanted to use ureq for the
requests he initiated outside of the AWS package. It's definitely odd to
require Unpin for us, but we can always remove bounds later, never add them.

Same goes for Sync. It might be overly restrictive but we can remove it
later.

Looking at Response, I see that it's currently Sync. When I'm back in front
of a computer I'm curious to check how that works without a Sync bound on
ReadWrite currently.

algesten:
I can maybe be swayed on Unpin.

> Same goes for Sync. It might be overly restrictive but we can remove it
later.

But, we don't need it currently! And I don't see how it could be useful for us. Why add bounds we don't and won't need?

I'm unsure why Response is Sync, we could look at that.

jsha:
I dug into why Response is Sync and realized we did this intentionally back in January (oops, I forgot!):

https://github.com/algesten/ureq/pull/475
https://github.com/algesten/ureq/issues/474

The goal was explicitly to not require Sync from TLS libraries. Looking at ones we know about:

- [rustls::Connection is Sync](https://docs.rs/rustls/latest/rustls/enum.Connection.html#impl-Sync).
- [openssl::ssl::SslStream is Sync](https://docs.rs/openssl/latest/openssl/ssl/struct.SslStream.html#impl-Sync).
- [mbedtls::ssl::Context is !Sync](https://docs.rs/mbedtls/latest/mbedtls/ssl/struct.Context.html#impl-Sync)

It makes sense that most TLS implementations are Sync. TLS is a full-duplex protocol, as is TCP, naturally. mbedtls seems to be the outlier in being !Sync. Looking at the code, I can't really tell what makes it so: https://docs.rs/mbedtls/latest/src/mbedtls/ssl/context.rs.html#70-90. According to https://tls.mbed.org/kb/development/thread-safety-and-multi-threading, "The default philosophy is that a single thread should only use or access one context at a same time, unless: The documentation for the functions that access the shared context explicitly states the function is thread-safe, or
You perform explicit locking yourself (perhaps in a wrapper function)." The mbedtls SSL [read](https://tls.mbed.org/api/ssl_8h.html#a5bbda87d484de82df730758b475f32e5) and [write](https://tls.mbed.org/api/ssl_8h.html#aa2c29eeb1deaf5ad9f01a7515006ede5) functions don't say anything about threading, so I guess mbedtls (the C library) is not full-duplex.

It's true we don't currently need a Sync bound, so maybe I'm being too fussy about it. There are two possible future things that we might want Sync for: pipelining (will never do, pipelining is dead), and h2 (big project, _probably_ won't happen but maybe?). Given the above finding that the mbedtls C library is really and truly not full-duplex, I am convinced: it's better to support the concrete present-day use case (support mbedtls) than preserve flexibility for some future abstract possibility. No Sync bound it is!

jsha:
Well okay, one more thing: After looking back at #474 and #475, I thought: `sync_wrapper` is such a tiny dependency, we should just take the parts of it we need and do them ourselves, to reduce our dependency tree. But then I realized `sync_wrapper` at its core does:

```
unsafe impl<T> Sync for SyncWrapper<T> {}
```

Pulling that into our own codebase would violate our own "no unsafe code" rule. And it feels a little like cheating to take such a tiny dependency whose main purpose is an `unsafe impl`.

Part of me wants to say "the one TLS library that's not full-duplex should use `SyncWrapper` itself." But I guess the issue is that `ureq` knows it never actually reads and writes to a `ReadWrite` concurrently, so only `ureq` can apply the `SyncWrapper` trick.

jsha:
In #511 I was surprised to find that I didn't need to do anything special to get MbedTlsStream to be `Sync`. But it turns out our example connector already uses a Mutex to accomplish that:

```rust
struct MbedTlsStream {
    context: Arc<Mutex<Context>>, //tcp_stream: TcpStream,
}
```

algesten:
I was thinking, there's a subtle difference between 1 `ReadWrite` that we provide as argument to `TlsConnector` and 2 the `ReadWrite` the connector returns back. 

1. would benefit from being sync as we see in mbedtls.
2. is good without, because we don't need it. 

Maybe we can do case 1 as `io: Box<dyn ReadWrite + Sync>` so we can have the optimal in both cases?

algesten:
But then… the asymmetry of `+ Sync` bound would be hard to uphold for the HTTPS proxy case. 🤔

jsha:
Yeah, I think having the bounds be symmetric is very nice.

Here's one way to think of it:

Advantage for having a `Sync` bound: Lets us get rid of a dependency (in particular one that is mainly oriented around a single `unsafe` use).

Disadvantage for having a `Sync` bound: It's harder to implement a TlsConnector for a backend that isn't Sync. But in this case "harder" just means "wrap it in a Mutex," as we've seen with the MbedTls connector. And that Mutex is quite cheap because in practice it's always uncontended. And for the two most important backends (openssl and rustls), no Mutex wrapper is needed because they are already Sync.

The disadvantage seems pretty small to me, and the advantage seems to align with a couple of our guiding principles for this crate (small dependency tree; avoid unsafe).

(Sorry for changing my mind somewhat from when I said "No Sync bound it is!" earlier)

algesten:
With regards to `Sync`, I think we need to have "two directions" in mind.

1) Types that ureq provides to the user, i.e. `Request`, `Respone`, `Agent`, `Error`. To maximize the utility of ureq, we should strive to have all these as `Sync`. That way the user can "stash away" an Agent or a Response and we would not "taint" the user's context with non-Sync.

2) Types we expect from the user, i.e. `impl TlsConnector`, `impl Resolver`, `impl Middleware`. For these, we should try to avoid `Sync` as much as reasonable, since that will make it easier for the user to integrate with ureq (we can go even further here in what traits we might be able to avoid, such as `Send`, `'static` etc).

For me these represent two different directions: 1 stuff we provide the user, 2 stuff we want from the user. And they have different requirements.

`Sync` for `Request` is in category 1, and I agree we can replace that `SyncWrapper` with a `Mutex`, since there is no contention. However we can't drop it completely, because we need `Request` to stay `Sync`.

`Sync` for `ReadWrite` is different. Up until this this PR it was purely category 2, but now it's getting mixed. If we imagine providing a future "use your own transport" function, we have another case where it's category 2. And I think it's correct for us to _NOT_ require `Sync` from the user if we can avoid it.

We will never have unsafe use directly in ureq, but crates we depend upon will always have it. `sync_wrapper` is entirely sound (I think it could be included in Rust std lib). This is the current situation with geiger.

image

algesten:
Merging this since I believe it's generally correct. Happy to have further discussions about it before we release 2.5.0.

PR #511, opened 2022-04-30
Add Debug and Sync to ReadWrite bounds

jsha:
And add Debug impl to RustlsStream and CustomTlsStream.

Example of what I proposed in #510


jsha:
The Debug impl was added in #510, and we're not adding the `Sync` bound.

jsha:
The Debug impl was added in #510, and we're not adding the `Sync` bound.

PR #513, opened 2022-05-07
Store response in vec for short responses

### LINKED ISSUE #264: Pool stream early in do_from_read when whole response is in buffer
In Response::do_from_read, if a response has Content-Length, and that Content-Length is less than the length of `stream.inner.buffer()`, we should read those bytes immediately into a Vec, place the Vec inside Response (as an alternative to Stream), and return the stream to the pool. That will make the stream available as soon as possible for reuse. More importantly, it means that a lot of Status errors will be poolable where they otherwise wouldn't be (because the caller returns before reading the body).

### LINKED ISSUE #505: Get content from reference to response
This possibly ties into #126.

mladedav:
closes #505 
closes #264 

When receiving a response internal buffers get filled right away. If the fetched size is the same as declared content length then the data is copied and the stream is released.

Somewhat related although as I found out not as much is an added API to access the internal buffer so that at least part (in effect 8 kB) of the response can be obtained without consuming the response.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -282,7 +282,7 @@ impl AgentBuilder {
                     self.max_idle_connections_per_host,
                 ),
                 #[cfg(feature = "cookies")]
-                cookie_tin: CookieTin::new(self.cookie_store.unwrap_or_else(CookieStore::default)),
+                cookie_tin: CookieTin::new(self.cookie_store.unwrap_or_default()),
                 resolver: self.resolver,
                 middleware: self.middleware,
             }),
--- a/src/response.rs
+++ b/src/response.rs
@@ -287,7 +287,7 @@ impl Response {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn into_reader(self) -> Box<dyn Read + Send> {
+    pub fn into_reader(self) -> impl Read + Send {
         let is_http10 = self.http_version().eq_ignore_ascii_case("HTTP/1.0");
         let is_close = self
             .header("connection")
@@ -938,7 +938,6 @@ mod tests {
             .unwrap();
         let stream = res.stream.get_mut();
         drop(stream);
-        println!("Prefetched {}", res.partial_body().len());
         let stream = res.stream.get_mut();
         assert!(matches!(stream, ResponseStream::Buffer(_)));
     }
@@ -953,7 +952,6 @@ mod tests {
             .unwrap();
         let stream = res.stream.get_mut();
         drop(stream);
-        println!("Prefetched {}", res.partial_body().len());
         let stream = res.stream.get_mut();
         assert!(matches!(stream, ResponseStream::Stream(_)));
     }
--- a/src/response.rs
+++ b/src/response.rs
@@ -1,4 +1,4 @@
-use std::io::{self, Read};
+use std::io::{self, Cursor, Read};
 use std::str::FromStr;
 use std::{fmt, io::BufRead};
 
@@ -33,6 +33,11 @@ const INTO_STRING_LIMIT: usize = 10 * 1_024 * 1_024;
 // https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html
 const MAX_HEADER_SIZE: usize = 100 * 1_024;
 const MAX_HEADER_COUNT: usize = 100;
+
+enum ResponseStream {
+    Stream(Box<Stream>),
+    Buffer(Vec<u8>),
+}
 
 /// Response instances are created as results of firing off requests.
 ///
@@ -69,7 +74,7 @@ pub struct Response {
     // Boxed to avoid taking up too much size.
     unit: Box<Unit>,
     // Boxed to avoid taking up too much size.
-    stream: SyncWrapper<Box<Stream>>,
+    stream: SyncWrapper<ResponseStream>,
     /// The redirect history of this response, if any. The history starts with
     /// the first response received and ends with the response immediately
     /// previous to this one.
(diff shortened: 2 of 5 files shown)


mladedav:
Sorry I didn't check git history. It's rebased now so it should run

algesten:
Sorry, I just now merged #510, because I realised it will have a bearing on this effort. Or at least I believe it will.

This change:

```rs
    pub fn into_reader(self) -> impl Read + Send {
```

to

```rs
    pub fn into_reader(self) -> Box<dyn Read + Send> {
```

is a breaking API change, and would require a bump of major version – something we don't want to do at this time.

However ultimately I think this API change arises from a slightly wrong way to tackle this problem. I believe the buffering in this PR currently happens "too high up" – it's implemented in `Request` where I think we want it somewhere in `Stream`.

# Idea

The stream inside `Request::stream` is currently used like so (loose naming):

`Deadline(Stream(BufReader(Tls(TcpStream))))`

* `DeadlineStream` enforces timeouts and implements `BufRead`, and delegates the bufreader stuff to inner.
* `Stream` implements `BufRead` and delegates to a wrapped `BufReader` over a `Box<dyn ReadWrite>`, which is either a TLS implementation on a "raw" socket.
* `Tls` is either a rustls, native-tls or user provided TLS wrapper. It might not be here if we're doing `http`.
* `TcpStream` is the raw socket.

`Stream` is the thing that goes back to the pool of reusable connections, which means we don't make this change in `Stream`. I don't think there's any particular reason our `BufReader` today sits inside `Stream` (I could be wrong!) – which means we can "yank" the `BufReader` out and make a new layer that we call `Buffered` like so:

`Deadline(Buffered(Stream(Tls(TcpStream))))`

`Buffered` would be an enum:

```rs
pub enum Buffered {
    Fully(Cursor<Vec<u8>>),
    Reader(BufReader<Stream>),
}
```

`Buffered` would implement `BufRead` and behave towards `Deadline` the way `Stream` does today.

This means `Response` would see no direct changes from this. However, the type inside `Request::stream` would change to `SyncWrapper<Buffered>`. Which in turn means we can expose new functions on `Response` to work with this buffer.

* `is_buffered_fully() -> bool`
* `try_buffer_fully(max: usize) -> io::Result<bool>`

etc...

mladedav:
Sorry for the sneaky API breakage, I have completely forgotten I changed that and didn't catch it when adding the changes to git.

That said this change is not necessary because `Box<dyn Read + Send>` implements `Read + Send`. So if the only issue is the signature and not the actual returned type this is not a breaking change (I have changed it back).

The way I saw it is that the buffered output is above the `DeadlineStream` object because there are no more deadlines involved and it is a stream just so that it can be used uniformly with the `into_reader` method.

What I have now in `Response` is basically `Buffered(Deadline(Stream(BufRead(Tls(TcpStream))))` except I have named it differently (which was just lack of imagination on my part).

If you want to have it changed so that the response itself does not need to know about the underlying implementation I can do that no problem. Just the original underlying problem seems to not be there so I don't know if it's needed.

jsha:
> Stream is the thing that goes back to the pool of reusable connections, which means we don't make this change in Stream. I don't think there's any particular reason our BufReader today sits inside Stream (I could be wrong!) – which means we can "yank" the BufReader out and make a new layer that we call Buffered like so:

I moved it from `BufReader(Stream(TcpStream))` to `Stream(BufReader(TcpStream))` in https://github.com/algesten/ureq/pull/261. The description on that pull is that I wanted to rely on `Stream` impl'ing `BufRead`, for instance in read_next_line.

If we move `BufReader` back to the other side of `Stream`, assuming that `Stream` is still the unit of pooling, we'll have to wrap and unwrap a Stream for each request. We have to be particularly careful when unwrapping: [`BufReader::into_inner` drops data on the floor.](https://doc.rust-lang.org/std/io/struct.BufReader.html#method.into_inner). That might be a concern, for instance, if a server sends us more bytes than promised. We should detect that and drop the connection. That should be possible to do by checking if `stream.buffer()` is empty at pooling time.

> This means `Response` would see no direct changes from this. However, the type inside `Request::stream` would change to SyncWrapper<Buffered>. Which in turn means we can expose new functions on Response to work with this buffer.

I think `Request` was a typo here - did you mean `Response`?

I'm not seeing the philosophical difference between what you propose, @algesten, vs what's in the PR. You'd like `Response.stream` to be an enum (stored bytes OR a wrapped TcpStream), but this PR already makes it an enum.

> However ultimately I think this API change arises from a slightly wrong way to tackle this problem. I believe the buffering in this PR currently happens "too high up" – it's implemented in Request where I think we want it somewhere in Stream.

(s/Request/Response again?)

I think `Response` is just the right place for it. `Stream` represents "a wrapped TcpStream, which we might give back to the pool." It doesn't make sense to give back to the pool a `Vec<u8>`. But it does make sense to say "This is a `Response` that has already given back its `Stream`. All reads from it will read from an internal `Vec<u8>`".

mladedav:
Hi @algesten, does this look ok for you with the clarifications? Should I change anything?

algesten:
@mladedav @jsha I think the changes in this PR are too extensive. I could be convinced we can do the buffering on the "Response level", but we need to abstract over the enum to make less changes to Response. I especially dislike any matching on the new enum to follow different code paths.

The main hurdle to get around is that `DeadlineStream` is directly wrapping `Stream`.

mladedav:
@algesten I'm sorry to say that I don't know how to achieve not matching on the implementation. I assume you don't want the code in `Response` to match on the implementation and delegate that branching into a method inside a method in the `enum`, correct?

The problem I see is that the `into_reader` method needs a lot of information from the `Response`. But if getting that information should be delegated to a method in the enum, then it needs the `Response` object, but it has been partially moved because we need first to take the stream (or we could just borrow the stream, create a dummy value and then use that as a discriminator but that feels like an antipattern).

Another option would be to take the stream implementation and put a dummy value in its place, which I don't like much either.

Any pointers on how to make the changes with better code would be greatly appreciated.

mladedav:
Hi @algesten. I was just wondering if this is something that could still be worked on. As I tried to explain in the last comment, I don't see a feasible technical solution to extracting the relevant parts outside the Response struct.

I would appreciate any guidance you could provide in tackling this issue.

Thank you

algesten:
Hi @mladedav. Sorry for the delay on this. Since I struggled to express what I wanted, I decided to spend some time today implement this PR in a different way. See #526.

What I want to achieve is less diverging code paths and changes to the public API.

algesten:
Close in favor of #531

lolgesten:
Yeah. My attempt wasn't too hot either. Jacob came to save us all from ourselves.  :)

PR #514, opened 2022-05-08
Remove sync_wrapper dep in favor of Mutex

jsha:
We unwrap the stream exactly once per response, and we know that case will be uncontended for the same reason `SyncWrapper` works: `into_reader()` takes `self`, so it must have exclusive ownership. Uncontended mutexes are extremely cheap. This saves us a dependency at a trivial performance cost.

Followup to discussion in https://github.com/algesten/ureq/pull/510#issuecomment-1114156889.


algesten (approved):
💯

PR #517, opened 2022-05-18
Return Err on deadline checked_add overflow instead of panic

guyo13:
Previous call to `unwrap()` on the checked_add return value panicked when it was None.
This change returns an Err result instead of panicking.


algesten:
This looks good. Thanks!

Can you run `cargo fmt` and push again?

guyo13:
Done!

algesten:
Thanks!

Issue #419, opened 2021-09-18
Request via specified NIC


benydc:
Is it possible to bind a request to a specific network interface?

algesten:
To solve this completely, might be beyond `ureq` capabilities right now. ureq relies on Rust's `std::net` to [connect a socket to the remote host](https://doc.rust-lang.org/1.55.0/std/net/struct.TcpStream.html#method.connect).

It implicitly does a DNS lookup on `addr` (which comes from the request URL). That lookup will result in some IP address, and then it would go via the local routing table for getting to that IP. There is no direct control over which NIC is being used.

This would create an issue for multihomed computers. I.e. if the routing table is such that there are multiple ways of reaching the same IP and that (what I assume is `libc`), does not resolve primarily to the route/NIC wanted.

Another situation would be that the URL can resolve to multiple IP addresses, and it ends up using "the wrong one" (split DNS scenario or similar), but it if selected another IP it would implicitly also select the correct NIC. There is some control that can be had over DNS lookup, check out [this example](https://github.com/algesten/ureq/blob/main/examples/ipv6.rs).

jsha:
I agree this is beyond ureq's capabilities. A sketch of how you might achieve something:

First, figure out which IP addresses are bound to which NIC. At the TCP layer we don't have the ability to say "use this NIC," but we do (in a platform-specific way) have the ability to say "use this source IP," and that _may_ amount to the ability to choose a NIC, depending on your setup.

Read this, about how to use a specific source IP: https://idea.popcount.org/2014-04-03-bind-before-connect/

You would then need to find a way of creating a TCP connection in Rust that is lower-level than net::TcpStream. On a quick Google, maybe https://github.com/libpnet/libpnet ? You'd use this method to set the source IP of a socket _before_ you dial it.

You would then need some way to package that up and hand it off to ureq. Right now that doesn't exist. I could imagine adding an integration point where the user could supply a Dialer that takes hostnames or IP addresses and produces a `Read + Write` or similar. However, this is pretty far down the line - first you would need to figure out how to dial from a specific source IP in Rust.

benydc:
@algesten @jsha thank you!
In the end I'm going to hand roll the client, using `net2` crate with TcpBuilder, or with `libpnet` crate or whatever I can find on google. This issue can now be closed, thank you!

Issue #301, opened 2021-01-14
feature: Custom connect function, like AgentBuilder.resolver


tv42:
Hi. I have a low-level concern where I want to perform setsockopt calls between socket and ~bind~connect syscalls, specifically `SO_MARK`. Would you be open to a pull request that adds an AgentBuilder.connect or some such method, that takes something like say

```
pub trait Connector: Send + Sync {
    pub fn connect(&self, addrs: Vec<SocketAddr>, timeout: Option<Duration>) -> IoResult<TcpStream>;
}
```

Feedback would be most welcome. I think I can write the code pretty easily, but wanted to reach out first.

tv42:
Also, is there a better type than `TcpStream`? Consider in-memory test implementations, UNIX domain sockets (`curl --unix-socket`), etc. All I could quickly find was https://crates.io/crates/multisock , is there no `Stream` in `std`?

jsha:
I think the overall idea sounds good. I'll have to think a little on the exact API. Do we want to take a `Vec<SocketAddr>` and let the connector decide about retries / internally? Or take a `SocketAddr` and have `ureq` do retries? I think `Vec<SocketAddr>` is probably right.

I think this trait should probably take a deadline rather than a timeout, since that's what ureq is using internally. Otherwise ureq would need to transform from a deadline back to a timeout.

> Also, is there a better type than TcpStream?

It would be nice to be more generic. For instance, rustls::StreamOwned only requires Read + Write: https://docs.rs/rustls/0.19.0/rustls/struct.StreamOwned.html. But we have some internal stuff that depends on `TcpStream` to set and clear timeouts. I don't think we can be generic and still have the ability to call those (unless we get into dynamic dispatch).

algesten:
The dream would be to be generic over Read/Write. I wonder if this proposed `Connector ` trait could be expanded so that also our internal configurations can be dealt with using an internal implementation of the trait.

Something like a `configure()` (or possibly even more functions) that we use from selected places where we need to delegate the handling of the underlying socket.

```
pub trait Connector: Send + Sync {
    pub fn connect(&self, addrs: Vec<SocketAddr>, deadline: Option<Instant>) -> io::Result<Box<io::Read + io::Write>>;
  pub fn configure(&self, socket: Box<io::Read + io::Write>, deadline: Option<Instant>) -> io::Result<()>;
}

tv42:
[...]

Issue #251, opened 2020-11-23
IPv6 Bind


GoldLine1337:
Hello, I'm trying to bind a IPV6 address as ip source for an exampe to facebook.com

can you give me an example of the code?

ycd:
I'd also love to have an example of this.

algesten:
ureq connects to whatever socket address it gets [here](https://github.com/algesten/ureq/blob/ec69c4282c08dbbb22b86a9247d068c48da0abaa/src/stream.rs#L380-L383)

The default std resolver, will figure out the IP address for the host name [using rust std lib](https://github.com/algesten/ureq/blob/ec69c4282c08dbbb22b86a9247d068c48da0abaa/src/resolve.rs#L14-L18).

The `SocketAddr` returned from the resolver can be a mix of ipv4 and ipv6 and ureq doesn't care.

I've made some example code of how to write a bespoke resolver in #337

ycd:
That's awesome @algesten, thank you!

PR #518, opened 2022-05-19
feat: Add Connector trait

zu1k:
I need to get a tcpStream established from a specific source IP, but I didn't find a proper `bind_connect` implementation in `ureq`.

I checked the historical issue and found a lot of discussions on related content. #301 #419 #251

At the beginning, I added the `bind_connect` implementation to `ureq`, but I quickly relized that `ureq` needs to be kept lightweight, and there may be similar requirements in the future, such as establishing connections from socks files, or other special requirements, so finally I abstracted a most basic implementation of the `Connector` trait, and implemented `BindConnector` based on this.

The definition of the `Connector` trait is the same as the `connect` and `connect_timeout` of std `TcpStream`. I think it may need to be improved according to the needs in the future, but it should be enough for now.

Signed-off-by: zu1k <i@zu1k.com>

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -1,0 +1,47 @@
+use std::fmt;
+use std::io::Result as IoResult;
+use std::net::{SocketAddr, TcpStream};
+use std::sync::Arc;
+use std::time::Duration;
+
+/// A custom Connector to override the default TcpStream connector.
+pub trait Connector: Send + Sync {
+    fn connect(&self, addr: &SocketAddr) -> IoResult<TcpStream> {
+        TcpStream::connect(addr)
+    }
+
+    fn connect_timeout(&self, addr: &SocketAddr, timeout: Duration) -> IoResult<TcpStream> {
+        TcpStream::connect_timeout(addr, timeout)
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct StdTcpConnector;
+
+impl Connector for StdTcpConnector {}
+
+#[derive(Clone)]
+pub(crate) struct ArcConnector(Arc<dyn Connector>);
+
+impl<R> From<R> for ArcConnector
+where
+    R: Connector + 'static,
+{
+    fn from(r: R) -> Self {
+        Self(Arc::new(r))
+    }
+}
+
+impl fmt::Debug for ArcConnector {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ArcConnector(...)")
+    }
+}
+
+impl std::ops::Deref for ArcConnector {
+    type Target = dyn Connector;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref()
+    }
+}
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 /// A custom Connector to override the default TcpStream connector.
-pub trait Connector: Send + Sync {
+pub trait TcpConnector: Send + Sync {
     fn connect(&self, addr: &SocketAddr) -> IoResult<TcpStream> {
         TcpStream::connect(addr)
     }
@@ -18,28 +18,28 @@ pub trait Connector: Send + Sync {
 #[derive(Debug)]
 pub(crate) struct StdTcpConnector;
 
-impl Connector for StdTcpConnector {}
+impl TcpConnector for StdTcpConnector {}
 
 #[derive(Clone)]
-pub(crate) struct ArcConnector(Arc<dyn Connector>);
+pub(crate) struct ArcTcpConnector(Arc<dyn TcpConnector>);
 
-impl<R> From<R> for ArcConnector
+impl<R> From<R> for ArcTcpConnector
 where
-    R: Connector + 'static,
+    R: TcpConnector + 'static,
 {
     fn from(r: R) -> Self {
         Self(Arc::new(r))
     }
 }
 
-impl fmt::Debug for ArcConnector {
+impl fmt::Debug for ArcTcpConnector {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "ArcConnector(...)")
+        write!(f, "ArcTcpConnector(...)")
     }
 }
 
-impl std::ops::Deref for ArcConnector {
-    type Target = dyn Connector;
+impl std::ops::Deref for ArcTcpConnector {
+    type Target = dyn TcpConnector;
 
     fn deref(&self) -> &Self::Target {
         self.0.as_ref()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -663,6 +663,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "socket2"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b5fac59a5cb5dd637972e5fca70daf0523c9067fcdc4842f053dae04a18f8e9"
+dependencies = [
+ "libc",
+ "windows-sys",
+]
+
+[[package]]
 name = "socks"
 version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


mcr (commented):
I am concerned about how this might interact with TLS.
Maybe the example could do both TLS and connect()?

zu1k:
> I am concerned about how this might interact with TLS.
> Maybe the example could do both TLS and connect()?

Sorry I didn't understand you. Do you mean that the `Connector` also should do the work of the TLS handshake and stripping? Or maybe I should rename it to `TcpConnector` to avoid confusion with TlsConnector?

I don't think it's a good idea to abstract `TcpConnector` and `TlsConnector` into one `Connector`, for this makes external use of `ureq` more complicated.

src/connect.rs
```diff
@@ -0,0 +1,47 @@
...
+/// A custom Connector to override the default TcpStream connector.
+pub trait Connector: Send + Sync {
+    fn connect(&self, addr: &SocketAddr) -> IoResult<TcpStream> {
+        TcpStream::connect(addr)
+    }
+
+    fn connect_timeout(&self, addr: &SocketAddr, timeout: Duration) -> IoResult<TcpStream> {
+        TcpStream::connect_timeout(addr, timeout)
+    }
+}
```
algesten, line 16: I think this should tie into the `ReadWrite` trait. https://github.com/algesten/ureq/blob/4bb6d3a4dbd7acafed801dae9d75df02207e1b87/src/stream.rs#L21

I.e. the connector trait should not produce a concrete `std::net::TcpStream`, but rather something that implements `ReadWrite`. I haven't followed through all the trait bound implications, but the starting point would be:

```rs
pub trait Connector: Send + fmt::Debug + 'static {
    type Socket;

    fn connect(&self, addr: &SocketAddr) -> io::Result<Self::Socket>;
    fn connect_timeout(&self, addr: &SocketAddr, timeout: Duration) -> io::Result<Self::Socket>;
    fn set_read_timeout(&self, socket: &mut Self::Socket);
}
```

and the agent would have something like:

```rs
    pub connector: Arc<dyn Connector<Socket = dyn ReadWrite>>,
```

Notice this abstraction would require use redo today's handling of `set_read_timeout` relying on `ReadWrite::socket()`. Unclear whether we should just go `set_read_timeout` straight on the `ReadWrite` trait.

Also minor detail, style wise we use `io::Error` instead of `IoError`.

algesten (changes requested):
Overall I'm really positive to this abstraction, however, I want to tie it to the `ReadWrite` trait instead of working with `std::net::TcpStream` directly.

That potentially makes the task a lot harder, because it has wider ramifications for the handling of `set_read_timeout` as well as some potentially tricky trait bounds to work out.

src/connect.rs
zu1k: `ureq` relies on the functions of `TcpStream` everywhere.
If we want to abstract it into a trait, we need to cover almost all of `TcpSteam`'s methods, I think this introduces too much complexity since `TcpStream` is already an abstraction for multiple platforms.

It looks like `hyper` adds a layer to be compatible with tcp-based h2 and quic-based h3, I don't think there is such a trait that can handle subsequent changes. I think abstracting away `TcpStream` should not be considered if not for the time being.

algesten: > If we want to abstract it into a trait, we need to cover almost all of `TcpSteam`'s methods

Really? In my mind it's pretty much only `set_read_timeout` – but maybe I'm not thinking this through?

zu1k: and `set_write_timeout`, `set_nonblocking`, `set_nodelay`

algesten: * `set_nodelay` can be set as soon as the socket has connected, no need to abstract over it (pass it as an argument to connect).
* `set_write_timeout` and `set_read_timeout` can probably be one single thing.
* `set_nonblocking` is used in a very specific way that can be abstracted.

I think we'd end up with something like:

```rs
pub trait Connector: Send + fmt::Debug + 'static {
    type Socket;

    fn connect(&self, addr: &SocketAddr, timeout: Option<Duration>, nodelay: bool) -> io::Result<Self::Socket>;
    fn set_read_write_timeout(&self, socket: &mut Self::Socket) -> io::Result<()>;
    fn check_server_closed(&self, socket: &mut Self::Socket) -> io::Result<()>;
}
```

Sure. It's a bigger change working this into the entire code base. Before embarking on it, maybe we should broaden the discussion.

@jsha what are your thoughts on this?

zu1k: Agree, this requires more discussion.

amunra:
I've been redirected here from #692. My requirement is pretty simple: Set the local bind interface address before establishing the connection, something that [`std::net::TcpStream`](https://doc.rust-lang.org/std/net/struct.TcpStream.html) can't do, thus it would be nice to allow the `ureq` user to do this by setting a connector on the agent.

It's probably a good idea to start summarising everything and document it in one go before jumping to a propoal.

Reading through the comments above there seem to be two candidate approaches with different tradeoffs and a set of desired features.

I'll start with the desired features (aka constraints):
* `(C1)` No breaking API changes.
* `(C2)` The ability to construct a custom `TcpStream`, for example with `.bind()` calls or custom TCP flags.
* `(C3)` Is composable with `ureq::stream::TlsConnector`. 
* `(C4)` Is compatible with `ureq::stream::Stream`:
    * `Stream` exposes `pub(crate) remote_addr: SocketAddr,`
* `(C5)` Is compatible with `ureq::stream::ReadWrite`:
    * `ReadWrite` exposes `fn socket(&self) -> Option<&TcpStream>`
* `(C6)` The ability to more generically connect a URL to something that can read and write, thus supporting other protocols such as unix domain sockets.

So far, there are two proposed approaches with partial coverage of these constraints:
* `(A1)` support a new tcp-specific connection trait that yields a connected `TcpStream`.
    * Satisfies `(C1)`, `(C2)`, `(C3)`, `(C4)`, `(C5)`.
* `(A2)` support a new generic connection trait that returns a `dyn ureq::stream::ReadWrite`.
    * Satisfies `(C1)`, `(C2)`, `(C4)`, `(C6)`.
    * Incompatible with `(C3)` as there may not be a DNS available to perform TLS auth.
    * Possibly incompatible with `(C5)`? **[Feedback desired]**
        *  Unclear if can fully work with `ReadWrite`, since all examples I could see return `Some(socket)` from `fn socket`: I don't fully understand if it's OK for it this field to be `None`.

TL;DR of the two approaches:
* `(A1)`:
    * Is simple to implement and slots right into the existing code.
    * Isn't generic enough for `(C6)`.
    * Easier to use for most users who mostly care about TCP.
    * Easily composes with TLS.
* `(A2)`:
    * Is trickier to implement due to the rest of the code relying on socket.
    * Is harder to use for users just caring about sockets (but examples can help here).
    * Is not (necessarily) composable with TLS, unless a DNS can be exposed.

@algesten (and others), kindly let me know where I'm going wrong and what my misunderstandings are: I've only stared at the code for 20 minutes :-).

amunra:
Can you point me to the timeout code you're referring to?

amunra:
I've taken a quick look at what would be involved in `(A2)`.
The first step would be to decouple `struct Stream` from `std::net::TcpStream`.
https://github.com/algesten/ureq/compare/main...amunra:ureq:connector

This is enough to compile the main code, but not the tests and doesn't even introduce a connnector trait yet.

Personally, I feel like sticking to TCP (i.e. implementing `(A1)`) may be a more pragmatic apprach, at the cost of dropping `(C6)`.

I think `(A2)` _can_ be brought to completion, but would complicate the existing logic a fair bit: The code I hacked so far feels very much square-peg-round-hole and cleaning (changing the `ReadWrite` trait) would require breaking API compatibility `(C1)`.

@algesten I'd appreciate your thoughts.

algesten:
Fair enough. Let's go with the pragmatic choice, A1.

amunra:
@zu1k, I wonder if it would make sensible to rename the trait to `TcpConnector`, this would allow more scope for a more general non-tcp specific connector to be written afterwards, should there be renewed scope for it.

zu1k:
> @zu1k, I wonder if it would make sensible to rename the trait to `TcpConnector`, this would allow more scope for a more general non-tcp specific connector to be written afterwards, should there be renewed scope for it.

It should be so.

algesten:
A variant of this is done in 3.x

PR #519, opened 2022-05-19
feat: Set multi query parameters

zu1k:
Add a `query_vec` function for setting multi query parameters. To avoid repeated deserialization and serialization. 

Signed-off-by: zu1k <i@zu1k.com>

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -389,22 +389,28 @@ impl Request {
     /// ```
     /// # fn main() -> Result<(), ureq::Error> {
     /// # ureq::is_test(true);
-    /// 
+    ///
     /// let query = vec![
     ///     ("format", "json"),
     ///     ("dest", "/login"),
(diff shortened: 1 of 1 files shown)


algesten:
Hi! Thanks for this! If we are to add this, I think the API should avoid `Vec`, and be:

```rs
 pub fn query_pairs(mut self, pairs: &[(&str, &str)]) -> Self {
```

or it could go all the way generic, but we generally have avoided "complicated" trait bounds in other contexts.

```rs
 pub fn query_pairs(mut self, pairs: impl IntoIter<(&str, &str)>) -> Self {
```

@jsha how do you feel about this API addition?

algesten:
I think this looks good. Unless @jsha objects, we can merge it.

jsha:
Late to the party, but his looks great to me!

PR #526, opened 2022-07-02
Buffer short responses

### LINKED ISSUE #264: Pool stream early in do_from_read when whole response is in buffer

### LINKED ISSUE #505: Get content from reference to response

algesten:
For small HTTP responses, we can store the entire response in an
internal buffer and release the stream back to the connection pool.

This is an alternative implementation to #513

Closes #505 
Closes #264

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -1,6 +1,7 @@
-use std::io::{self, Read};
+use std::io::{self, Cursor, Read};
 use std::str::FromStr;
 use std::sync::Mutex;
+use std::time::Duration;
 use std::{fmt, io::BufRead};
 
 use chunked_transfer::Decoder as ChunkDecoder;
@@ -69,7 +70,7 @@ pub struct Response {
     // Boxed to avoid taking up too much size.
     unit: Box<Unit>,
     // Boxed to avoid taking up too much size.
-    stream: Mutex<Box<Stream>>,
+    stream: Mutex<ResponseStream>,
     /// The redirect history of this response, if any. The history starts with
     /// the first response received and ends with the response immediately
     /// previous to this one.
@@ -81,6 +82,34 @@ pub struct Response {
     length: Option<usize>,
     /// The compression type of the response body.
     compression: Option<Compression>,
+}
+
+enum ResponseStream {
+    /// A streaming response.
+    Stream(Box<Stream>),
+    /// A fully prebuffered response.
+    Buffered(Cursor<Vec<u8>>),
+    /// Used temporarily when prebuffering a response.
+    Empty,
+}
+
+impl ResponseStream {
+    fn set_read_timeout(&self, timeout: Option<Duration>) -> io::Result<()> {
+        if let ResponseStream::Stream(stream) = self {
+            stream.set_read_timeout(timeout)
+        } else {
+            Ok(())
+        }
+    }
+
+    #[cfg(test)]
+    fn written_bytes(self) -> Vec<u8> {
+        match self {
+            ResponseStream::Stream(stream) => stream.written_bytes(),
+            ResponseStream::Buffered(buffer) => buffer.into_inner(),
+            _ => unreachable!(),
+        }
+    }
 }
 
 /// index into status_line where we split: HTTP/1.1 200 OK
@@ -261,6 +290,22 @@ impl Response {
     /// # }
     /// ```
     pub fn into_reader(self) -> impl Read + Send {
+        let stream = self.take_stream();
+        self.make_reader(stream)
+    }
+
+    /// Replaces `Response::stream` with ResponseStream::Empty and returns the result.
+    fn take_stream(&self) -> ResponseStream {
+        let mut stream = self.stream.lock().unwrap();
+        assert!(
+            !matches!(*stream, ResponseStream::Empty),
+            "ResponseStream already taken"
+        );
+        std::mem::replace(&mut *stream, ResponseStream::Empty)
+    }
+
+    /// Constructs a reader given the headers in this Response and the provided ResponseStream.
+    fn make_reader(&self, stream: ResponseStream) -> Box<dyn Read + Send> {
         //
         let is_http10 = self.http_version().eq_ignore_ascii_case("HTTP/1.0");
         let is_close = self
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -261,6 +290,22 @@ impl Response {
...
+        let mut stream = self.stream.lock().unwrap();
+        assert!(
+            !matches!(*stream, ResponseStream::Empty),
+            "ResponseStream already taken"
+        );
+        std::mem::replace(&mut *stream, ResponseStream::Empty)
+    }
+
+    /// Constructs a reader given the headers in this Response and the provided ResponseStream.
+    fn make_reader(&self, stream: ResponseStream) -> Box<dyn Read + Send> {
```
jsha, line 308: This could take owned `self`, to make clear it can only be called once per `Response`.

src/response.rs
```diff
@@ -291,27 +336,37 @@ impl Response {
...
+                        &unit.url,
+                        LimitedRead::new(stream, len),
+                    )),
+                    (false, None) => Box::new(stream),
+                }
+            }
+            ResponseStream::Buffered(buffer) => match (use_chunked, limit_bytes) {
+                (true, _) => Box::new(ChunkDecoder::new(buffer)),
+                (false, Some(len)) => Box::new(LimitedRead::new(buffer, len)),
+                (false, None) => Box::new(buffer),
```
jsha, line 367: For the Buffered case, `(use_chunked, limit_bytes)` should always be `false, Some(n)`. If were using chunked encoding, that means we don't know the length up front, so we can't say "we know we've already read the full body" based solely on the headers. Note: this means the `let length = get_header(&headers, "content-length")` calculation below on line 553 has to be more refined to use the same logic as here. Specifically: we use content-length only if `Transfer-Encoding` is not set ("[If a message is received with both a Transfer-Encoding and a Content-Length header field, the Transfer-Encoding overrides the Content-Length](https://httpwg.org/specs/rfc9112.html#message.body.length)")

By the same logic, limit_bytes should always be `Some` because if we're in the Buffered case, we know that `Content-Length` is present.

src/response.rs
```diff
@@ -507,18 +562,44 @@ impl Response {
 
         let url = unit.url.clone();
 
-        Ok(Response {
+        // Prefetch data if it's small enough to be buffered in one go.
+        let peeked_len = stream.fill_buf()?.len();
```
jsha, line 566: I think we shouldn't call `fill_buf()` here. It [was already called by BufReader at least once](https://doc.rust-lang.org/nightly/src/std/io/buffered/bufreader.rs.html#257-369), the moment we read from this response. Calling it again will result in another syscall, even in the common case where we've already read the full body but just don't know it yet because we haven't checked.

Also I think `peeked_len` should be called `bytes_available` or similar, since we haven't already peeked.

src/response.rs
```diff
@@ -507,18 +562,44 @@ impl Response {
...
+
+        let resp = Response {
             url,
             status_line,
             index,
             status,
             headers,
             unit: Box::new(unit),
-            stream: Mutex::new(Box::new(stream.into())),
+            stream: Mutex::new(ResponseStream::Empty),
```
jsha, line 576: It feels awkward to construct the Response with `Empty` (an invalid state!) and then replace it with one or the other thing later. Can we construct the ResponseStream first, and then build Response with it directly?

src/response.rs
```diff
@@ -507,18 +562,44 @@ impl Response {
...
             length,
             compression,
-        })
+        };
+
+        let boxed_stream = ResponseStream::Stream(Box::new(stream.into()));
+
+        let rstream = if is_fully_buffered {
+            // We must read the data through the PoolReturnRead to properly return the stream
+            // to the pool once read to end. This also ensures we handle chunked responses correctly.
```
jsha, line 586: We shouldn't be handling chunked responses here because the content length is known.

jsha (changes requested):

src/response.rs
algesten: Very true. Silly me.

src/response.rs
algesten: Not easily. The only way to get the stream back to the pool, is to read the buffered response via the full reader. I looked at refactoring the `PoolReturnRead` logic, but it's very involved – depends on `Stream`.

This is why `make_read` does not take `self`

src/response.rs
algesten: It can't. I specifically made this variant of `make_reader` to _not_ take self so that we can create a reader without destroying the `Response` struct. See other comment.

src/response.rs
algesten: Yes. I'll fix this.

src/response.rs
algesten: So how do we know whether to buffer the response?

src/response.rs
jsha: Ah, that makes sense.

Rather than refactoring the PoolReturnRead logic, what about bypassing / duplicating it? Once we've concluded that we're reading by Content-Length, the logic is simple: we take N bytes from the BufReader, reset the stream's deadlines, and call `agent.state.pool.add(&PoolKey::new(url, agent.config.proxy.clone(), stream)`.

src/response.rs
algesten: Ok. I'll do that.

src/response.rs
algesten: Groan. More pooling code? :)

Do you think it will be simpler than what we got here?

jsha: Heh, yeah, more pooling code is a downside. I think it could possibly be simpler, in part because it lets us play to the strengths of BufReader: namely that you get access to a slice of bytes, then inform BufReader "hey, I took N of those bytes," avoiding any `.read()` call.

But I'm okay with the create-and-modify.

algesten:
Close in favor of #531

Issue #516, opened 2022-05-09
Is ureq's MIT license legal?


sycured:
Hello,

I'm actually doing a big check (SBOM, licensing & co) on my different work, and obtain a red line about one dependency used by ureq: chunked_transfer which is Apache-2.0 only, a license that I can't use inside 3 projects.

ureq is dual-licensing Apache-2.0 and MIT, so can you confirm that your MIT isn't violating Apache-2.0?

For example, I can mention:

> I want to add code from apache 2.0 to MIT license. How does one do that?
> request that the copyright owner of the apache 2.0 licensed works also release what they did under MIT license. OTHERWISE, you will have to maintain the apache 2.0 license
> source:  https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project#comment157333_6758

>  If MIT Licensed code is combined with Apache License 2.0 code, then the derivative code might be Apache License 2.0.
> source: https://techlib.fr/app/1274/is-apache-2-0-compatible-with-mit

Thanks a lot for your help

algesten:
Can we solve this with feature flags?

Keep `chunked_transfer` as a default dependency, but point out in the README that the project can't be MIT licensed if the dependency is enabled?

sycured:
Yes, it'll work

Is it possible to have ureq working without this dependency?

algesten:
Well. It won't be as good :)

Chunked transfer is used in HTTP/1.1 when the length of a body is not known. Many requests would work without it, but not all.

Another option is to talk to the upstream maintainer.


A third option is to implement this ourselves – I have done it once before https://github.com/algesten/hreq-h1/blob/master/src/chunked.rs – we could remove `async` from this code and use that.

jsha:
Posted on https://github.com/frewsxcv/rust-chunked-transfer/issues/2

pinkforest:
PR's up at chunked-transfer:
- https://github.com/frewsxcv/rust-chunked-transfer/pull/13

algesten:
Closing since we're moving to ureq 3.x. This is however solved in 2.x too.

Issue #515, opened 2022-05-08
Sync bound on ReadWrite


jsha:
Following up the conversation in https://github.com/algesten/ureq/pull/510#issuecomment-1114202093. If we merge #514 I think that is a satisfactory resolution for me, but it's worth having a bit more discussion here. @algesten said:

> With regards to `Sync`, I think we need to have "two directions" in mind.
>
> 1) Types that ureq provides to the user, i.e. `Request`, `Respone`, `Agent`, `Error`. To maximize the utility of ureq, we should strive to have all these as `Sync`. That way the user can "stash away" an Agent or a Response and we would not "taint" the user's context with non-Sync.
>
> 2) Types we expect from the user, i.e. `impl TlsConnector`, `impl Resolver`, `impl Middleware`. For these, we should try to avoid `Sync` as much as reasonable, since that will make it easier for the user to integrate with ureq (we can go even further here in what traits we might be able to avoid, such as `Send`, `'static` etc).

You're right that it's nice to the user to require as few bounds as possible for user-provided objects. But `impl TlsConnector` is notably different than `Resolver` and `Middleware`. For `Resolver` and `Middleware` we expect a wide variety of users to implement them themselves. For `TlsConnector`, there are only a handful of TLS implementations in the world. We can look at the actual implementations to see what bounds are reasonable, and which are burdensome. So far we've looked at three TLS implementations and found that `Sync` is trivial for native-tls and rustls; it is every so slightly non-trivial for mbedtls: the wrapper has to use a Mutex to account for the fact that the library is not full duplex. Given that, I think it makes more sense philosophically to put the burden on the mbedtls ureq adapter rather than on ureq itself.

> We will never have unsafe use directly in ureq, but crates we depend upon will always have it. `sync_wrapper` is entirely sound (I think it could be included in Rust std lib). This is the current situation with geiger.

Yes, crates we depend upon will always have some unsafe. What's the reason we accept `unsafe` in crates we depend upon but not in our own crate? Here's my reasoning: some things can't be done without it. For instance, `flate2` uses FFI to depdend on a highly optimized system library. Similarly `ring` uses FFI to call out to highly optimized assembly. `log` uses `unsafe` to mutate a global static. We depend on those libraries because (a) we need them to get things done, and (b) they are heavily depended upon that we expect the soundness of their unsafe blocks has been looked at by a decent number of people.

[...]

PR #527, opened 2022-07-02
Release 2.5.0

algesten:


algesten:
@jsha I think it's time for a release

Question is whether we want to land #526 and do something about #516. 

Anything else you been thinking about?

jsha:
Yes, let's do a release! I'll review #526. I think we shouldn't block a release on #516, but should ask the upstream maintainer.

I'm also still thinking about #515. I'd love to resolve that before cutting the release since it implicates API guarantees. I'll post on that issue.

PR #530, opened 2022-07-07
Simplify ReadWrite interface

jsha:
Previously, ReadWrite had methods `is_poolable` and `written_bytes`, which were solely for the use of unittests.

This replaces `written_bytes` and `TestStream` with a `struct Recorder` that implements `ReadWrite` and allows unittests to access its recorded bytes via an `Arc<Mutex<Vec<u8>>>`. It eliminates `is_poolable`; it's fine to pool a Stream of any kind.

The new `Recorder` also has some convenience methods that abstract away boilerplate code from many of our unittests.

I got rid of `Stream::from_vec` and `Stream::from_vec_poolable` because they depended on `TestStream`. They've been replaced by `NoopStream` for the pool.rs tests, and `ReadOnlyStream` for constructing `Response`s from `&str` and some test cases.

### DIFF
--- a/examples/custom-tls.rs
+++ b/examples/custom-tls.rs
@@ -52,9 +52,6 @@ impl ReadWrite for CustomTlsStream {
     fn socket(&self) -> Option<&TcpStream> {
         self.0.socket()
     }
-    fn is_poolable(&self) -> bool {
-        self.0.is_poolable()
-    }
 }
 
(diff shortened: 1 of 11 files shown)


algesten:
This is a really nice simplification. I like it a lot!

algesten (approved):

PR #531, opened 2022-07-07
Buffer short responses (2)

jsha:
In Response, rather than storing a `Box<Stream>` and turning it into a `Box<dyn Read + Send>`, do that transformation at construction time and store a `reader` on the Response.

This allows us to use the same length calculations as we use for LimitedRead, and avoid turning Stream into an enum. Since `Cursor<Vec<u8>>` is naturally `Read + Send`, we don't need a wrapper type; we can Box it an return it directly.

This happens to save us an allocation, since we never box the Stream, just the reader.

This is an alternative implementation to #513 and #526

Closes https://github.com/algesten/ureq/issues/505
Closes https://github.com/algesten/ureq/issues/264

This is built on top of #530, because written_bytes was getting in the way of this implementation. The commit that is unique to this branch is 9ecb86a7fc4d142d0a15aa7f93acbbc3e9467b3b.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -1,4 +1,4 @@
-use std::io::{self, Read};
+use std::io::{self, Cursor, Read};
 use std::str::FromStr;
 use std::{fmt, io::BufRead};
 
@@ -67,8 +67,7 @@ pub struct Response {
     headers: Vec<Header>,
     // Boxed to avoid taking up too much size.
     unit: Box<Unit>,
-    // Boxed to avoid taking up too much size.
-    stream: Box<Stream>,
+    reader: Box<dyn Read + Send + Sync + 'static>,
     /// The redirect history of this response, if any. The history starts with
     /// the first response received and ends with the response immediately
     /// previous to this one.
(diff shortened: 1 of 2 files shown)


algesten (approved):
This does indeed look like a better route than mine.

I pushed some commits that I think made sense. Let me know what you think.

I had ever seen `[0u8].repeat(x)` to construct a `Vec` before. I tested it on godbolt with optimize level 3 and cursory it looks a bit less efficient than just allocating the correct number of bytes. Obviously both these methods might be irrelevant once we get `ReadBuf`.

jsha:
> I had ever seen [0u8].repeat(x) to construct a Vec before.

Neither had I! But I was having a brain freeze on the right way to construct a Vec of zeroes, like when you can't quite think of a word but say a word that sounds like it instead. 😂 Thanks for the patch there; of course `vec![0; len]` is much tidier and cooler.

Your change look good. I especially like that we can eliminate the unwrap/re-wrap of DeadlineStream. That is a good bit of tidying.

algesten (approved):

PR #533, opened 2022-07-27
Simplify Response object

jsha:
Now that we create a Reader immediately on construction, we can get rid of `unit`, `length`, and `compression` by passing the necessary parameters directly to stream_to_reader. This also lets us avoid temporarily making a half-constructed Response so we can call stream_to_reader on it. Now stream_to_reader is an associated method that doesn't take `Self` so it can be called during construction.

Followup to #531.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -257,22 +257,22 @@ impl Response {
     }
 
     fn stream_to_reader(
+        unit: &Unit,
         stream: DeadlineStream,
-        method: &str,
-        url: &Url,
         status: u16,
         http_version: &str,
         headers: &[Header],
         compression: Option<Compression>,
-        agent: &Agent,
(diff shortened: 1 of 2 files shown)


algesten:
@jsha In my mind, the whole reason for `Unit` is to encapsulate the request arguments. I think this PR does do a nice cleanup, but we could use `Unit` as an argument into `stream_to_reader`. I pushed a commit to that effect.

Reading the code there is another problem.

1. Here we remove [`content-length`](https://github.com/algesten/ureq/pull/533/files#diff-6f9f8f882acd334c5bf0b598b309afef5f90f63d20b4cc6c26751d66bcb1dba6R522) from `headers`
2. We pass `headers` to `stream_to_reader`
3. Then we try to parse `content-length` from headers in `stream_to_reader`.

algesten:
Closing to clean up for 3.x

PR #536, opened 2022-08-25
Fix: discards cookie header while redirection

walkure:
When ureq receives 302 redirect, it calls `Unit::new` without discarding the cookie header.
https://github.com/algesten/ureq/blob/134d82ecf4f8905f4ec84080adb1839f2de115ea/src/unit.rs#L224

So, the request header which being sent to redirected target will contain cookies both before and after the redirect.

Is is intended action?

When using ureq for http scraping, ureq failure to login some web sites because of session disruption.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -218,10 +218,10 @@ pub(crate) fn connect(
         // reuse the previous header vec on redirects.
         let mut headers = unit.headers;
 
-        // on redirects we don't want to keep "content-length". we also might want to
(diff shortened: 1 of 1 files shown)


algesten:
@jsha this does seem reasonable. Since you're "the spec guy", just wanted to check you're cool with landing this.

@walkure two things this PR needs.

1. Run `cargo fmt`
2. A test establishing this behavior. Copy `redirect_no_keep_auth_different_host` which tests that `authorization` isn't kept.

algesten:
@jsha you ok with landing this?

PR #542, opened 2022-09-20
Add method to obtain buffered body from response

### LINKED ISSUE #505: Get content from reference to response

mladedav:
/closes #505 

Initial implementation, I would like feedback on the direction.

I admit the matching across the new `Reader` is not nice, but there is currently no way to discriminate between `dyn Read + Send + Sync + 'static` that is backed by a connection or `Cursor`. I tried making downcasting work instead but was unable to do so.

Unresolved thins:
This is missing configuration for size of the buffer.
Docs are missing.
The method returns `None` if the response was compressed.


src/response.rs
```diff
@@ -82,6 +82,22 @@ pub struct Response {
     compression: Option<Compression>,
 }
 
+enum Reader {
+    Buffered(Box<Cursor<Vec<u8>>>),
```
algesten, line 86: ~What does the `Box` do here? Isn't `Cursor<Vec<u8>>` enough since `Vec` will be on the heap anyway?~

Strike that. It's because we don't want to change `into_reader` signature further down.

Did you try going the route of having `Box<Reader>` in the struct and remove the internal boxes in the enum? Just curious if it gets less boxy.

algesten (approved):
Overall I like it. We haven't made any config for how big the buffer is, but I that can come as a follow-up.

Small code hygiene question in PR.

src/response.rs
mladedav: I didn't try that because of the variant `Connection(Box<dyn Read + Send + Sync + 'static>),`. This one needs to be boxed anyway (because I cannot use `dyn` on its own) and then the other variant ended with a `Box` for the reason you already found.

If we wanted to use concrete types in the enum, there would have to be more types because currently `Connection` is used for an open TCP socket as well as a compression wrapper on both `Cursor` and connection.

mladedav:
Ok, I've ran cargo fmt and added some docs so this should be fine then. I'll look into the buffer size and compression.

algesten:
Would you mind rebasing off the latest main, so we can see a clean test run?

mladedav:
It's rebased, you can try running it

algesten:
Sorry. It's needing another rebase due to a change that just landed in main.

mladedav:
It's there

mladedav:
Hi, I'm not using `ureq` anymore and this got pretty old with more conflicts so I'll close this PR.

Feel free to reuse the code if you find it useful.

PR #544, opened 2022-09-28
add HTTPS proxy support

puffi:
The addition of HTTPS proxy support requires a change of the `connect_host` interface. The return type is changed, but since this is only used internally in this one file and has visibility pub(crate) this should not be a breaking change.

The TCP Stream now gets wrapped inside the connect_host function which makes the `from_tcp_stream` function of Stream unused.

Signed-off-by: Mira Limbeck <m.limbeck@proxmox.com>

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -409,12 +409,21 @@ fn send_prelude(unit: &Unit, stream: &mut Stream) -> io::Result<()> {
         // HTTP proxies require the path to be in absolute URI form
         // https://www.rfc-editor.org/rfc/rfc7230#section-5.3.2
         match proxy.proto {
-            Proto::HTTP => format!(
-                "{}://{}{}",
-                unit.url.scheme(),
-                unit.url.host().unwrap(),
-                unit.url.path()
-            ),
+            Proto::HTTP => match unit.url.port() {
+                Some(port) => format!(
+                    "{}://{}:{}{}",
+                    unit.url.scheme(),
+                    unit.url.host().unwrap(),
+                    port,
+                    unit.url.path()
+                ),
+                None => format!(
+                    "{}://{}{}",
+                    unit.url.scheme(),
+                    unit.url.host().unwrap(),
+                    unit.url.path()
+                ),
+            },
             _ => unit.url.path().into(),
         }
     } else {


algesten:
Would you mind rebasing off latest main? So we can see a clean test run.

puffi:
It should be rebased now, at least I see the force-push mentioned here.

Sorry, first time doing that here on Github.

puffi:
Sorry about that, it seems to be the unused `from_tcp_stream` function I left in which leads to the errors.
Do you want me to remove it from the source?

algesten:
@puffi Yeah, that be great. Makes it compile cleanly in the CI. 👍

algesten:
@jsha what do you think about the extra `Box<dyn>` indirection we get in this PR?

I think the PR looks fine overall, but a bit concerned about this.

src/stream.rs
```diff
@@ -407,27 +410,38 @@ pub(crate) fn connect_host(
         stream.set_write_timeout(unit.agent.config.timeout_write)?;
     }
 
-    if proto == Some(Proto::HTTPConnect) {
+    if proto == Some(Proto::HTTPConnect) || proto == Some(Proto::HTTPSConnect) {
         if let Some(ref proxy) = proxy {
-            write!(stream, "{}", proxy.connect(hostname, port)).unwrap();
-            stream.flush()?;
+            let mut s;
```
jsha, line 415: I'd love to have a more descriptive variable name here, and a type annotation. With the current code, the reader has to read _forward_ in the code to figure out what type `s` is.

src/stream.rs
```diff
@@ -407,27 +410,38 @@ pub(crate) fn connect_host(
...
+            if proto == Some(Proto::HTTPSConnect) {
+                s = unit
+                    .agent
+                    .config
+                    .tls_config
+                    .connect(&proxy.server, Box::new(stream))?;
+            } else {
+                s = Box::new(stream);
+            }
+            write!(s, "{}", proxy.connect(hostname, port)).unwrap();
```
jsha, line 425: I see that we already have this `.unwrap()`, but it would be nice to cleanup while we're here to use a `?` instead.

src/stream.rs
```diff
@@ -317,12 +317,12 @@ pub(crate) fn connect_https(unit: &Unit, hostname: &str) -> Result<Stream, Error
     Ok(Stream::new(https_stream, remote_addr))
 }
 
-/// If successful, returns a `TcpStream` and the remote address it is connected to.
+/// If successful, returns a `Box<dyn ReadWrite>` around a TcpStream and the remote address it is connected to.
```
jsha, line 320: This comment is not strictly accurate, since this could now be wrapping a TcpStream _or_ a TLS stream.

jsha (changes requested):
I think the `Box<dyn>` indirection is not a big deal - we already take a `Box<dyn>` (for good reason) in `stream_to_reader`. This introduces a second level of dynamic dispatch, and it would be nice to collapse them into a single level.

In terms of code structure, we have `connect_http` and `connect_https`, which both call `connect_host`. For the latter, it then wraps the result of `connect_host` in an HTTPS stream. Coincidentally (or not), this is where the existing layer of dynamic dispatch is introduced. It would be nice for the "HTTPS proxy" code path to go through `connect_https` as well, which would require some code reorganization. Conceptually we have:

 - direct connect, plaintext
 - direct connect, HTTPS
 - proxy, plaintext
 - proxy, HTTPS

It would be nice if we could have a set of functions that reflect that more explicitly.

DanGould:
Any reason this shouldn't be re-opened? I started implementing the same functionality before I found this PR. I ran into a small issue where we're calling [`Response::do_from_stream` now](https://github.com/algesten/ureq/blob/2300a5b3787b1ac800314dc47f075f4850026bb3/src/stream.rs#L458) which takes ownership of the tls_stream that we ultimately need to return, and there's no easy way to clone a `Box<dyn ReadWrite>` returned from `tls_conf.connect(..)`, but it seems surmountable.

PR #550, opened 2022-10-18
Fix gzip + chunk pool problem

### LINKED ISSUE #549: Compressed, chunked response inhibits connection reuse
```rust
env_logger::init();
let agent: Agent = ureq::AgentBuilder::new().build();
for _ in 0..2 {
    let resp = agent.get("https://www.jammersreviews.com/st-tos/s1/dagger.php").call().unwrap();
    println!("Connection: {}", resp.header("connection").unwrap());
    std::io::copy(&mut resp.into_reader(), &mut io::sink()).unwrap();
}
```
results in (RUST_LOG=trace)
```
[2022-10-17T01:02:37Z DEBUG ureq::stream] connecting to www.jammersreviews.com:443 at 172.67.140.111:443
[2022-10-17T01:02:37Z DEBUG rustls::client::hs] No cached session for DnsName(DnsName(DnsName("www.jammersreviews.com")))
[...]
[2022-10-17T01:02:37Z DEBUG ureq::stream] created stream: Stream(RustlsStream)
[2022-10-17T01:02:37Z DEBUG ureq::unit] sending request GET https://www.jammersreviews.com/st-tos/s1/dagger.php
[2022-10-17T01:02:37Z DEBUG ureq::unit] writing prelude: GET /st-tos/s1/dagger.php HTTP/1.1
    Host: www.jammersreviews.com
    User-Agent: ureq/2.5.0
    Accept: */*
    accept-encoding: gzip
[2022-10-17T01:02:37Z DEBUG rustls::client::tls13] Ticket saved
[2022-10-17T01:02:37Z DEBUG rustls::client::tls13] Ticket saved
[2022-10-17T01:02:37Z DEBUG ureq::response] Chunked body in response
[2022-10-17T01:02:37Z DEBUG ureq::unit] response 200 to GET https://www.jammersreviews.com/st-tos/s1/dagger.php
Connection: keep-alive
[2022-10-17T01:02:37Z DEBUG ureq::stream] dropping stream: Stream(RustlsStream)
[2022-10-17T01:02:37Z DEBUG ureq::stream] connecting to www.jammersreviews.com:443 at 172.67.140.111:443
[2022-10-17T01:02:37Z DEBUG rustls::client::hs] Resuming session
[...]
[2022-10-17T01:02:37Z DEBUG ureq::stream] created stream: Stream(RustlsStream)
[2022-10-17T01:02:37Z DEBUG ureq::unit] sending request GET https://www.jammersreviews.com/st-tos/s1/dagger.php
[2022-10-17T01:02:37Z DEBUG ureq::unit] writing prelude: GET /st-tos/s1/dagger.php HTTP/1.1
    Host: www.jammersreviews.com
    User-Agent: ureq/2.5.0
    Accept: */*
    accept-encoding: gzip
[2022-10-17T01:02:37Z DEBUG rustls::client::tls13] Ticket saved
[2022-10-17T01:02:37Z DEBUG rustls::client::tls13] Ticket saved
[2022-10-17T01:02:37Z DEBUG ureq::response] Chunked body in response
[2022-10-17T01:02:37Z DEBUG ureq::unit] response 200 to GET https://www.jammersreviews.com/st-tos/s1/dagger.php
Connection: keep-alive
[2022-10-17T01:02:37Z DEBUG ureq::stream] dropping stream: Stream(RustlsStream)
```
As can be seen, though TLS does session resumption, the TCP stream gets closed and reopened between requests. I believe this is caused by the fact that the `GzDecoder` knows when to expect the end of the gzip stream, never tries to read beyond it, and therefore never makes the read that causes the `ChunkDecoder` to give it Ok(0), which is what the `PoolReturnRead` looks for to know when to return the `Stream` to its pool. I was able to get it to work properly by using a `MultiGzDecoder` instead of a plain `GzDecoder`, at least in my private, heavily-modified version of ureq. But I don't think any such thing is available for Brotli.

algesten:
Close #549

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -446,6 +446,7 @@ mod tests {
     // decoder reads the exact amount from a chunked stream, not past the 0. This
     // happens because gzip has built-in knowledge of the length to read.
     #[test]
+    #[cfg(feature = "gzip")]
     fn read_exact_chunked_gzip() {
         let gz_body = vec![
             b'E', b'\r', b'\n', // 14 first chunk
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -304,10 +304,8 @@ impl<R: Read + Sized + Done + Into<Stream>> Read for PoolReturnRead<R> {
 mod tests {
     use std::io;
 
-    use crate::response::Compression;
     use crate::stream::{remote_addr_for_test, Stream};
     use crate::ReadWrite;
-    use chunked_transfer::Decoder as ChunkDecoder;
 
     use super::*;
 
@@ -448,6 +446,9 @@ mod tests {
     #[test]
     #[cfg(feature = "gzip")]
     fn read_exact_chunked_gzip() {
+        use crate::response::Compression;
+        use chunked_transfer::Decoder as ChunkDecoder;
+
         let gz_body = vec![
             b'E', b'\r', b'\n', // 14 first chunk
             0x1F, 0x8B, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x03, 0xCB, 0x48, 0xCD, 0xC9,
@@ -462,8 +463,6 @@ mod tests {
             b'0', b'\r', b'\n', //
             b'\r', b'\n', //
         ];
-
-        println!("{:?}", gz_body);
 
         impl ReadWrite for io::Cursor<Vec<u8>> {
             fn socket(&self) -> Option<&std::net::TcpStream> {
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -464,18 +464,6 @@ mod tests {
             b'\r', b'\n', //
         ];
 
-        impl ReadWrite for io::Cursor<Vec<u8>> {
-            fn socket(&self) -> Option<&std::net::TcpStream> {
-                None
-            }
-        }
-
-        impl From<io::Cursor<Vec<u8>>> for Stream {
-            fn from(c: io::Cursor<Vec<u8>>) -> Self {
-                Stream::new(c, "1.1.1.1:8080".parse().unwrap())
-            }
-        }
-
         let agent = Agent::new();
         let url = Url::parse("https://example.com").unwrap();
 
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -302,8 +302,12 @@ impl<R: Read + Sized + Done + Into<Stream>> Read for PoolReturnRead<R> {
 
 #[cfg(test)]
 mod tests {
-    use crate::stream::remote_addr_for_test;
+    use std::io;
+
+    use crate::response::Compression;
+    use crate::stream::{remote_addr_for_test, Stream};
     use crate::ReadWrite;
+    use chunked_transfer::Decoder as ChunkDecoder;
 
     use super::*;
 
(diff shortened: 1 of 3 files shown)


algesten:
@jsha gentle reping 🤗

jsha:
I will take another look at this soon! Thanks for the ping.

jsha:
This isn't quite what I had in mind with https://github.com/algesten/ureq/issues/549#issuecomment-1286639797. The DidReadZero thing is cool, but we can get away without it, only wrapping another reader on the outside, not on the outside and inside.

I wrote up a demo to try and show what I meant here:

```rust
use std::io::{Read, Error, ErrorKind};

const BUFFER_SIZE: usize = 8000;

pub(crate) struct BrotliReader<R: Read> {
    inner: Option<brotli_decompressor::Decompressor<R>>
}

impl<R: Read> BrotliReader<R> {
    fn new(inner: R) -> Self {
        Self { inner: Some(brotli_decompressor::Decompressor::new(inner, BUFFER_SIZE)) }
    }

    fn verify_wrapped_reader_done(&mut self) -> std::io::Result<usize> {
        let r = match self.inner.take() {
            Some(r) => r,
            None => return Ok(0),
        };
        let mut inner = r.into_inner();
        match inner.read(&mut [0u8; 1]) {
            Ok(0) => Ok(0),
            Err(e) => Err(e),
            Ok(_) => Err(Error::new(ErrorKind::InvalidData, "extra data after valid Brotli stream")),
        }
    }
}

impl<R: Read> Read for BrotliReader<R> {
    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        let inner = match &mut self.inner {
            Some(d) => d,
            None => return Ok(0),
        };
        match inner.read(buf) {
            Ok(0) => self.verify_wrapped_reader_done(),
            Ok(n) => Ok(n),
            r => r,
        }
    }
}

#[cfg(test)]
mod tests {
    use std::io::{Cursor, ErrorKind};
    use std::error::Error;
    use super::BrotliReader;

    #[test]
    fn test_no_leftovers() -> Result<(), Box<dyn Error>> {
        // Output from this command:
        // (brotli -c <<<"hello"; echo goodbye) | xxd -p | sed 's/../\\x&/g'
        let compressed_with_extra = b"\x8f\x02\x80\x68\x65\x6c\x6c\x6f\x0a\x03\x67\x6f\x6f\x64\x62\x79\x65\x0a";
        let mut reader = BrotliReader::new(Cursor::new(compressed_with_extra));
        let output = std::io::read_to_string(&mut reader);
        assert!(output.is_err());
        assert_eq!(output.err().unwrap().kind(), ErrorKind::InvalidData);
        Ok(())
    }
}
```

Unfortunately, the test fails! But if you change BUFFER_SIZE to 10 (the size of the "good" Brotli bytes), it passes. This is because of the internal buffering in the Brotli decompressor. With a BUFFER_SIZE of 8000, it has already read the full 18 byte input into the buffer, and then `into_inner` drops the unused 8 bytes on the floor, so reading from the underlying Cursor returns Ok(0) as if everything is fine. I filed an upstream issue: https://github.com/dropbox/rust-brotli/issues/84. Without a fix for that, I don't think we can properly fix this for Brotli on our side. I think we should do the simple fix for gzip (switch to MultiGzDecoder), and wait for an upstream fix for Brotli.

algesten:
What guarantee do we have that the extra read isn't going to sync wait for the next byte from the server?

algesten:
Or put another way: I think technically in rust you should only read to _once_. 

There's no guarantee that multiple reads to 0 will be a good state. 

We can guarantee that ourselves by knowing exactly what deeper readers are hiding behind that into inner.

My solve was trying to ensure we don't rely on such deeper behavior.

jsha:
> What guarantee do we have that the extra read isn't going to sync wait for the next byte from the server?

The next layer down is one of:

 1. a chunked_transfer::Decoder
 2. a LimitedRead
 3. a plain stream (for close-delimited responses, in HTTP/1.0)
 
For all of these, if the response is done, they will return Ok(0), not sync wait. If the response is _not_ done, they should sync wait and when the next byte eventually comes through, the response should error out (due to extra junk at the end). For (3), that sync wait might result in an eventual `Ok(0)` when the TCP connection is closed.

> I think technically in rust you should only read to zero once.

This is a good point. Specifically [`io::Read::read`](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read) doesn't make any requirement that `Ok(0)` means a stream is permanently done, and in some cases it explicitly doesn't mean done:

> for [File](https://doc.rust-lang.org/nightly/std/fs/struct.File.html), it is possible to reach the end of file and get zero as result, but if more data is appended to the file, future calls to read will return more data.

One solution to this is the equivalent of [std::iter::Fuse](https://doc.rust-lang.org/nightly/std/iter/struct.Fuse.html) but for `Read`. I'm surprised this doesn't exist in the stdlib! But it would be easy to write our own, and wrap LimitedRead / Decoder in it. Alternately, we could have a marker trait that indicates "this reader implements fused semantics" (which LimitedRead does, but I think chunked_transfer::Decoder does not, out of the box).

PR #554, opened 2022-11-12
add badges to crates.io and docs.rs to README.md

wolthom:


algesten:
Thanks for this!

The problem is that the readme is generated from the lib.rs - so any changes must be done there and then run `cargo readme`

wolthom:
@algesten I tried to update the PR. If I understood `cargo-readme` correctly, I added the section to the `lib.rs` docstring and regenerated the `README.md` using `cargo readme > README.md`. Is that correct?

algesten:
That's exactly right. Thanks!

I'll take a look.

src/lib.rs
```diff
@@ -9,6 +9,25 @@
...
+//!  </a>
+//!  <!-- Docs -->
+//!  <a href="https://docs.rs/ureq">
+//!    <img src="https://img.shields.io/badge/docs-latest-blue.svg?style=flat-square"
+//!      alt="docs.rs docs" />
+//!  </a>
+//!  <!-- Downloads -->
+//!  <a href="https://crates.io/crates/ureq">
+//!    <img src="https://img.shields.io/crates/d/ureq.svg?style=flat-square"
+//!      alt="Download" />
```
algesten, line 27: Should this alt-text say "Crates.io downloads" or something like that?

wolthom: Good suggestion, thank you! I updated the alt text and also slightly changed the badge positions to show up at the top. In my opinion this looks a bit cleaner.

algesten:
Thanks

PR #556, opened 2022-11-17
Fix connection reuse for empty response bodies

### LINKED ISSUE #555: Zero-length responses break connection reuse
Here's the code for handling responses with fixed-length bodies:

https://github.com/algesten/ureq/blob/28d667ade9425b1d7b71f15c2444597b439651eb/src/response.rs#L325-L343

We set up the `PoolReturnRead`, and then if the body is short enough, we call `.read_exact` on the `PoolReturnRead` to both fetch the body and, as a side-effect, force `PoolReturnRead` to notice that the body has been consumed and that it's free to return the connection to the pool.

However, if the response body is empty (e.g. on a 304 Not Modified), then `len` is 0, and we call `read_exact` on a zero-length buffer. In this case, [the default implementation of `std::io::Read::read_exact`](https://doc.rust-lang.org/src/std/io/mod.rs.html#446) exits immediately. In particular, **it does not call `PoolReturnRead::read` at all**, so `PoolReturnRead` has no chance to notice that the stream has been exhausted. Then we drop the `PoolReturnRead`, and since it believes the body hasn't been exhausted, it drops the underlying connection and on our next request we'll have to set up a new connection from scratch.

(Ironically, this seems to make my program slower on a warm cache than a cold cache, because in many cases fetching the data is cheaper than restarting the connection.)

njsmith:
For empty response bodies, there's no guarantee that `PoolReturnRead::read` will ever be called, so we might never notice that the body is exhausted and can be returned to the pool. Solve this by adding an exhaustion check to `PoolReturnRead::new`. Annoyingly, this requires changing its API (adding a new trait bound + making `new` fallible), but fortunately it's an internal API and the callers are easy to update.

Fixes #555

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -234,13 +234,17 @@ pub(crate) struct PoolReturnRead<R: Read + Sized + Into<Stream>> {
     key: PoolKey,
 }
 
-impl<R: Read + Sized + Into<Stream>> PoolReturnRead<R> {
-    pub fn new(agent: &Agent, url: &Url, reader: R) -> Self {
-        PoolReturnRead {
+impl<R: Read + Sized + Done + Into<Stream>> PoolReturnRead<R> {
+    pub fn new(agent: &Agent, url: &Url, reader: R) -> Result<Self, std::io::Error> {
+        let mut pooler = PoolReturnRead {
             agent: agent.clone(),
             key: PoolKey::new(url, agent.config.proxy.clone()),
             reader: Some(reader),
+        };
+        if pooler.reader.as_ref().unwrap().done() {
+            pooler.return_connection()?;
         }
+        Ok(pooler)
     }
 
     fn return_connection(&mut self) -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


jsha:
Great find on the big, and great analysis. Thanks!

In terms of PR implementation: wouldn't it be simpler to change `stream_to_reader` to have a special case for `len() == 0` and read into a 1-byte array to trigger the return?

njsmith:
That's actually what I did initially, but (a) I felt pretty dirty having a type `PoolReturnRead` with semantics where correct usage *required* doing that funky dance, rather than having it encapsulated in the type, and nervous that some refactoring or re-usage of `PoolReturnRead` later could reintroduce the bug, and (b) this made it easier to test, because I can test `PoolReturnRead` in isolation instead of having to trigger a fake `stream_to_reader` call.

jsha:
I've proposed an alternate approach in #565. Would be interested to hear your thoughts!

jsha:
Fixed in #565. Thanks so much for pointing this out and making the first pass at a fix.

PR #560, opened 2022-11-26
Switch to MultiGzDecoder

jsha:
This is more correct, since all gzip streams can consist of multiple members. It has the happy side effect that it causes gzipped responses to reliably return their streams to the pool.

Small fix for #549. Uses @algesten's nice unittest from #550.


jsha:
Broken lints are fixed in #561

algesten (approved):
So we're saying MultiGz is more correct?

jsha:
Yep! There's an old issue in the flate2 crate saying the same thing, with agreement from a maintainer:

https://github.com/rust-lang/flate2-rs/issues/178

PR #564, opened 2022-11-28
Don't panic when rustls-native-certs errors

### LINKED ISSUE #563: `AgentBuilder::new` panics (possibly because of ill-formed certificates)
Recently, the following issue showed up with the `juliaup` program (more info [here](https://github.com/JuliaLang/juliaup)):
https://github.com/JuliaLang/juliaup/issues/487

The backtrace suggests, that the panic happens during the call to `AgentBuilder::new`. We suspect this could be related to encountering a bad certificate. David found some potential clues here:
https://github.com/rustls/rustls-native-certs/issues/15 when a 

This issue from `reqwests` seems to be related as well:
https://github.com/seanmonstar/reqwest/pull/1316

Is this something you are aware of or could take a look at? Maybe the `reqwests` solution could work here as well.
Thanks!

jsha:
Fixes #563

### DIFF
--- a/src/rtls.rs
+++ b/src/rtls.rs
@@ -92,7 +92,7 @@ fn root_certs() -> rustls::RootCertStore {
         }
     }
     if valid_count == 0 && invalid_count > 0 {
-        error!("no valid certificates loaded by rustls-native-certs. all HTTP requests will fail.");
+        error!("no valid certificates loaded by rustls-native-certs. all HTTPS requests will fail.");
     }
     root_cert_store
 }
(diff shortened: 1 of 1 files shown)


src/rtls.rs
```diff
@@ -64,20 +64,37 @@ impl Write for RustlsStream {
...
+                    &err,
+                    &cert
+                );
+            }
+        }
     }
-
-    root_store
+    if valid_count == 0 && invalid_count > 0 {
+        error!("no valid certificates loaded by rustls-native-certs. all HTTP requests will fail.");
```
algesten, line 95: HTTPS?

algesten (approved):
Just one tiny thing. 


Thanks!

PR #566, opened 2022-12-01
Split response body type into own function

jsha:
This allows removing the hack where we create a Response with an empty `reader`, then immediately mutate it to set the real reader. It also happens to allow us to get rid of 3 fields of Response that were only used to pass information to `stream_to_reader`.

I've tried to keep the structure and logic of the body calculation as close to the same as possible, for ease of review and to avoid introducing bugs. I think there are some [followup fixes we can make to the logic](https://github.com/algesten/ureq/pull/343), which will be made easier by having it in a self contained function.

This is the refactoring I mentioned in #565. If we land this first, we [can get rid of the `error!` log](https://github.com/algesten/ureq/pull/565#discussion_r1034168939).

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -150,14 +150,14 @@ impl Header {
     }
 }
 
-pub fn get_header<'a, 'b>(headers: &'b [Header], name: &'a str) -> Option<&'b str> {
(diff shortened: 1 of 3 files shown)


algesten (approved):
Love it. That Box<Unit> in response always irked me.

PR #568, opened 2022-12-04
Re-export CookieStore

### LINKED ISSUE #548: `ureq` doesn't expose `CookieStore` and this makes `ureq 2.5.0` a breaking change
In order to set cookies in an `AgentBuilder` you need to create a `CookieStore`, however `ureq` doesn't re-export it: to do so you need to explicitly depend on either the `cookie` or the `cookie_store` crates.

However this is not just poor ergonomics, it creates a semver hazard: since the direct dependency is not linked to `ureq`'s, and `0.x` updates are considered not compatible semver-wise, any update to `ureq`'s `cookie` dependency is a breaking change[^1] because it results in two copies of the `cookie`/`cookie_store` crate.

This in particular happened with the 2.5.0 update, which broke a project of mine that used `ureq` with cookies.
My preferred fix would be to expose `CookieStore` and allow using that instead of depending on an external non-synced crate. Note that `Cookie` is already re-exported.

[^1]: technically you can avoid naming `CookieStore`, but it requires some weird hack to make type inference work, so it's something you probably want to avoid.

jsha:
Fixes #548

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -423,7 +423,7 @@ pub use crate::stream::{ReadWrite, TlsConnector};
 #[cfg(feature = "cookies")]
 pub use cookie::Cookie;
 #[cfg(feature = "cookies")]
-pub use cookie::CookieStore;
+pub use cookie_store::CookieStore;
 
 #[cfg(feature = "json")]
 pub use {serde, serde_json};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -422,6 +422,8 @@ pub use crate::stream::{ReadWrite, TlsConnector};
 // re-export
 #[cfg(feature = "cookies")]
 pub use cookie::Cookie;
+#[cfg(feature = "cookies")]
+pub use cookie::CookieStore;
 
 #[cfg(feature = "json")]
 pub use {serde, serde_json};


algesten:
Can we really just export CoolueStore struct? Don't we need to export the whole crate?

jsha:
hm, I think you are right.

algesten (approved):

jsha:
I looked at https://docs.rs/cookie_store/latest/cookie_store/struct.CookieStore.html and I think _almost_ everything you need to make one is available on that type (plus serde and cookie::Cookie).

The one unfortunate exception: https://docs.rs/cookie_store/latest/cookie_store/struct.CookieStore.html#method.from_cookies

This is a pretty useful constructor method that takes `cookie_store::Cookie` instead of the similar `cookie::Cookie`. So in theory we should re-export that too? But the idea of having two re-exported Cookie types seems just too horrible, confusion-wise.

Also should we re-export Url? Or just plan a major release if Url has a major release?

algesten:
Is this going down a rabbit hole we don't want to go down?

Hard to backtrack on it.

algesten:
This would close #678

arnim279:
Do you have any updates on this? [ureq 2.9.5 upgrading its dependencies](https://github.com/algesten/ureq/commit/9d770431cf8c9b15eb7b8e2aed7eee681774df7e) made this issue reappear.

thanhminh-vht:
Any update on this issue?

algesten:
> Any update on this issue?

Yeah. ureq3 is soon done and I will move away from exposing any non-1.x dependency.

algesten:
Closing since this is solved in 3.x (and it will re-export much fewer types).

Issue #570, opened 2022-12-06
Handle chunked transfter encoding with additional trailer


daxhuiberts:
I am talking to a webserver which uses chunked transfer encoding with an additional trailer, similar to what is explained here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer

But I'm getting a "Error while decoding chunks" when requesting it with ureq. It's a bit similar to https://github.com/algesten/ureq/issues/325, but in this case the response is correct. Before the last '\r\n' the trailing headers are added to the response stream, which are then ended with a double set of '\r\n'.

I'm not necessarily interested in the trailing headers,  but I am interested in properly reading the response without getting an error.

From what I've read in https://github.com/algesten/ureq/pull/454 and https://github.com/frewsxcv/rust-chunked-transfer/pull/11 the double '\r\n' handling is done in the rust-chunked-transfer crate, but proper trailing handling is not the concern of the rust-chunked-transfer crate. That means that the rust-chunked-transfer crate should not or allow not to handle the double '\r\n' handling so ureq can process the trailing headers.

I'm not sure what the best approach is here or whether this is something you'd like to handle in ureq. I'd like to pick this up if it's clear what the approach would be.

algesten:
At this point I'm open to forking the upstream project into a local mod or just write our own take. 

The one danger is request smuggling, care must be taken to ensure that's not possible.

jsha:
I'm in favor of accepting chunked responses with trailers, and maybe eventually providing access to them, though the interface would need some thought.

 I happen to have a branch where I already forked rust-chunked-transfer into a local mod, in pursuit of #559. I'll put that up for review now, independent of the #559 changes (which still need some work).

algesten:
ureq 3.x is accepting trailers, but they are skipped. This shouldn't be too hard to fix.

algesten:
The solve is not super easy. But something like `Body::trailers()` to access them, however it's only possible once reaching the end of the body.

Issue #559, opened 2022-11-26
Redesigning PoolReturnRead


jsha:
PoolReturnRead is important because we need to be able to hand something back to the user that (a) implements Read and (b) returns the Stream to the Pool as soon as it is correct to do so. But it relies on a property of the Read interface that is not actually guaranteed and in fact is often not true. It relies on upstream Readers always reading until they get to `Ok(0)`. We [tried to fix this with the `Done` trait](https://github.com/algesten/ureq/pull/509), for the LimitedRead case. But it's come up again in https://github.com/algesten/ureq/issues/549 and https://github.com/algesten/ureq/issues/555, which makes me think we are using the wrong abstraction.

PoolReturnRead only ever wraps two types: `LimitedRead` and `ChunkDecoder`. We can move the "return to pool" logic down a level and make those two types responsible for returning their Stream when it is viable. For instance, `LimitedRead` could have an `Option<R>`, and so long as it has bytes left to read and that `Option` is `Some`, it would read from the inner reader. If any read would leave zero bytes left, it would `take()` the Option and return it to the pool before returning from `read()`. For `ChunkedDecoder` we would need a `BufRead` as the inner reader, so that it could look ahead and see if the next bytes are the end-of-stream indicator, and do the same thing.

The end result of these would be that in almost all cases, the stream would be returned to the pool on the second-to-last read (the one returning `Ok(n)`), not the last read (returning `Ok(0)`). And if the user never calls a read that returns `Ok(0)`, we don't have to care.

algesten:
Sounds good overall.

> The end result of these would be that in almost all cases, the stream would be returned to the pool on the second-to-last read (the one returning Ok(n)), not the last read (returning Ok(0)). And if the user never calls a read that returns Ok(0), we don't have to care.

Doesn't this mean that we could have garbage data after the body? How do we "guarantee" the `Stream` is correct for the next pool reuse?

jsha:
When we return the Stream to the pool, we can check that the `inner` `BufReader`'s buffer is empty.

algesten:
Closing since we're moving to ureq 3.x.

PR #571, opened 2022-12-07
Fork chunked transfer

jsha:
For #559 and #570, it's useful to have our own in-tree chunked transfer module that we can modify as needed.

This is broken into useful commits: the first is a straight copy, followed by a couple of adjustments to work in the new context.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ brotli = ["brotli-decompressor"]
 
 [dependencies]
 base64 = "0.13"
-chunked_transfer = "1.2"
 cookie = { version = "0.16", default-features = false, optional = true}
 once_cell = "1"
 url = "2"
--- a/src/chunked/LICENSE
+++ b/src/chunked/LICENSE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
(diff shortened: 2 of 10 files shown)


algesten (approved):
Yep!

frewsxcv:
Hi! I'm the creator of the chunked-transfer crate. Are there specific features you need in the crate? I'm happy to give you both commit/publish access

lolgesten:
@frewsxcv Hi! 

Since you were busy, we decided to fork, but we rather upstream our changes to benefit the entire community. 

We raised a PR (regarding \r\n in the last chunk), and have another change in our source tree that we could potentially upstream. We also have some thoughts about trailing headers, and I'm a bit unclear on whether that affects the chunked-transfer crate.

Either way, yes please! We would like help out maintain this crate.

jsha:
Indeed! And to add to algesten's comment: I want to make some changes so that Decoder takes a BufRead instead of a Read. This unlocks some work about better stream pooling in ureq. The simplest way to do this is to change the trait bound from Read to BufRead, which of course would be an API break in upstream.

Also I'd like to "fuse" a Decoder so once it reaches the end chunk, further reads are guaranteed to `Ok(0)`.

frewsxcv:
Just added you both as contributors. Feel free to merge whatever PRs and commit directly to `master`

algesten:
Thanks!

PR #572, opened 2022-12-08
Update CHANGELOG for 2.6.0

jsha:
All aboard the 2.6.0 train!

@algesten I'm choosing not to incorporate #568 into this release; I'd like to do a little more research and thinking about what the right thing is to do here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 2.6.0
+
+## Added
+ * Response::remote_addr() (#489)
+ * Request::query_pairs() - make query params from an Iterator of pairs (#519)
+
+## Fixed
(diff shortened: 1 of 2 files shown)


algesten:
@jsha sounds good regarding #568.

Can we land #454? I rebased and added the fix in our local fork.

jsha:
In https://github.com/algesten/ureq/pull/454#issuecomment-1344905604 and https://github.com/algesten/ureq/issues/325#issuecomment-1345186926 it seems like the issue may be already fixed or may have a different fix than we thought. Let's proceed with the release.

jsha:
And published: https://crates.io/crates/ureq/2.6.0

PR #578, opened 2023-01-03
gzip: examine Content-Length header before removing

### LINKED ISSUE #575: actual git version functions into_xxx causes hang on prompt bug
Hello,

i use the actual git version (that solves a problem with the cookiestore for me)  and rust compiles without an error but the code hang on every function call like:

```
let test = httpreq.into_string().unwrap();
or
let json: serde_json::Value = httpreq.into_json().unwrap();

```

He hangs and i must break the compiled program.

jsha:
Fixes #575.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2.6.1
+
+## Fixed
+ * gzip: examine Content-Length header before removing (#578)
+
 # 2.6.0
 
 ## Added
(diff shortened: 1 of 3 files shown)


jsha:
Merging and doing the release myself because this is a pretty significant bug and I'm confident in the fix.

algesten (commented):
Good catch! Thanks for quick turnaround!

Issue #579, opened 2023-01-08
Docs incorrect about serde_derive?


n8henrie:
The docs seem to suggest that one can add / use `serde_derive` as an alternative to `serde` (in my case for the `json` stuff): 

https://github.com/algesten/ureq/blob/032bfb4b276cf133b727dbb2746f558a96e56a15/src/response.rs#L468

> An alternative way to is to use the `serde_derive` crate.
Put this in Cargo.toml: serde_derive = "1"
use serde_derive::{Deserialize};

But it doesn't seem to work. Am I misunderstanding how this is supposed to work?

```console
$ cargo new foo && cd $_
$ cargo add ureq --features=json
$ cargo add serde_derive
$ cat <<'EOF' > src/main.rs
use serde_derive::Deserialize;

#[derive(Deserialize)]
struct Response;

fn main() {
    todo!()
}
EOF
$ cargo build
   Compiling foo v0.1.0 (/private/var/folders/kb/tw_lp_xd2_bbv0hqk4m0bvt80000gn/T/tmp.8KxiUrC2C9/foo)
error[E0463]: can't find crate for `serde`
 --> src/main.rs:3:10
  |
3 | #[derive(Deserialize)]
  |          ^^^^^^^^^^^ can't find crate
  |
  = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0463`.
error: could not compile `foo` due to previous error
```

I may be just using it wrong, but I'd be happy to submit a documentation PR if anything needs updating.

jsha:
Thanks for the report! I honestly don't know why that error happens, since serde_derive should pull in serde transitively. But it looks like the documented way to use serde's derive macros is by importing them from serde, not serde_derive:

https://serde.rs/derive.html

Probably we should remove the mention of serde_derive. Thoughts @algesten ?

algesten:
@jsha Yes. `serde_derive` was "the old way", which I think still worked. However increasingly people (including myself) are using only `serde` with the feature `derive` turned on. We should change the docs to reflect this.

n8henrie:
Figured something out -- this works (without depending on `serde` itself):

```rust
use serde_derive::Deserialize;

#[derive(Deserialize)]
#[serde(crate = "ureq::serde")]
struct Response;

fn main() {
    todo!()
}
```

Does that PR for the docs look okay?

PR #581, opened 2023-01-09
Fix docs for serde_derive

n8henrie:
If only using `serde_derive` and not `serde` this is required for each
use of `#[derive(Deserialize)]`.

Fixes https://github.com/algesten/ureq/issues/579

See also:
- https://github.com/serde-rs/serde/issues/856
- https://github.com/serde-rs/serde/issues/1465

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -465,10 +465,6 @@ impl Response {
     /// // Put this in Cargo.toml: serde = { version = "1", features = ["derive"] }
     /// use serde::{Deserialize};
     ///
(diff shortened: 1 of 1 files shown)


jsha:
Thanks for the PR! Per #579 I think we'd rather remove the mention of serde_derive entirely. Are you interested in doing that?

jsha (approved):

jsha:
Thanks!

n8henrie:
Rebased to drop the prior commit

algesten (approved):
Thanks!

PR #583, opened 2023-01-16
Allow non-empty connection pools to be dropped

### LINKED ISSUE #582: Socket leak in v2.6
I'm observing that sockets are not being closed in the latest version. After making a lot of requests, you will run out of file descriptors and your process will fail.

jsha:
This breaks a reference cycle between PoolReturner and Agent that was causing Agents (and their contained ConnectionPool) to never be dropped so long as there was any stream in the ConnectionPool. This cause sockets to leak over time, particularly when the convenience functions ureq::get(), ureq::post(), etc were used, since those functions create a new Agent each time.

Fixes #582.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2.6.2
+
+## Fixed
+ * Non-empty connection pools were never dropped (#NNN)
+
 # 2.6.1
 
 ## Fixed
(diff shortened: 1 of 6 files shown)


algesten (approved):
Excellent analysis in #582 and fix. Thank you!

keyvank:
Wow, thanks for the quick fix!

jsha:
Published as version 2.6.2.

PR #587, opened 2023-01-30
Proxy: Only use HTTP CONNECT for HTTPS requests

lwagner94:
Previously, `ureq` used `HTTP CONNECT` for *all* requests, including plain-text HTTP requests. However, some proxy servers like Squid only allow tunneling via the `HTTP CONNECT` method on port 443 - since it is usually only used to proxy HTTPS requests. As a result, it was not possible to use `ureq` with a Squid server in its default configuration.

With the changes from this commit, ureq will interact with HTTP proxies in a more standard-conforming way, where `CONNECT` is only used for HTTPS requests. HTTP request paths are transformed in such a way that they comply with RFC 7230 [1].

Tested against Squid 4.13 in standard configuration, with and without basic authentication, for HTTP and HTTPS requests.

[1] https://www.rfc-editor.org/rfc/rfc7230#section-5.3.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,14 +198,14 @@ to encode the request body using that.
 
 ## Proxying
 
-ureq supports two kinds of proxies,  HTTP [`CONNECT`], [`SOCKS4`] and [`SOCKS5`], the former is
+ureq supports two kinds of proxies,  [`HTTP`], [`SOCKS4`] and [`SOCKS5`], the former is
 always available while the latter must be enabled using the feature
 `ureq = { version = "*", features = ["socks-proxy"] }`.
 
 Proxies settings are configured on an [Agent] (using [AgentBuilder]). All request sent
 through the agent will be proxied.
 
-### Example using HTTP CONNECT
+### Example using HTTP
 
 ```rust
 fn proxy_example_1() -> std::result::Result<(), ureq::Error> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -224,14 +224,14 @@
 //!
 //! # Proxying
 //!
-//! ureq supports two kinds of proxies,  HTTP [`CONNECT`], [`SOCKS4`] and [`SOCKS5`], the former is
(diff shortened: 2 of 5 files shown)


algesten:
Hi there!

Thanks for taking the time to do this. I know very little about HTTP proxies, so I am very grateful you are contributing here.

Is there a risk this is a breaking change for anyone? Would that mean they're relying on non-spec compliant stuff?

lwagner94:
Hi!
Thanks so much for taking the time to look into my PR.

I believe the chances of breaking any client code are rather slim. Breakage should only occur in case a proxy *only* supports tunneling HTTP CONNECT. I don't think that this is something worth considering, since setups like this would also break *a lot* of other HTTP clients. curl seems to have the '-p/--proxy-tunnel' option, which makes it behave the same way as ureq currently does, while other popular HTTP client libraries (e.g. reqwest, requests) do not seem to have a mechanism like this. 

To be on the safe side, we could add a configuration option, similar to curl - but I would make it opt-in rather than opt-out, so that it still behaves the same as you would expect by default.

puffi:
I've talked with @lwagner94 a bit about this, and it seems the RFC is rather vague regarding the HTTP CONNECT for HTTPS only.
All the RFC [0] mentions is that it is a way to create a tunnel where the TCP connection is passed through. In theory this should be possible for HTTP as well, but in practice this doesn't seem to be the case according to Lukas.

So in theory this may have implications regarding some use cases, but in practice it should not since CURL and other popular libraries only seem to use it for HTTPS connections.


[0] https://www.rfc-editor.org/rfc/rfc7231#section-4.3.6

algesten:
Cool. I trust you. If you rebase off main, that lint error will go away. Then I can merge.

lwagner94:
Rebased to current main.

lwagner94:
Interesting, the failing test (executed with `cargo test --no-default-features --features native-tls json`) seems to work fine for me locally.

algesten:
Weird. Hm. We're getting similar errors in main. Let's merge and fix main with follow-up PR.

Issue #589, opened 2023-02-07
Customizable `Response`s for middlewares


kade-robertson:
Hello!

This is partially related to https://github.com/algesten/ureq/issues/363 in terms of my personal use case, but I wanted this to be more of a generic request.

Currently, it isn't _really_ possible to construct anything but a basic, text-only `Response` object (as in, one not generated by making a request). I've run into this limitation recently on one of my projects -- I've implementing a caching middleware [here](https://github.com/kade-robertson/ddragon/blob/34012e522e615c3669d47c583899effc3a80231d/src/cache_middleware.rs#L22) as a very basic replacement for the functionality I can get from something like [http-cache-reqwest](https://github.com/06chaynes/http-cache). This generally works okay for UTF-8 data such as serialized JSON responses like I primarily deal with, but falls apart when downloading images. I ended up having to add an escape hatch and a custom caching wrapper around my request to deal with that particular instance.

The current "faking" mechanism is _almost_ what I'd expect, but the missing piece here is accepting a `&[u8]` content body and (possibly) a more structured way of specifying headers like `Content-Type`. 
- Maybe this doesn't make sense to overload `Response::new` with and rather a `ResponseBuilder` or some `.set_body(&mut self)`-like method would be the way to go. 
- Maybe `Response` and `Request` can `#[derive(Serialize, Deserialize)]`?

This might be leaning towards making this project more complex then it aims to be, but I think given that a middleware trait already exists, it makes sense to flesh this out a bit more than it is currently.

kade-robertson:
Looking at this a bit more, maybe what makes the most sense is using [`http::Response`](https://docs.rs/http/latest/http/response/struct.Response.html) as the tool for constructing the response, and have this crate's `Response` implement `From<http::Response<T>>`. 
- It seems reasonably thorough and provides the level of customization I'd be looking for I think
- Relatively lightweight in terms of build time impact
- Doesn't pull in too many additional dependencies
- MSRV of the current version (`0.2.8`) is below what _seems_ to be the MSRV of this version (1.53 or 1.56 based on the MSRV issue here)
- Reasonably easy to put behind a feature if this isn't desired to be included by default

algesten:
Hi! Thanks for thinking about this!

I think an optional dependency on the http crate would be good. Something that both allows conversion in both directions for both `Request` and `Response` (if possible, I haven't thought about this deeply. I.e. we would provide:

```rs
impl<T: what bound?!> From<http::Request<T>> for Request
[...]

PR #591, opened 2023-02-13
feat: support interop between ureq and http

kade-robertson:
Preliminary look at some of the ideas discussed in https://github.com/algesten/ureq/issues/589. Took some inspiration from the layout of https://github.com/algesten/ureq/pull/520 since that covered a similar thing.

## General issues

`http::Response` has no `Content-Type` or `Content-Length` validation.

Nearly everything here calls some fallible method. In these instances, they are either `.unwrap()`ed, or replaced with a sane default (or skipped, if applicable). There's a decent argument to be made that we should fail if any of these instances occur, meaning we'd want to implement `TryFrom` instead. 

## Response

This works _relatively_ well. I chose `AsRef<[u8]> + Send + Sync + 'static` as this works automatically for `Vec<u8>` and `Strings`, and most body types can be relatively easily converted into one of those two. It also makes it fairly easy to package into the type `Response.reader` is expected to be. In the other direction, a `Response` can be converted into an `http:Response` with a body that is either directly from `.into_reader()` or `into_string()`. Maybe worth considering one for `Vec<u8>`?

Notably, pretty much all of the `Response` properties are now `pub(crate)`. We don't really have a builder syntax here so without fleshing out something like a `ResponseBuilder`, this is the change that touches the least of the project and doesn't expose anything to the outside world.

## Request

This ends up being a bit awkward. It isn't really possible to convert a `Request` to `http::Request` in either direction, because `Request` does have a builder pattern _and_ doesn't know its body until a `.send_` method is called, which is returning a `Result<Response>`. The closest analog ends up being `http::request::Builder` -- I'd argue this isn't even a valuable implementation to have.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,7 @@ jobs:
           - native-certs
           - gzip
           - brotli
+          - http
     env:
       RUST_BACKTRACE: "1"
       RUSTFLAGS: "-D dead_code -D unused-variables -D unused"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,7 +58,7 @@ jobs:
           - native-certs
           - gzip
           - brotli
-          - http
+          - http-interop
     env:
       RUST_BACKTRACE: "1"
       RUSTFLAGS: "-D dead_code -D unused-variables -D unused"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ cookies = ["cookie", "cookie_store"]
 socks-proxy = ["socks"]
 gzip = ["flate2"]
 brotli = ["brotli-decompressor"]
+http = ["dep:http"]
 
 [dependencies]
 base64 = "0.21"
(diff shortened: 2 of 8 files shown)


algesten (commented):
Overall I think this looks good. Thanks!

We should probably have test cases for all the conversions, which could potentially be rustdoc tests. I'm wondering how we document this.

kade-robertson:
> Overall I think this looks good. Thanks!
> 
> We should probably have test cases for all the conversions, which could potentially be rustdoc tests. I'm wondering how we document this.

I've gone through and done that, though I'm not sure if we want to be more thorough or what else to include.

algesten:
Let's do it.

algesten:
Thank you for all the work on this!

MarijnS95:
> ## Request
> This ends up being a bit awkward. It isn't really possible to convert a `Request` to `http::Request` in either direction, because `Request` does have a builder pattern _and_ doesn't know its body until a `.send_` method is called, which is returning a `Result<Response>`. The closest analog ends up being `http::request::Builder` -- I'd argue this isn't even a valuable implementation to have.

@kade-robertson making you aware of #656: `into_parts()` exists to split the body from the "request parts", perhaps that is sensible to implement?

Issue #602, opened 2023-03-30
overall_timeout_reading_json fails with serde_json 1.0.94


jsha:
We have a unittest failing in CI. This was initially confusing because it was not failing for me locally. I realized I had a Cargo.lock locally with older versions of various dependencies. Removing the Cargo.lock caused the test failure to reproduce. I wound up tracking it down to serde_json 1.0.94. There seems to have been a change so our code that downcasts errors from serde's Error::source no longer works.

joshstoik1:
For reference: https://github.com/serde-rs/json/issues/1012

PR #603, opened 2023-03-30
lock serde_json to 1.0.93

jsha:
This avoids a bug in more recent version that prevents us from downcasting errors to see if they are TimedOut errors.

Quick fix for #602 while we look for a root cause.


Cargo.toml
```diff
@@ -33,7 +33,7 @@ once_cell = "1"
 url = "2"
 socks = { version = "0.3", optional = true }
 serde = { version = "1", optional = true }
-serde_json = { version = "1", optional = true }
+serde_json = { version = "=1.0.93", optional = true }
```
algesten, line 36: Wonder if this could be a less than range to not force users to a specific version?

algesten (commented):
Do we know if serde considers it a bug?

jsha:
> Do we know if serde considers it a bug?

Dunno! I haven't had time to track it into serde yet, just wanted to get our CI passing again.

algesten (approved):

PR #605, opened 2023-03-30
feat: expose `local_addr` from `Response`

kubuzetto:
Expanding on #591 which adds `remote_addr` to `Stream` and `Response`; this PR also exposes the local address through the `Response`.

Retrieving `local_addr` this way can be helpful when the user needs to know the client port which is selected arbitrarily from the ephemeral port range. Similarly, if the client is running on a device with multiple NICs that can route to the remote address; one might want to know from which local IP address the request was sent. (In [our](https://github.com/picussecurity) case, we'd like to to record the local IP and port to track them in WAF logs).

(This is my first contribution to the repo; and I couldn't see a contribution guideline. If there is a process that I missed I'd be happy to follow it)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -568,8 +568,8 @@ mod tests {
 
         let resp = agent.get("http://www.google.com/").call().unwrap();
         assert_eq!(
-            "text/html; charset=ISO-8859-1",
-            resp.header("content-type").unwrap()
+            "text/html;charset=ISO-8859-1",
+            resp.header("content-type").unwrap().replace("; ", ";")
         );
         assert_eq!("text/html", resp.content_type());
     }
@@ -581,8 +581,8 @@ mod tests {
 
         let resp = agent.get("https://www.google.com/").call().unwrap();
         assert_eq!(
-            "text/html; charset=ISO-8859-1",
-            resp.header("content-type").unwrap()
+            "text/html;charset=ISO-8859-1",
+            resp.header("content-type").unwrap().replace("; ", ";")
         );
         assert_eq!("text/html", resp.content_type());
     }
@@ -597,8 +597,8 @@ mod tests {
 
         let resp = agent.get("https://www.google.com/").call().unwrap();
         assert_eq!(
-            "text/html; charset=ISO-8859-1",
-            resp.header("content-type").unwrap()
+            "text/html;charset=ISO-8859-1",
+            resp.header("content-type").unwrap().replace("; ", ";")
         );
         assert_eq!("text/html", resp.content_type());
     }
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -78,7 +78,6 @@ pub(crate) struct AgentConfig {
     pub redirect_auth_headers: RedirectAuthHeaders,
     pub user_agent: String,
     pub tls_config: TlsConfig,
-    pub get_local_addr: bool,
 }
 
 /// Agents keep state between requests.
@@ -263,7 +262,6 @@ impl AgentBuilder {
                 redirect_auth_headers: RedirectAuthHeaders::Never,
                 user_agent: format!("ureq/{}", env!("CARGO_PKG_VERSION")),
                 tls_config: TlsConfig(crate::default_tls_config()),
-                get_local_addr: false,
             },
             max_idle_connections: DEFAULT_MAX_IDLE_CONNECTIONS,
             max_idle_connections_per_host: DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST,
@@ -619,30 +617,6 @@ impl AgentBuilder {
     /// ```
     pub fn tls_connector<T: TlsConnector + 'static>(mut self, tls_config: Arc<T>) -> Self {
         self.config.tls_config = TlsConfig(tls_config);
-        self
-    }
-
-    /// Whether local_addr will be read from the tcp socket.
-    ///
-    /// local_addr is the address from which the client sends the request.
-    /// Reading it invokes a syscall, therefore it is disabled by default.
-    ///
-    /// Defaults to false.
-    ///
-    /// ```
-    /// # fn main() -> Result<(), ureq::Error> {
-    /// # use std::net::SocketAddr;
-    /// # ureq::is_test(true);
-    /// let agent = ureq::builder()
-    ///     .get_local_addr(true)
-    ///     .build();
-    /// let response = agent.get("http://httpbin.org/get").call()?;
-    /// assert!(response.local_addr().is_some());
-    /// # Ok(())
-    /// # }
-    /// ```
-    pub fn get_local_addr(mut self, get_local_addr: bool) -> Self {
-        self.config.get_local_addr = get_local_addr;
         self
     }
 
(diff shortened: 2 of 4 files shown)


algesten (commented):
Hi! Welcome to ureq!

Overall this seems like a good idea. I just need to read the PR with fresh eyes tomorrow since it's a bit late here. 

Thanks!

jsha:
This generally looks good. One small tweak: for remote_addr, we chose to store it on the Response because it was "free" - we automatically know the remote addr because we connected to it.

local_addr isn't free. We need to make a syscall to get it. Not a huge deal, but given that checking the local_addr is very very rare, I'd prefer to avoid incurring that small cost on every request. Instead, I'd like to fetch it only when the local_addr accessor is called. That means the local_addr accessor will have a different signature than remote_addr. It will return a Result. But that's `Ok` by me; they are different because they do different things.

kubuzetto:
That'a really good point; I'll try to make the changes tonight. Rather than the return type, I guess the biggest change will be the type of the `local_addr` field in both `Stream` and `Response`. A lazy-initialized type which is passed the TCPStream's `local_addr()` as callback maybe? I'll tinker with it for a while.

> It will return a `Result`. But that's `Ok` by me

Ha, saw what you did there 😄

jsha:
> I guess the biggest change will be the type of the local_addr field in both Stream and Response. A lazy-initialized type which is passed the TCPStream's local_addr() as callback maybe? I'll tinker with it for a while.

I think there's no need to make it a stored field; we can fetch and return it fresh on each call.

kubuzetto:
Hmm. This turned out to be more complicated than I anticipated.

`local_addr` function belongs to the `TcpStream`, and passing it as a `Fn() -> io::Result<SocketAddr>` is bound to fail at some point due to lifetimes. I tried to trick the borrow checker by adding an `AddrResolver` trait (which `TcpStream`, `Stream`, `Box<Stream>`,... implements), given that `Stream` is passed to the `Resolver`, but `Resolver`s constructor consumes the `Stream` at some point in favor of some `Box<dyn Read + ...>`, so no matter what I tried it turned into a dead end. In either case it was becoming very fragile, as many intermediate types had to implement AddrResolver for no apparent reason.

An alternative I'm considering is this: a connection option that says "I want you to record the `local_addr` value", according to which we decide to invoke the syscall and fill a `local_addr: Option<SocketAddr>` field. Do you think it makes sense? Otherwise can you give me some pointers on how to proceed with this?

algesten:
Sorry. I've been really busy at work. I hope to  circle around to this coming weekend.

algesten:
@kubuzetto I pushed to your branch a simpler version of this idea. The `TcpStream` is available in the `Response`, so we don't need to pass the `local_addr` around.

kubuzetto:
LGTM, thank you!

algesten:
@jsha you ok with adding this config on to agent?

I tried fishing this out of the depths of the wrapped body, but there's so many levels of LimitReader, DeadlineRead etc that all work of `dyn Read`, which means fishing the `TcpStream` out is a rather big change. Similarly I looked at cloning the `TcpStream` before we giving it up to the body reader wrapper, only to realize that cloning does a `fcntl` – which means we're back to an extra syscall.

If you're happy, we can merge.

jsha:
Hm, I really don't like exposing an `Agent::get_local_addr` config setting to control getting this. It would be much nicer if all the user had to do was fetch the address when they want it.

It occurred to me to check how other systems do this. So for instance I found how Go does it: https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/net/sock_posix.go;l=116-175

It turns out Go fetches the local and remote addresses from every outbound connection (and also from inbound connections but that's not relevant to us):

```go
	// Record the local and remote addresses from the actual socket.
	// Get the local address by calling Getsockname.
	// For the remote address, use
	// 1) the one returned by the connect method, if any; or
	// 2) the one from Getpeername, if it succeeds; or
	// 3) the one passed to us as the raddr parameter.
	lsa, _ = syscall.Getsockname(fd.pfd.Sysfd)
```

Also hyper offers a [local_addr](https://docs.rs/hyper/latest/hyper/client/connect/struct.HttpInfo.html#method.local_addr) accessor with no error values. I haven't dug in to see how it is implemented but probably similar?

Also skimming the Go code it looks like there is a possibility that "remote address we actually connected to" is different from "remote address from the user," which may also be part of why it's worthwhile to call Getsockname to find out both the remote and local addr for the socket.

So in short I think I was being too conservative in suggesting we avoid the syscall above; other systems that are pickier about performance than we are choose to do so, so presumably it is worthwhile. I'm sorry you both invested time in trying to find ways to avoid it, unnecessarily! I think the earlier concept, where we get and store the local addr at connection time, was on the right track.

kubuzetto:
I made some changes so that the `local_addr` syscall is invoked unconditionally. I kept the `Option<SocketAddr>` return type so that we don't have to pass a made-up value for it in `http_interop.rs`, and handle it in the call site. I also pushed a trivial fix to a flaky test (`connect_http_google`).

> I'm sorry you both invested time in trying to find ways to avoid it, unnecessarily

No problem at all, I really enjoyed working on both solutions 🙂

src/lib.rs
```diff
@@ -568,8 +568,8 @@ mod tests {
 
         let resp = agent.get("http://www.google.com/").call().unwrap();
         assert_eq!(
-            "text/html; charset=ISO-8859-1",
-            resp.header("content-type").unwrap()
+            "text/html;charset=ISO-8859-1",
+            resp.header("content-type").unwrap().replace("; ", ";")
```
algesten, line 572: What's going on here?

algesten (commented):
Overall looks good, but what are these changes to the tests?

src/lib.rs
kubuzetto: Response sometimes contains `text/html;charset=ISO-8859-1`, sometimes `text/html; charset=ISO-8859-1` (note the space after the semicolon); making the test flaky.

jsha:
Thanks for the updates! Let's remove the Option from `local_addr()`. I'd like to keep the symmetry between `remote_addr()` and `local_addr()`. I agree it's ugly to make up a fake IP address in http_interop.rs, but I think it's the best we can do. The `http` crate doesn't have any notion of `local_addr()` and `remote_addr()`. Hyper does, and stores those in the `extensions` of an `http::Response`. But we can't pull them out without depending on Hyper. It would be a shame to make our API worse just for the sake of a missing transform to another crate's types.

The test updates look fine to me. I'd like to independently update our tests so none of them rely on external Internet resources, since that makes them flaky.

jsha:
Oh, also: we should not call `.unwrap()` when getting local_addr from the stream. Instead we should propagate the error.

kubuzetto:
Removed the `Option`; handled the `unwrap` with `map_err` instead

algesten (approved):
This looks good to go!

Thanks!

algesten:
After fixing the lint ;)

kubuzetto:
Lint is fixed, one of the flaky tests is failing

kubuzetto:
Hi @algesten, is there anything I should be doing for this PR?

algesten (approved):

PR #608, opened 2023-04-03
Fix cookie leak on redirect

amoskvin:
This removes the cookie header on redirect, otherwise if there are multiple redirects, a new cookie header gets appended each time, while the existing one remains. This could result in the cookies being leaked to a third party.


amoskvin:
ping @jsha @algesten

jsha (approved):
Thanks for spotting, and thanks for the fix! I'll prep a release with this and some other recent changes.

jsha:
This is released in 2.7.0.

PR #610, opened 2023-04-04
fix: cannot use http_proxy connect non-standard http ports

stelofan:
Previously, when making http proxy plain-text HTTP requests, the port was omitted from the proxy request target, which prevented connecting to non-standard http ports.

if the request url contains a port, This PR add this port to the proxy request target.


algesten (approved):

algesten:
LGTM! Thanks!

PR #617, opened 2023-04-28
feat: set headers on agent level.

yonson2:
This PR enables setting `Request` headers on an agent level so that they will be used on all requests made by that agent. Its a small utility function that saves some boilerplate in case all requests need a custom header (like a `Bearer <TOKEN>` for authentication).

```rust
use ureq::{Agent, AgentBuilder};

let agent: Agent = AgentBuilder::new()
    .set("Accept", "text/plain")
    .build();

let body: String = agent.get("http://example.com/page").call()?.into_string()?;

```

I don't know if this is something that you would like to have on `ureq` but it is useful in my use-case.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use url::Url;
 
+use crate::header::{self, Header};
 use crate::middleware::Middleware;
 use crate::pool::ConnectionPool;
 use crate::proxy::Proxy;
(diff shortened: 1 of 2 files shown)


algesten:
Hi! Welcome to ureq!

I think we used to have this function in ureq 1.x, but we removed it for 2.x. I'm searching the issues, but can't for my life find the reason why we dropped it. @jsha do you remember?

algesten:

PR #618, opened 2023-04-29
Allow a connection attempt to timeout and keep trying resolved addresses

ex-nihil:
Before this change `connect_host` only attempted to connect to the first resolved SocketAddr.

My interpretation was that this was the indented use of `agent.config.timeout_connect` but, I'm unsure as the default value seems to be 30 sec.

Breaking change for any projects depending on the `timeout_connect` config, maybe introduce a separate configuration for this timeout.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -2,6 +2,7 @@ use log::debug;
 use std::io::{self, BufRead, BufReader, Read, Write};
 use std::net::SocketAddr;
 use std::net::TcpStream;
+use std::ops::Div;
 use std::time::Duration;
 use std::time::Instant;
 use std::{fmt, io::Cursor};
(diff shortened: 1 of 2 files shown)


algesten:
The intention is that `connect_timeout` is regardless of how many `SocketAddr` there might be. 

Is your problem that the failing socket address takes longer than 30 seconds to fail?

algesten:
> edit: My first address never responds

As far as I know all OSes will timeout socket connection attempts _at some point_. Some quick googling seems to say for windows it's 72 seconds and for Linux 60 seconds. In many cases I would image the timeout to be shorter. For example, if the host is there, but port closed, the SYN would be responded with an RST and the fail be instant.

The 30 second value is for the overall the connection attempt, not per socket addr. On your host, could you try lowering the OS default timeout and see if that helps?

I wonder what cURL does here, does it fail faster than the OS so it can try more addresses?

ex-nihil:
> I wonder what cURL does here, does it fail faster than the OS so it can try more addresses?
Curl failed over pretty much instantly when it was happening.

It seems the Spotify server is no longer non-responsive for the first host, so unable to recreate right now.
I will try to setup a test the behavior later tonight.

ex-nihil:
Added a testcase to this PR `socket_addr_fail_over` recreating it. Your suggestion with lowering the OS timeout should absolutely work, but the question is how curl was able to figure it out. 

I would need to setup a DNS server to compare.

ex-nihil:
Well, found why curl worked.
https://github.com/curl/curl/blob/faebcee349d45f46638add6dda3185425c7afac6/lib/connect.c#L752

algesten:
Right. That seems reasonable. We could do that two. _if there are at least two addresses_, split `connect_timeout` in 2 and try both after each other.

algesten:
Let's land this.

Thanks!

PR #620, opened 2023-05-04
fix #557: read proxy response until \r\n\r\n

### LINKED ISSUE #557: HTTPConnect proxy doesn't read response until second \r\n\r\n as required by rfc2068
When using tinyproxy as proxy and another tinyproxy instance as upstream proxy, the proxy returns `Via: 1.1 tinyproxy (tinyproxy/1.10.0)` in addition to `HTTP/1.0 200 Connection established`. This trips up ureq's TLS handling since ureq doesn't read until \r\n\r\n, but just until the buffer is not fully filled on read. So once it sees the `Connection established` message it will verify the response and continue with the next step (TLS, if it connects to a HTTPS site).
This leads to the following (strace output):
```
30071 17:06:59.803841 connect(5<TCP:[60677]>, {sa_family=AF_INET, sin_port=htons(8888), sin_addr=inet_addr("192.168.22.112")}, 16) = -1 EINPROGRESS (Operation now in progress) <0.000039>
30071 17:06:59.803973 ioctl(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, FIONBIO, [0]) = 0 <0.000007>
30071 17:06:59.804072 poll([{fd=5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, events=POLLOUT}], 1, 29999) = 1 ([{fd=5, revents=POLLOUT}]) <0.000007>
30071 17:06:59.804099 setsockopt(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, SOL_SOCKET, SO_RCVTIMEO_OLD, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0", 16) = 0 <0.000006>
30071 17:06:59.804120 setsockopt(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, SOL_SOCKET, SO_SNDTIMEO_OLD, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0", 16) = 0 <0.000005>
30071 17:06:59.804141 sendto(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, "CONNECT example.com:443 HTTP/1.1\r\nHost: example.com.com:443\r\nUser-Agent: something/1.0.0\r\nProxy-Connection: Keep-Alive\r\n\r\n", 128, MSG_NOSIGNAL
, NULL, 0) = 128 <0.000013>
30071 17:06:59.804176 recvfrom(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, "HTTP/1.0 200 Connection established\r\n", 256, 0, NULL, NULL) = 37 <0.043128>
30071 17:06:59.847368 getrandom("\xfd", 1, 0) = 1 <0.000007>
30071 17:06:59.847400 getrandom("\xc0\x73\x83\x2b\x08\xc2\xfd\x27\x89\xbb\x10\x28\x80\xc0\xd8\xd6\x5f\x4a\x00\x6c\x73\xa9\x97\xd4\x0f\x58\x87\xf9\x70\x46\xdc\xbf", 32, 0) = 32 <0.000006>
30071 17:06:59.847458 getrandom("\x19\xce\xaf\x07\xcf\xad\x1f\xae\x3e\x55\x2a\x6c\x18\xce\x97\x24\x79\x49\xc2\xc6\xbf\xa2\x69\x09\x94\x8d\xd3\x5b\xb6\x6b\xaf\x87", 32, 0) = 32 <0.000007>
30071 17:06:59.847479 getrandom("\x0a\x24\x7d\x39\x08\xec\xe9\x60\x17\xa0\xf4\x2e\x3a\x4f\x8f\x1d\xd5\xd1\x00\x04\x61\xd0\x45\x11\xa1\x81\xd5\x0f\xb2\x83\xd5\x00", 32, 0) = 32 <0.000006>
30071 17:06:59.847559 writev(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, [{iov_base="\26\3\1\0\362\1\0\0\356\3\3\n$}9\10\354\351`\27\240\364.:O\217\35\325\321\0\4a\320E\21\241\201\325\17\262\203\325\0 \31\316\257\7\317\255\37\256>U*
l\30\316\227$yI\302\306\277\242i\t\224\215\323[\266k\257\207\0\24\23\2\23\1\23\3\300,\300+\314\251\3000\300/\314\250\0\377\1\0\0\221\0+\0\5\4\3\4\3\3\0\0\0\25\0\23\0\0\20example.com\0\v\0\2\1\0\0\n\0\10\0\6\0\35\0\27\0\30\0\r\0\24\0\22
\5\3\4\3\10\7\10\6\10\5\10\4\6\1\5\1\4\1\0\27\0\0\0\5\0\5\1\0\0\0\0\0\22\0\0\0003\0&\0$\0\35\0 M\371\265BMbI\336\353\233\27\234\213\324\346\33c\t\\\324\260w\371m\321\2049\370\277.\36L\0-\0\2\1\1\0#\0\0", iov_len=247}], 1) = 247 <0.000013>
30071 17:06:59.847594 recvfrom(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, "Via: 1.1 tinyproxy (tinyproxy/1.10.0)\r\nProxy-agent: tinyproxy/1.10.0\r\n\r\n", 18437, 0, NULL, NULL) = 72 <0.000007>
30071 17:06:59.847624 close(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>) = 0 <0.000017>
```
The gist is, that it starts the TLS step:
```
30071 17:06:59.847559 writev(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, [{iov_base="\26\3\1\0\362\1\0\0\356\3\3\n$}9\10\354\351`\27\240\364.:O\217\35\325\321\0\4a\320E\21\241\201\325\17\262\203\325\0 \31\316\257\7\317\255\37\256>U*
l\30\316\227$yI\302\306\277\242i\t\224\215\323[\266k\257\207\0\24\23\2\23\1\23\3\300,\300+\314\251\3000\300/\314\250\0\377\1\0\0\221\0+\0\5\4\3\4\3\3\0\0\0\25\0\23\0\0\20example.com\0\v\0\2\1\0\0\n\0\10\0\6\0\35\0\27\0\30\0\r\0\24\0\22
\5\3\4\3\10\7\10\6\10\5\10\4\6\1\5\1\4\1\0\27\0\0\0\5\0\5\1\0\0\0\0\0\22\0\0\0003\0&\0$\0\35\0 M\371\265BMbI\336\353\233\27\234\213\324\346\33c\t\\\324\260w\371m\321\2049\370\277.\36L\0-\0\2\1\1\0#\0\0", iov_len=247}], 1) = 247 <0.000013>
```
And then still receives something from the proxy:
```
30071 17:06:59.847594 recvfrom(5<TCP:[192.168.22.111:51328->192.168.22.112:8888]>, "Via: 1.1 tinyproxy (tinyproxy/1.10.0)\r\nProxy-agent: tinyproxy/1.10.0\r\n\r\n", 18437, 0, NULL, NULL) = 72 <0.000007>
```
This then trips it up since it expected a response from the TLS request.

Based on the rfc2068 there should be a CRLF after every header, and once the headers are finished, another CRLF before the body. So in total 2x CRLF at the end [0], but ureq doesn't seem to read until it gets 2 CRLFs right after another.

[0] https://www.ietf.org/rfc/rfc2068.txt (4.1 - Message Types, 4.2 - Message Headers, 6.1 - Status Line)

puffi:
Headers are terminated by CRLF according to the spec [0]. Between headers and the body there's an additional CRLF. This means we have to parse until we get 2 CRLFs right after each other to know when we can stop reading from the stream.

To do that we read byte by byte and try matching the last 4 bytes when we receive an LF ('\n'). That seems similar to the way curl does it [1].

The response parsing doesn't have to be changed since only the status code is of interest which always comes first [0].

[0] https://www.ietf.org/rfc/rfc2068.txt (6 Response)
[1] https://github.com/curl/curl/blob/89f6fafedbb57b62fddf58223cfa92307bf51a68/lib/cf-h1-proxy.c#L476

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -14,6 +14,7 @@ use crate::error::ErrorKind;
 use crate::pool::{PoolKey, PoolReturner};
 use crate::proxy::Proxy;
 use crate::unit::Unit;
+use crate::Response;
 use crate::{error::Error, proxy::Proto};
 
 /// Trait for things implementing [std::io::Read] + [std::io::Write]. Used in [TlsConnector].
@@ -433,18 +434,12 @@ pub(crate) fn connect_host(
             .unwrap();
             stream.flush()?;
 
-            let mut proxy_response = Vec::new();
-
-            loop {
-                let mut buf = vec![0; 256];
-                let total = stream.read(&mut buf)?;
-                proxy_response.append(&mut buf);
-                if total < 256 {
-                    break;
-                }
-            }
-
-            Proxy::verify_response(&proxy_response)?;
+            let s = stream.try_clone()?;
+            let pool_key = PoolKey::from_parts(unit.url.scheme(), hostname, port);
+            let pool_returner = PoolReturner::new(&unit.agent, pool_key);
+            let s = Stream::new(s, remote_addr, pool_returner);
+            let response = Response::do_from_stream(s, unit.clone())?;
+            Proxy::verify_response(&response)?;
         }
     }
 
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -1,6 +1,9 @@
 use base64::{prelude::BASE64_STANDARD, Engine};
 
-use crate::error::{Error, ErrorKind};
+use crate::{
+    error::{Error, ErrorKind},
+    Response,
+};
 
 /// Proxy protocol
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
@@ -163,20 +166,10 @@ Proxy-Connection: Keep-Alive\r\n\
         )
     }
 
-    pub(crate) fn verify_response(response: &[u8]) -> Result<(), Error> {
-        let response_string = String::from_utf8_lossy(response);
-        let top_line = response_string
-            .lines()
-            .next()
-            .ok_or_else(|| ErrorKind::ProxyConnect.new())?;
-        let status_code = top_line
-            .split_whitespace()
-            .nth(1)
-            .ok_or_else(|| ErrorKind::ProxyConnect.new())?;
-
-        match status_code {
-            "200" => Ok(()),
-            "401" | "407" => Err(ErrorKind::ProxyUnauthorized.new()),
+    pub(crate) fn verify_response(response: &Response) -> Result<(), Error> {
+        match response.status() {
+            200 => Ok(()),
+            401 | 407 => Err(ErrorKind::ProxyUnauthorized.new()),
             _ => Err(ErrorKind::ProxyConnect.new()),
         }
     }


src/stream.rs
```diff
@@ -435,13 +435,21 @@ pub(crate) fn connect_host(
 
             let mut proxy_response = Vec::new();
 
-            loop {
-                let mut buf = vec![0; 256];
-                let total = stream.read(&mut buf)?;
-                proxy_response.append(&mut buf);
-                if total < 256 {
+            let mut buf = [0u8; 1];
+            while let Ok(total) = stream.read(&mut buf) {
```
algesten, line 439: The `?` on `read` in previous version propagated the IO error out. This stops the loop on `Err`, silently discards it, which is probably not what we want. I suggest keep using the `loop + read` variant.

src/stream.rs
```diff
@@ -435,13 +435,21 @@ pub(crate) fn connect_host(
...
+                if total == 0 {
+                    // no more data
                     break;
                 }
+                proxy_response.push(buf[0]);
+                match buf[0] {
+                    b'\n' => {
+                        if proxy_response.ends_with(b"\r\n\r\n".as_slice()) {
```
algesten, line 447: `ends_with` is a quite expensive call, however it's mitigated by that we only do it when we encounter an `\n`, so on balance probably ok.

I don't think `.as_slice()` does anything for you here.

src/stream.rs
```diff
@@ -435,13 +435,21 @@ pub(crate) fn connect_host(
...
-                proxy_response.append(&mut buf);
-                if total < 256 {
+            let mut buf = [0u8; 1];
+            while let Ok(total) = stream.read(&mut buf) {
+                if total == 0 {
+                    // no more data
                     break;
                 }
+                proxy_response.push(buf[0]);
+                match buf[0] {
```
algesten, line 445: I think this match can be replaced with `if buf[0] = b'\n' { … }`

algesten (changes requested):
Hi @puffi 

Thanks for this PR! I've made some comments.

puffi:
Thank you for the review!

Seeing the CI runs failing, I remembered that I forgot to run clippy as well.

I've force-pushed the changes.

src/stream.rs
puffi: According to the docs for Vec [0] and the source [1] this is a simple byte comparison of a subslice.
This shouldn't be that expensive.

[0] https://doc.rust-lang.org/std/vec/struct.Vec.html#method.ends_with
[1] https://doc.rust-lang.org/src/core/slice/mod.rs.html#2303-2305

src/stream.rs
```diff
@@ -435,11 +435,16 @@ pub(crate) fn connect_host(
...
-                let mut buf = vec![0; 256];
                 let total = stream.read(&mut buf)?;
-                proxy_response.append(&mut buf);
-                if total < 256 {
+                if total == 0 {
+                    // no more data
+                    break;
+                }
+
+                proxy_response.push(buf[0]);
```
algesten, line 446: While this code is a big improvement on the previous (which looks broken), there's one problem here.

Perhaps a very slim chance, but a misbehaving proxy could potentially just keep writing anything but `\r\n\r\n` and cause a denial of service attack by exhausting the available RAM by this `proxy_response` ever growing.

In the "normal" header parsing this is capped by `MAX_HEADER_COUNT` and `MAX_HEADER_SIZE`. We need to do _something_ here as well. Ideally it would use the same logic as the main header parsing (maybe we can break that out into a separate subfunction and reuse in both call sites?)

Or we need to cap the proxy_response at something like 1MB.

algesten (changes requested):
Looks good! Just thought of one more thing.

src/stream.rs
puffi: Ok, will look into that.

puffi: I've looked into it and we could actually just use the `Response` struct for parsing the response.
The easiest way to do this would be to pass a Stream to Response::do_from_stream. It requires cloning the TcpStream though.
I'm not sure if I'm missing something regarding Stream and its handling of the underlying TcpStream, especially in regards to dropping.

The response can then be used to check the status code instead of parsing the string again in verify_response.

Would that be acceptable for you or would you prefer a different solution?

algesten (approved):

algesten:
Thanks!

PR #621, opened 2023-05-09
Switch to rustls-webpki

messense:
Same as rustls: https://github.com/rustls/rustls/blob/975e1ca5949b73d43f922aea8a248b96bfcfae74/rustls/Cargo.toml#L22

Closes #609


algesten:
Thanks! Seems like the future of webpki i relation to rustls. I don't quite get why it's needed, but certainly have nothing against it.

PR #635, opened 2023-06-23
fix: pass IPv6 addresses as host name in TLS connections

haraldh:
rustls does not like the brackets `[]` in `rustls::ServerName::try_from()`.


algesten (approved):
LGTM. Thanks!

I guess there are no cases where this is the incorrect thing to do?

haraldh:
Sorry, for the missing `cargo fmt`. I blame my IDE :)

PR #638, opened 2023-07-02
Add convertion from Response to http::Response<Vec<u8>>

ToBinio:
This PR adds the possibility to convert a `Response` to a `http::Response` where the body is a `Vec<u8>`


algesten (approved):
LGTM! Thanks!

PR #642, opened 2023-07-15
fix: set up path to ureq

mcr:
We need this one line fix to be able to "cargo run" for this example.
Probably this should be turned into a proper example that can be run from the top of ureq.
I would appreciate some other people testing this.  On an ubuntu 20.04 laptop, it fails to connect, yet on an Devuan desktop, it works just fine.  Same network. Same IPv4 and IPv6 connectivity.  Still reviewing cargo tree.


algesten:
Thanks!

PR #643, opened 2023-07-27
feat: recognise system proxy env var;

LinHeLurking:
A simple commit that automatically selects proxy by environment variables.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 features = ["tls", "native-tls", "json", "charset", "cookies", "socks-proxy"]
 
 [features]
-default = ["tls", "gzip"]
+default = ["tls", "gzip", "proxy-from-env"]
 tls = ["webpki", "webpki-roots", "rustls"]
 native-certs = ["rustls-native-certs"]
 json = ["serde", "serde_json"]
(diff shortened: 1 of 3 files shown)


algesten (changes requested):
Hi! Welcome to ureq!

We don't want to do this in our standard `new()` constructor. It introduces potentially unclear behavior where proxy settings are picked up without the user's control. I.e. it should be possible to create an `AgentBuilder` that ignores the environment variables.

I think this function should be broken into two halves.

1. `Proxy::new_from_env() -> Result<Proxy>`, i.e. move the env logic to a new constructor in `Proxy`.
2. A feature flag `proxy-from-env` that is on by default, and only affects the default agent (i.e. for ureq::get).

That way someone who is using a proxy, can bypass it either by deliberately clearing the feature flag, or by using their own `AgentBuilder` without setting the `Proxy`.

LinHeLurking:
> Hi! Welcome to ureq!
> 
> We don't want to do this in our standard `new()` constructor. It introduces potentially unclear behavior where proxy settings are picked up without the user's control. I.e. it should be possible to create an `AgentBuilder` that ignores the environment variables.
> 
> I think this function should be broken into two halves.
> 
> 1. `Proxy::new_from_env() -> Result<Proxy>`, i.e. move the env logic to a new constructor in `Proxy`.
> 2. A feature flag `proxy-from-env` that is on by default, and only affects the default agent (i.e. for ureq::get).
> 
> That way someone who is using a proxy, can bypass it either by deliberately clearing the feature flag, or by using their own `AgentBuilder` without setting the `Proxy`.

Good suggestion! I've made the changes in new commit.

algesten:
Thanks!

This was sorted in a slightly different way in #649

PR #646, opened 2023-08-10
Fix mbedtls connector

mcr:
These changes update the mbedtls connector example to work with the latest mbedtls crate (0.11.0)
I'd like to change this from a self-contained example to an actual example that gets compiled regularly.


algesten:
Thanks

Issue #322, opened 2021-02-15
Automatically pick up SOCKS configuration from environment variables


Shnatsel:
As discussed in #298, HTTP and SOCKS proxy configuration is not currently detected automatically based on environment variables. I feel this goes against the crate's stated goals somewhat, since the README says

> Ureq's first priority is being easy for you to use.

and most other Rust clients that support SOCKS also automatically pick up its configuration from an environment variable.

jsha:
This makes sense to me. 👍🏻

algesten:
Sure. Let's do it.

markmmm:
Which environment variables should be supported `http_proxy`, `https_proxy`? More: https://gist.github.com/yougg/5d2b3353fc5e197a0917aae0b3287d64 ?  (In case I find time to look into it, knowing which environment variables will help :) )

njsmith:
It looks like https://crates.io/crates/env_proxy has already done research on how folks expect these env vars to work, so that might be a good place to start.

Dolpheyn:
> It looks like https://crates.io/crates/env_proxy has already done research on how folks expect these env vars to work, so that might be a good place to start.

If we were to use this crate, we have to have a URL in order to get the correct proxy from env vars(using `env_proxy::for_url_str`), which in our case is given in `Request::new`. Can(or should) there be a way to change the Agent's config when instantiating a new `Request`?

One other way I can think of is we store all the possible proxies in `Agent` and choose to use the correct one in `Unit::connect_socket`.

If this issue is still free, I want to try and take a stab at it. May need some guidance :)

bemug:
I'm interested in this, @Dolpheyn did you give it a go?

rruckley:
Anyone looking at this? Can't be that hard to implement.

algesten:
Yeah, I'm lucky I never had to work behind proxies, so from what I know no one picked it up.

algesten:
This was fixed in 2.8.0 (and I just released 2.9.0). 

I believe this is done.

PR #649, opened 2023-08-11
Auto detect proxy from environments

dojiong:
solve #322 
----
Just like reqwest, auto detect proxy from `ALL_PROXY`/`HTTPS_PROXY`/`HTTP_PROXY` [reqwest::proxy.rs](https://github.com/seanmonstar/reqwest/blob/master/src/proxy.rs#L834)

Add `AgentBuilder::no_proxy` if someone don't want to use system's proxy

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -37,7 +37,7 @@ pub enum RedirectAuthHeaders {
 /// Accumulates options towards building an [Agent].
 pub struct AgentBuilder {
     config: AgentConfig,
-    no_proxy: bool,
+    try_proxy_from_env: bool,
     max_idle_connections: usize,
     max_idle_connections_per_host: usize,
     /// Cookies saved between requests.
@@ -264,7 +264,7 @@ impl AgentBuilder {
                 user_agent: format!("ureq/{}", env!("CARGO_PKG_VERSION")),
                 tls_config: TlsConfig(crate::default_tls_config()),
             },
-            no_proxy: false,
+            try_proxy_from_env: false,
             max_idle_connections: DEFAULT_MAX_IDLE_CONNECTIONS,
             max_idle_connections_per_host: DEFAULT_MAX_IDLE_CONNECTIONS_PER_HOST,
             resolver: StdResolver.into(),
@@ -280,7 +280,7 @@ impl AgentBuilder {
     // not implement clone, so we have to give ownership to the newly
     // built Agent.
     pub fn build(mut self) -> Agent {
-        if self.config.proxy.is_none() && !self.no_proxy {
+        if self.config.proxy.is_none() && self.try_proxy_from_env {
             if let Some(proxy) = Proxy::try_from_system() {
                 self.config.proxy = Some(proxy);
             }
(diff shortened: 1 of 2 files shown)


src/proxy.rs
```diff
@@ -67,6 +67,30 @@ impl Proxy {
...
+            "ALL_PROXY",
+            "all_proxy",
+            "HTTPS_PROXY",
+            "https_proxy",
+            "HTTP_PROXY",
+            "http_proxy"
+        );
+        None
+    }
+
```
algesten, line 93: I like this. Nice!

src/agent.rs
```diff
@@ -277,7 +279,12 @@ impl AgentBuilder {
     // AgentBuilder to be used multiple times, except CookieStore does
     // not implement clone, so we have to give ownership to the newly
     // built Agent.
-    pub fn build(self) -> Agent {
+    pub fn build(mut self) -> Agent {
+        if self.config.proxy.is_none() && !self.no_proxy {
+            if let Some(proxy) = Proxy::try_from_system() {
+                self.config.proxy = Some(proxy);
+            }
+        }
```
algesten, line 287: I don't think this is the place to do this. It changes the default behavior (which is to not pick up any proxy settings), and that might be a breaking change for some users (thus requiring a major version bump).

I think we need to feature-flag this to be opt-in behavior. Or am I being overly cautious?

algesten (changes requested):

src/agent.rs
dojiong: I think that's fine to change this default behavior, the user who use these envs is known the traffic will in proxy, it's the default behavior in many libraries or tools.

Narsil: I think you're correct. This is enabling very bad behavior. 
`curl` and `python.requests` accept this.. but this is VERY bad in my opinion, just made a small proof of concept, I can intercept all traffic...

Narsil: Just for fun: https://github.com/Narsil/test_proxy.

Seems crazy now that I'm aware of it...

Narsil: Also `requests` had until very recently a very nice flag to disable certs... https://github.com/psf/requests/issues/6071 good combo.

dojiong: > Just for fun: https://github.com/Narsil/test_proxy.

that's what a proxy do, sniffe or cache or vpn.

> Also requests had until very recently a very nice flag to disable certs... https://github.com/psf/requests/issues/6071 good combo.

this issuse is a empty-env bug(should not disable verify on empty env, because empty-config is easy to misuse), it does not indicates CURL_CA_BUNDLE is insecure, it's really useful in some cases.

dojiong: proxy env is a simple way to redirect traffic when it's hard to setup a global vpn

Narsil: https://github.com/seanmonstar/reqwest/issues/1937

algesten (approved):

src/agent.rs
algesten: On balance I'm not swayed by having this turned on by default. My main problem is that it breaks backwards compatibility of ureq. For now, this defaults to off.

algesten:
Thanks for the work in this PR!

PR #652, opened 2023-08-21
add into_vec

dev-ardi:
There's still some work to do but this would be the general idea

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -481,6 +481,17 @@ impl Response {
         }
     }
 
+    pub fn into_vec(self) -> Result<Vec<u8>, Error> {
+        let size = self
+            .header("Content-Type")
+            .unwrap_or_default()
+            .parse::<usize>()
+            .unwrap_or_default();
+        let mut buf = Vec::with_capacity(size);
+        self.into_reader().read_to_end(&mut buf)?;
+        Ok(buf)
+    }
+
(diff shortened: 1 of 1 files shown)


algesten:
The idea is to use `into_reader()` for this.

```rs
    let mut bytes: Vec<u8> = Vec::with_capacity(len);
    resp.into_reader()
        .read_to_end(&mut bytes)?;
```

If that's unclear, we could look into making the doc better.

dev-ardi:
Isn't this the same code that I sent?

src/response.rs
```diff
@@ -481,6 +481,17 @@ impl Response {
         }
     }
 
+    pub fn into_vec(self) -> Result<Vec<u8>, Error> {
+        let size = self
+            .header("Content-Type")
```
johan-bjareholt, line 486: Is this supposed to be "Content-Length"?
Also, what happens if e.g. "Transfer-Encoding: chunked" is used?
Using into_reader seems more convenient.

dev-ardi: > Is this supposed to be "Content-Length"?

Oops it is :)

> Also, what happens if e.g. "Transfer-Encoding: chunked" is used?

Well in that case more allocations will be made, this is just a small optimization

algesten:

Issue #653, opened 2023-08-22
Use `cargo-deny` to prevent duplicate dependencies


emilk:
`ureq v2.7.1` imports two different versions of `rustls-webpki`

```
    = rustls-webpki v0.100.2
      ├── ureq v2.7.1
      └── webpki-roots v0.23.1
          └── ureq v2.7.1
          
    = rustls-webpki v0.101.4
      └── rustls v0.21.6
          └── ureq v2.7.1
```

This leads to extra compile time and code bloat.

We can prevent this by running `cargo-deny` on CI https://github.com/EmbarkStudios/cargo-deny

algesten:
Yeah. This would be really nice. A PR changing the github CI would be most welcome!

fredizzimo:
There's also a security issue in the 0.100.2, version, so it would be nice to get this updated. Dependabot reports it here for us https://github.com/neovide/neovide/security/dependabot/15

PR #661, opened 2023-09-30
Add cargo-deny check to CI

emilk:
[cargo-deny](https://github.com/EmbarkStudios/cargo-deny) is an amazing tool that protects from:

* duplicated crates (code bloat)
* copy-left licenses in the dependency tree
* RUSTSEC advisories

Try it:
```sh
cargo install cargo-deny
cargo deny --all-features --log-level error check
```

* Closes https://github.com/algesten/ureq/issues/653

Note that this requires the `Cargo.lock` file to be checked in.


algesten:
Oh wow. This is very nice!

I guess I should hold off on that 2.8.0 release and see if we can get this in.

algesten:
Great! Thanks!

algesten:
Hej @emilk

This is new to me. So when we now have a problem with duplicates: https://github.com/algesten/ureq/actions/runs/6361674777/job/17276257519 

What do we do?

* Wait for it to sort itself out (when upstreams dep fix it)?
* Lock down our deps?

emilk:
`cargo tree -i` is an invaluable tool to find the causes of duplicate dependencies.

Often you can solve it by strategically downgrading or upgrading some crates using `cargo update -p $crate --precise $version`.

Sometimes you need to make a PR to an upstream package, or poke their maintainer to release a new version.

For small duplicate dependencies you can add an exception in `deny.toml`

---

In this case I believe you can solve it with `cargo update -p url --precise 2.3.1`

This will update the versions in `Cargo.lock`, proving to cargo-deny that it is possible to build `ureq` without duplicate dependencies. Note that this does NOT prevent users from upgrading `url` to `2.4.1` again, and reintroducing that duplicate dependency (that is out of your hands, as it should be)

algesten:
Perfect! Thanks for the explanation. I didn't know the `cargo update` commands for this, but I believe I did the same, manually in `Cargo.toml`.

Thanks!

emilk:
You did this::

```toml
url = "=2.3.1"
```

which is NOT the same. This forces url 2.3.1 on all downstream users of `ureq`, meaning they cannot update `url`, even if 2.3.1 is shown to have some vulnerability fixed by e.g. 2.4

It is almost never the right choice to use `=` in a `Cargo.toml` version string - better to update just the `Cargo.lock` using `cargo update -p url --precise 2.3.1`

algesten:
Right. So what you're saying is that we are only going to make the lock file have 2.3.1, which satisfies `url = "2"`, but we're not going to help any consumer of ureq to avoid the double dep by locking down the version in toml?

emilk:
Exactly - consumers will have to do a similar `cargo update -p url --precise 2.3.1` if they also wanna avoid duplicate dependencies

PR #670, opened 2023-10-10
Enable all features in documentation and utilize `doc_auto_cfg`

MarijnS95:
Testing all features instead of just a specific set makes it possible for the CI to spot various mistakes in currently-untested doc comments, such as those behind the `http-interop` feature.  More importantly, enabling _all features_ on docs.rs as well massively helps visibility in available features and functionality to the end user.  And `doc_cfg`/ `doc_auto_cfg` automatically provide them the right information block with a (list of) feature(s) they need to enable to get access to a certain type or function.

Note that this requires a `nightly` feature and is hence behind a `--cfg` flag that is only enabled when building for docs.rs (using `nightly`). It can be reproduced locally via `RUSTDOCFLAGS=--cfg=docsrs cargo +nightly doc`.

Additionally the CI script is updated to pass `-Dwarnings` to `rustdoc` so that it can adequately fail the CI when incorrect documentation is encountered.


MarijnS95:
image

Note that I'm thinking of yoinking these `Requires feature` doc-comments now: they're ambiguous/duplicate and will in this case only enable the `http` crate but not the `http-interop` module (maybe this should have been a `http = ["dep:http"]` wrapper feature)?

algesten:
Thanks!

MarijnS95:
Oh. I did not expect the CI to succeed just yet because I have a bunch of warnings locally on Rust 1.73 :frowning:

MarijnS95:
Trying to sort that out in //github.com/algesten/ureq/pull/673

PR #672, opened 2023-10-10
http_interop: Remove requirement for header values to be UTF-8

MarijnS95:
CC @kade-robertson

The current interop implementation forces fallibility around header values when they are not UTF-8, both in conversions from `http` to `ureq` as well as the other way around.

This should not be necessary as both `http` and `ureq` treat these as opaque byte arrays internally, but unfortunately the current open API for `ureq` and `http` make it a bit nasty to deal with.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -98,7 +98,7 @@ impl Header {
 
     /// The header value.
     ///
-    /// For non-utf8 headers this returns None (use [`Header::value_raw()`]).
+    /// For non-utf8 headers this returns [`None`] (use [`Header::value_raw()`]).
     pub fn value(&self) -> Option<&str> {
         let bytes = &self.line.as_bytes()[self.index + 1..];
         from_utf8(bytes)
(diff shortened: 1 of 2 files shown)


algesten (approved):
LGTM

algesten:
Can we construct a test for this?

MarijnS95:
Sure! I added them for requests but probably also have to add them for responses... And found that the current `str`-based convenience API makes it a bit tedious to test these :)

algesten:
Looks great! Thanks!

Maybe that test helper function need to be allowed unused for some test feature flag combo?

MarijnS95:
Yeah. It's a bit nasty though as it's a useful function, just not something to expose publicly and the conditional is quite large now...

algesten (approved):

algesten:
Thank you! This looks great!

MarijnS95:
@algesten thanks! This should be the last of my PRs to `http_interop`.

PR #673, opened 2023-10-11
CI: Drop redundant `no-default-features` from docs and test private items

MarijnS95:
I added `--all-features` in #670 without removing `--no-default-features` which now has no effect anymore.

In addition it is useful to make sure that documentation on private items also adheres to valid intradoc, and somewhat strangely enabling this also causes additional lint warnings on public documentation...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
           RUSTDOCFLAGS: -Dwarnings
         with:
           command: doc
-          args: --no-default-features --no-deps --all-features
+          args: --no-deps --all-features --document-private-items
   build_and_test:
     name: Test
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


MarijnS95:
As expected: https://github.com/MarijnS95/ureq/actions/runs/6480377560/job/17595786779

algesten:
How would private items interact with `#![deny(missing_docs)]`?

I would like us to have `#![deny(missing_docs)]`, but I don't want to force us to document private items.

MarijnS95:
@algesten this lint is enforced by `rustdoc` (as well as `rustc`: https://doc.rust-lang.org/rustdoc/lints.html) but only for public items even when passing `--no-default-features`. It is only complaining about 2 public type aliases, 7 public functions and 1 public trait that is undocumented :tada: 

Note that if you're doing this, the crate might as well have pass for missing backticks and intradoc links, as I did for one comment in this PR :)

algesten:
Alright. Let's do it!

MarijnS95:
Can always revert if it turns out to cause future issues :)

PR #674, opened 2023-10-11
Drop redundant or wrong `Requires feature` documentation

MarijnS95:
As mentioned in https://github.com/algesten/ureq/pull/670#issuecomment-1756381424 this sort of documentation is superfluous now that `doc_cfg`/ `doc_auto_cfg` **automatically** generates colored blocks in the documentation denoting what `cfg`s (including features) need to be met in order to enable a symbol.

In addition the `http` feature only exists to turn on the `http` crate, while the type conversions in the `http_interop.rs` module are only provided behind the `http-interop` feature.

The `charset` feature documentation is left in place, because that affects additional functionality deeper down inside the `send_string()` function.


algesten:
Thanks!

Issue #676, opened 2023-10-27
Sometimes fails to retrieve a response when the server doesn't read the body


johan-bjareholt:
I have the use-case where we have a client that sends a multipart to a server and if there is something wrong in the first part it will not read the following parts.
The issue we have is that if there is something wrong in the first part and the server responds without reading the whole request, we will get a "Connection reset by peer", while what we want is to get the "400 Bad Request" which is what the server responds with.

It seems like the bigger the request body is, the more likely it is that it will get "Connection reset by peer". If the message is small, e.g. if you set the "huge_string" variable in the code to something smaller, it works does respond with 200 OK.

Another use-case that might be more common where this issue might also occur, would be if a server responds early with a "413 Request Entity Too Large".

I have written a basic test client and server to demonstrate the issue

#### Test client
```rust
use ureq;

fn main() -> Result<(), ureq::Error> {
    let huge_string = "abcde ".repeat(50000);
    let body: String = ureq::post("http://127.0.0.1:7878")
        .send_string(&huge_string)?
        .into_string()?;
    println!("body: {}", body);
    Ok(())
}
```

#### Test server
```rust
use std::{
    io::{prelude::*, BufReader},
    net::{TcpListener, TcpStream},
};

fn main() {
    let listener = TcpListener::bind("127.0.0.1:7878").unwrap();

    for stream in listener.incoming() {
        let stream = stream.unwrap();

        handle_connection(stream);
    }
}

fn handle_connection(mut stream: TcpStream) {
    let read = false;
    let respond = true;

    if read {
        let mut buf_reader = BufReader::new(&mut stream);

        let mut http_request = vec![0u8; 1024*1024];
        let bytes_read = buf_reader.read(&mut http_request).unwrap();
        let request = std::str::from_utf8(&http_request[0..bytes_read]).unwrap();
        println!("Request: {:#?}", request);
    }
    //stream.shutdown(std::net::Shutdown::Read).unwrap();

    if respond {
        let status_line = "HTTP/1.1 200 OK";
        let contents = "Hello, I am a responding server!";
        let length = contents.len();

        let response =
            format!("{status_line}\r\nContent-Length: {length}\r\n\r\n{contents}");

        println!("Sending response");
        stream.write_all(response.as_bytes()).unwrap();
    }
    println!("Done");
}

```

algesten:
[...]

PR #679, opened 2023-11-07
Add support for "Expect: 100-continue" header

johan-bjareholt:
Works well, very easy to use by just setting the expect header like this

```rust
let huge_string = "abcde ".repeat(500);
let req = ureq::post("http://127.0.0.1:5000")
    .set("Expect", "100-continue");
 let res = req.send_string(&huge_string).unwrap();
println!("res: {:?}", res);
let body = res.into_string().unwrap();
println!("body: {}", body);
```

## To-do

- [x] A timeout to send body even if server does not understand "Expect: 100-continue"
    - [x] If the client timed out waiting for the HTTP/1.1 100 Continue, then proceed with sending the body without waiting for an "100 Continue" response.
    - [ ] (optional) Make the timeout configurable (1000ms default)
- [x] Follow the spec and on 417 resend the whole request but without the expect-100 header.
- [ ] (optional) Set "Expect: 100-continue" by default
    - Only set 100-continue if there is a body, if the body has a known size over X MB or always if the size is not known

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -561,7 +561,7 @@ impl Response {
     /// let resp = ureq::Response::do_from_read(read);
     ///
     /// assert_eq!(resp.status(), 401);
-    pub(crate) fn do_from_stream(stream: Stream, unit: Unit) -> Result<Response, Error> {
+    pub(crate) fn do_from_stream(stream: Stream, unit: &Unit) -> Result<Response, Error> {
         let remote_addr = stream.remote_addr;
 
         let local_addr = match stream.socket() {
@@ -609,7 +609,7 @@ impl Response {
         }
 
         let reader =
-            Self::stream_to_reader(stream, &unit, body_type, compression, connection_option);
+            Self::stream_to_reader(stream, unit, body_type, compression, connection_option);
 
         let url = unit.url.clone();
 
@@ -766,7 +766,7 @@ impl FromStr for Response {
             &request_reader,
             None,
         );
-        Self::do_from_stream(stream, unit)
+        Self::do_from_stream(stream, &unit)
     }
 }
 
@@ -1150,7 +1150,7 @@ mod tests {
             &request_reader,
             None,
         );
-        let resp = Response::do_from_stream(s.into(), unit).unwrap();
+        let resp = Response::do_from_stream(s.into(), &unit).unwrap();
         assert_eq!(resp.status(), 200);
         assert_eq!(resp.header("x-geo-header"), None);
     }
@@ -1206,7 +1206,7 @@ mod tests {
         );
         Response::do_from_stream(
             stream,
-            Unit::new(
+            &Unit::new(
                 &agent,
                 "GET",
                 &"https://example.com/".parse().unwrap(),
@@ -1238,7 +1238,7 @@ mod tests {
         );
         let resp = Response::do_from_stream(
             stream,
-            Unit::new(
+            &Unit::new(
                 &agent,
                 "GET",
                 &"https://example.com/".parse().unwrap(),
--- a/src/response.rs
+++ b/src/response.rs
@@ -49,6 +49,11 @@ enum BodyType {
     CloseDelimited,
 }
 
+pub(crate) enum PendingReader {
+    BeforeBodyStart,
+    Reader(Box<dyn Read + Send + Sync + 'static>),
+}
+
 /// Response instances are created as results of firing off requests.
 ///
 /// The `Response` is used to read response headers and decide what to do with the body.
@@ -81,7 +86,7 @@ pub struct Response {
     pub(crate) index: ResponseStatusIndex,
     pub(crate) status: u16,
     pub(crate) headers: Vec<Header>,
-    pub(crate) reader: Box<dyn Read + Send + Sync + 'static>,
+    pub(crate) reader: PendingReader,
     /// The socket address of the server that sent the response.
     pub(crate) remote_addr: SocketAddr,
     /// The socket address of the client that sent the request.
@@ -282,7 +287,12 @@ impl Response {
     /// # }
     /// ```
     pub fn into_reader(self) -> Box<dyn Read + Send + Sync + 'static> {
-        self.reader
+        match self.reader {
+            PendingReader::Reader(reader) => reader,
+            PendingReader::BeforeBodyStart => panic!(
+                "It is not valid to call into_reader before Request::stream_to_reader is called"
+            ),
+        }
     }
 
     // Determine what to do with the connection after we've read the body.
@@ -548,39 +558,33 @@ impl Response {
         })
     }
 
-    /// Create a response from a Read trait impl.
+    /// Create a response from a DeadlineStream, reading and parsing only the status line, headers
+    /// and its following CRLF.
     ///
-    /// This is hopefully useful for unit tests.
-    ///
-    /// Example:
-    ///
-    /// use std::io::Cursor;
-    ///
-    /// let text = "HTTP/1.1 401 Authorization Required\r\n\r\nPlease log in\n";
-    /// let read = Cursor::new(text.to_string().into_bytes());
-    /// let resp = ureq::Response::do_from_read(read);
-    ///
-    /// assert_eq!(resp.status(), 401);
-    pub(crate) fn do_from_stream(stream: Stream, unit: &Unit) -> Result<Response, Error> {
-        let remote_addr = stream.remote_addr;
+    /// Since this function only reads the status line, header and the following CRLF, the returned
+    /// Response will have an empty reader and does not take ownership of DeadlineStream.
+    /// To read the following data, the DeadlineStream can be read again after the call to this
+    /// function.
+    pub(crate) fn read_response_head(
+        stream: &mut DeadlineStream,
+        unit: &Unit,
+    ) -> Result<Response, Error> {
+        let mut bytes_read = 0;
+        let remote_addr = stream.inner_ref().remote_addr;
 
-        let local_addr = match stream.socket() {
+        let local_addr = match stream.inner_ref().socket() {
             Some(sock) => sock.local_addr().map_err(Error::from)?,
             None => std::net::SocketAddrV4::new(std::net::Ipv4Addr::new(127, 0, 0, 1), 0).into(),
         };
 
-        //
-        // HTTP/1.1 200 OK\r\n
-        let mut stream = stream::DeadlineStream::new(stream, unit.deadline);
-
         // The status line we can ignore non-utf8 chars and parse as_str_lossy().
-        let status_line = read_next_line(&mut stream, "the status line")?.into_string_lossy();
+        let status_line =
+            read_next_line(stream, "the status line", &mut bytes_read)?.into_string_lossy();
         let (index, status) = parse_status_line(status_line.as_str())?;
-        let http_version = &status_line.as_str()[0..index.http_version];
 
         let mut headers: Vec<Header> = Vec::new();
         while headers.len() <= MAX_HEADER_COUNT {
-            let line = read_next_line(&mut stream, "a header")?;
+            let line = read_next_line(stream, "a header", &mut bytes_read)?;
             if line.is_empty() {
                 break;
             }
@@ -595,23 +599,8 @@ impl Response {
             ));
         }
 
-        let compression =
-            get_header(&headers, "content-encoding").and_then(Compression::from_header_value);
-
-        let connection_option =
-            Self::connection_option(http_version, get_header(&headers, "connection"));
-
-        let body_type = Self::body_type(&unit.method, status, http_version, &headers);
-
-        // remove Content-Encoding and length due to automatic decompression
-        if compression.is_some() {
-            headers.retain(|h| !h.is_name("content-encoding") && !h.is_name("content-length"));
-        }
-
-        let reader =
-            Self::stream_to_reader(stream, unit, body_type, compression, connection_option);
-
         let url = unit.url.clone();
+        let reader = PendingReader::BeforeBodyStart;
 
         let response = Response {
             url,
@@ -624,6 +613,54 @@ impl Response {
             local_addr,
             history: vec![],
         };
+        Ok(response)
+    }
+
+    /// Attach a stream to Response, for reading the body.
+    ///
+    /// The response reader also uncompresses the body if it is compressed.
+    pub(crate) fn take_body(&mut self, stream: DeadlineStream, unit: &Unit) -> Result<(), Error> {
+        let compression =
+            get_header(&self.headers, "content-encoding").and_then(Compression::from_header_value);
+
+        let connection_option =
+            Self::connection_option(self.http_version(), get_header(&self.headers, "connection"));
+
+        let body_type = Self::body_type(
+            &unit.method,
+            self.status(),
+            self.http_version(),
+            &self.headers,
+        );
+
+        // remove Content-Encoding and length due to automatic decompression
+        if compression.is_some() {
+            self.headers
+                .retain(|h| !h.is_name("content-encoding") && !h.is_name("content-length"));
+        }
+
+        self.reader = PendingReader::Reader(Self::stream_to_reader(
+            stream,
+            unit,
+            body_type,
+            compression,
+            connection_option,
+        ));
+
+        Ok(())
+    }
+
+    /// Create a Response from a DeadlineStream
+    ///
+    /// Parses and comsumes the header from the stream and creates a Response with
+    /// the stream as the reader. The response reader also uncompresses the body
+    /// if it is compressed.
+    pub(crate) fn do_from_stream(
+        mut stream: DeadlineStream,
+        unit: &Unit,
+    ) -> Result<Response, Error> {
+        let mut response = Self::read_response_head(&mut stream, unit)?;
+        response.take_body(stream, unit)?;
         Ok(response)
     }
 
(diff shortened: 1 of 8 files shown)


johan-bjareholt:
This is a follow up to issue #676

src/response.rs
```diff
@@ -595,24 +590,15 @@ impl Response {
...
-        let connection_option =
-            Self::connection_option(http_version, get_header(&headers, "connection"));
-
-        let body_type = Self::body_type(&unit.method, status, http_version, &headers);
-
-        // remove Content-Encoding and length due to automatic decompression
-        if compression.is_some() {
-            headers.retain(|h| !h.is_name("content-encoding") && !h.is_name("content-length"));
+        if consume {
+            stream.fill_buf().unwrap();
```
johan-bjareholt, line 594: Don't unwrap here

src/unit.rs
```diff
@@ -261,6 +261,13 @@ fn connect_inner(
         debug!("sending request {} {}", method, url);
     }
 
+    let mut expect_100_continue = false;
+    for header in &unit.headers {
+        if header.name() == "Expect" && header.value().unwrap() == "100-continue" {
```
johan-bjareholt, line 266: Don't unwrap here

algesten:
Thanks for looking into this!

Overall looks like good changes. You can use the `test.sh` script in the root to roughly run the same tests the CI will do.

src/response.rs
```diff
@@ -548,39 +548,34 @@ impl Response {
...
-        let remote_addr = stream.remote_addr;
-
-        let local_addr = match stream.socket() {
+    /// Since this function only reads the status line, header and the following CRLF, the returned
+    /// Response will have an empty reader and does not take ownership of DeadlineStream.
+    /// To read the following data, the DeadlineStream can be read again after the call to this
+    /// function.
+    /// If consume is set to true, the status line and headers will be consumed in the
+    /// DeadlineStream, so next call to read will start after the headers and their CRLF.
+    pub(crate) fn do_head_from_stream(stream: &mut DeadlineStream, unit: &Unit, consume: bool) -> Result<Response, Error> {
```
algesten, line 560: Let's rename this something like `read_response_head`

src/response.rs
```diff
@@ -548,39 +548,34 @@ impl Response {
...
             None => std::net::SocketAddrV4::new(std::net::Ipv4Addr::new(127, 0, 0, 1), 0).into(),
         };
 
-        //
-        // HTTP/1.1 200 OK\r\n
-        let mut stream = stream::DeadlineStream::new(stream, unit.deadline);
-
         // The status line we can ignore non-utf8 chars and parse as_str_lossy().
-        let status_line = read_next_line(&mut stream, "the status line")?.into_string_lossy();
+        let (bytes_read, status_line) = read_next_line(stream, "the status line")?;
```
algesten, line 570: Would it maybe be better to have `read_next_line` take a third argument `running_total: &mut usize`, so we don't need to return and remember to add to the total?

src/response.rs
```diff
@@ -595,24 +590,15 @@ impl Response {
...
+            stream.fill_buf()?;
+            stream.consume(tot_bytes_read);
         }
 
-        let reader =
-            Self::stream_to_reader(stream, &unit, body_type, compression, connection_option);
-
         let url = unit.url.clone();
 
+        let reader = Box::new(Cursor::new(vec![]));
```
algesten, line 600: This extra allocation on every (regular, also not expect-100) call irks me a bit.

What if we replace `reader: Box<dyn Read + Send + Sync + 'static>` in `Response` with an enum `PendingReader` that has two values which explains the state.

`PendingReader::BeforeBodyStart` and `PendingReader::Reader(Box<dyn …>)`

Could be modeled as an `Option` too, but maybe a bit clearer with an enum. What do you think?

src/response.rs
```diff
@@ -627,6 +613,41 @@ impl Response {
...
+    ///
+    /// Example:
+    ///
+    /// use std::io::Cursor;
+    ///
+    /// let text = "HTTP/1.1 401 Authorization Required\r\n\r\nPlease log in\n";
+    /// let read = Cursor::new(text.to_string().into_bytes());
+    /// let resp = ureq::Response::do_from_read(read);
+    ///
+    /// assert_eq!(resp.status(), 401);
```
algesten, line 628: This doc is just "lol" at this point (it used to make sense). Not your fault, but moving it around seems a bit much :D

algesten (changes requested):
Some comments.

Was thinking of a bigger refactor might be clearer.

What if the part that just reads the header returns a new type `ResponseHead` (not public). Then there are two methods to "complete" that type into the public `Response`, either `without_body` (for expect-100) and `with_body` for the normal case?

It would potentially be cleaner than trying to make `Response` be slightly different depending...

jsha:
Thanks so much for the PR! I'll review it soon.

jsha:
Oops, I posted that before I saw @algesten had already provided a review! But I will still plan to find some time and take a look soon.

src/response.rs
johan-bjareholt: I thought about Option too, but yeah it's more clear with enum so I went with that.

johan-bjareholt:
I removed the "consume" argument from "read_response_head", the reason why it existed was because I was testing against a broken http server that for some strange reason responded with "100 Continue" twice... It took me embarassly long until I realised this and found that out in wireshark.

I have now tested this code against a libsoup2 server and a python flask http server, works fine there.

algesten:
I think this looks pretty good. Wonder if @jsha agrees?

jsha (approved):
This also looks good to me! Thanks so much for writing it.

jsha:
Looks like you've got a clippy error. Mind fixing that and I'll merge?

johan-bjareholt:
> Looks like you've got a clippy error. Mind fixing that and I'll merge?

Done!

johan-bjareholt:
Also, regarding the broken client. I'm not sure if this is a flask bug or something. To be more exact, when using "default" flask it's broken, but when using flask with werkzeug "serving.make_server" it works fine. Might be a bug in flask?


testclient
```rust
fn request() {
    let string = "abcde ".repeat(5);

    let req = ureq::post("http://127.0.0.1:5000/")
        .set("Expect", "100-continue");
    let res = req.send(string.as_bytes()).unwrap();
    println!("res2: {:?}", res);
    assert_eq!(res.status(), 200);
    let body = res.into_string().unwrap();
    assert_eq!("Hello, world!", body);
}

fn main() {
    request();
}

#[test]
fn test_a() {
    for i in 1..1000 {
        println!("\nRun {i}\n");
        request();
    }
}
```

Flask with werkzeug serving.make_server (works)
```python
#!/usr/bin/env python3

from flask import Flask, request
from werkzeug import serving

app = Flask(__name__)

@app.route('/', methods = ['POST'])
def a():
    return "Hello, world!"


if __name__=='__main__':
    app.use_reloader = False
    server = serving.make_server(
        host="0.0.0.0",
        port=5000,
        app=app,
        ssl_context=None)
    server.serve_forever()
```

Flask with app.run (broken)
```python
#!/usr/bin/env python3

from flask import Flask, request

app = Flask(__name__)

@app.route('/', methods = ['POST'])
def a():
    return "Hello, world!"


if __name__=='__main__':
    app.run()
```

Wireshark of broken server
```
POST / HTTP/1.1
Host: 127.0.0.1:5000
User-Agent: ureq/2.8.0
Accept: */*
Expect: 100-continue
accept-encoding: gzip
Transfer-Encoding: chunked

HTTP/1.1 100 Continue

HTTP/1.1 100 Continue

1e
abcde abcde abcde abcde abcde 
0

HTTP/1.1 200 OK
Server: Werkzeug/2.2.2 Python/3.9.2
Date: Fri, 10 Nov 2023 07:39:39 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 13
Connection: close
```

What happens is that because we get two 100-continue, our ureq client consumes the first 100-continue as it should and then puts the second 100-continue as its body, which the test client then asserts on as it expects a 200 OK.

algesten:
I might have dreamt this. But I seem to recall curl waits _for a while_ before sending the body regardless of a 100 response. This is to interoperate with servers that doesn't understand expect-100

https://www.rfc-editor.org/rfc/rfc7231.html#section-5.1.1

algesten:
Ah yes `CURLOPT_EXPECT_100_TIMEOUT_MS` defaults to 1000ms

johan-bjareholt:
Sounds reasonable, the only risk I see with this is if the server is just slow with responding with the 100-continue. So it might be a bit racy, seems like it would be important in that case to be able to configure the timeout?

algesten:
I think there are two things we want to consider. 

1. A timeout to send body anyway. 
2. Follow the spec and on 417 resend the request without the expect-100 header.

jsha:
I think the racy-ness can be resolved like so:

If the client timed out waiting for the `HTTP/1.1 100 Continue`, then when it tries to read headers again after sending the body, it's possible it may receive an `HTTP/1.1 100 Continue` at that time, followed by the final headers. If the "reading headers after body" step gets a `100` (vs any other code), it should respond to that by reading headers a second time, and returning those.

Here's the most up-to-date RFC (substantively the same as RFC 7231 on this issue): https://www.rfc-editor.org/rfc/rfc9110#field.expect

> A client MUST NOT generate a 100-continue expectation in a request that does not include content.
> A client that will wait for a [100 (Continue)](https://www.rfc-editor.org/rfc/rfc9110#status.100) response before sending the request content MUST send an [Expect](https://www.rfc-editor.org/rfc/rfc9110#field.expect) header field containing a 100-continue expectation.
> A client that sends a 100-continue expectation is not required to wait for any specific length of time; such a client MAY proceed to send the content even if it has not yet received a response. Furthermore, since [100 (Continue)](https://www.rfc-editor.org/rfc/rfc9110#status.100) responses cannot be sent through an HTTP/1.0 intermediary, such a client SHOULD NOT wait for an indefinite period before sending the content.
> A client that receives a [417 (Expectation Failed)](https://www.rfc-editor.org/rfc/rfc9110#status.417) status code in response to a request containing a 100-continue expectation SHOULD repeat that request without a 100-continue expectation, since the 417 response merely indicates that the response chain does not support expectations (e.g., it passes through an HTTP/1.0 server).

I think we're seeing that the "time out if no early response" SHOULD is actually pretty necessary for decent operation, so let's add it. @algesten any thoughts on wanting to include it as part of this PR or as a follow-on?

I think the "retry on 417" SHOULD is good but I think it's not as urgent.

algesten:
Splitting those apart.

> A client MUST NOT generate a 100-continue expectation in a request that does not include content.

We should definitely follow this. No expect unless there's a body.

> A client that will wait for a 100 (Continue) response before sending the request content MUST send an header field containing a 100-continue expectation.

✅  Doing that already.

> A client that sends a 100-continue expectation is not required to wait for any specific length of time; such a client MAY proceed to send the content even if it has not yet received a response.

I think we should follow curl behavior here. Have a 1000ms default timeout and make it configurable on `Agent`.

> Furthermore, since 100 (Continue) responses cannot be sent through an HTTP/1.0 intermediary, such a client SHOULD NOT wait for an indefinite period before sending the content.

Same as previous. But yeah. We shouldn't wait forever, even if it's not a MUST.

> A client that receives a 417 status code in response to a request containing a 100-continue expectation SHOULD repeat that request without a 100-continue expectation, since the 417 response merely indicates that the response chain does not support expectations (e.g., it passes through an HTTP/1.0 server).

I think we should be good citizens and do this too, even if it's not a MUST. However on this point I can be convinced otherwise. Doesn't seem like a big deal though.

---------------

Whether the work is added on to this PR or a new PR, I don't think matters much. Either way I think we need to do this before shipping a new version with this functionality.

Just throwing out there: Should we do like curl and _always_ send expect-100 headers and wait 1000ms when we have content?

@johan-bjareholt You already done a lot here. Thanks! It's up to you whether you want to take on these further requirements. Let us know if you intend to continue, or we will work out a plan for the rest of the work. No pressure!

johan-bjareholt:
> Whether the work is added on to this PR or a new PR, I don't think matters much. Either way I think we need to do this before shipping a new version with this functionality.

The pro of not merging it would be that then there wouldn't be a hurry to fix it before the next release, as it doesn't become a blocker. If we don't find it to be complete enough for a release, maybe we shouldn't merge it?

> Just throwing out there: Should we do like curl and always send expect-100 headers and wait 1000ms when we have content?

To do it unconditionally seems like it would be against the point of what "Expect: 100-continue" is supposed to solve, possibility to deny requests early that are large to not waste network bandwidth and processing? To do it if the request payload is bigger than X megabytes would seem reasonable to me however.

> @johan-bjareholt You already done a lot here. Thanks! It's up to you whether you want to take on these further requirements. Let us know if you intend to continue, or we will work out a plan for the rest of the work. No pressure!

Thanks, I highly appreciate the fast and thorough support! 

I would love to help out some more,  but the coming week will be a bit busy for me. So if you have some patience, I could continue working on this. For my use-case, the current solution works but the suggested improvements would also help.

jsha:
> The pro of not merging it would be that then there wouldn't be a hurry to fix it before the next release, as it doesn't become a blocker. If we don't find it to be complete enough for a release, maybe we shouldn't merge it?

Good point. I agree.

> > Just throwing out there: Should we do like curl and always send expect-100 headers and wait 1000ms when we have content?
>
> To do it unconditionally seems like it would be against the point of what "Expect: 100-continue" is supposed to solve, possibility to deny requests early that are large to not waste network bandwidth and processing? To do it if the request payload is bigger than X megabytes would seem reasonable to me however.

Yeah, I agree here too. [Here's what the curl docs say](https://everything.curl.dev/http/post/expect100):

> curl sends this Expect: header by default if the POST it will do is known or suspected to be larger than just minuscule. curl also does this for PUT requests.

I like this idea. We could do this for the known-length send methods like [send_bytes](https://docs.rs/ureq/latest/ureq/struct.Request.html#method.send_bytes). For [send](https://docs.rs/ureq/latest/ureq/struct.Request.html#method.send) we won't know if the body will be miniscule or not. We could assume that `send` will always be non-miniscule because the point of passing an `impl Read` is that the data be too big or too dynamic to fit in a `&[u8]` easily.

> I would love to help out some more, but the coming week will be a bit busy for me. So if you have some patience, I could continue working on this. For my use-case, the current solution works but the suggested improvements would also help.

We're happy to wait. Let's say if you come back to in the next few weeks, great; if not we'll pick up your PR and run with it. Thanks!

johan-bjareholt:
I have unfortunately been swamped the past few weeks so have not been able to work on this. Neither do I see myself having time to start working on the rest before the middle of february. If someone else has time to look at it, that'd be much appreciated.

johan-bjareholt:
Tried to rebase to fix merge conflict, but tests are failing on main again.

Pushed fix for main in seperate PR https://github.com/algesten/ureq/pull/742

src/response.rs
```diff
@@ -282,7 +287,12 @@ impl Response {
     /// # }
     /// ```
     pub fn into_reader(self) -> Box<dyn Read + Send + Sync + 'static> {
-        self.reader
+        match self.reader {
+            PendingReader::Reader(reader) => reader,
+            _ => panic!(
```
johan-bjareholt, line 292: I have seen a rare case where this occurs, have not figured out why yet. Needs to be fixed.

src/unit.rs
```diff
@@ -277,14 +284,20 @@ fn connect_inner(
     }
     let retryable = unit.is_retryable(&body);
 
+    let mut stream = stream::DeadlineStream::new(stream, unit.deadline);
+
+    if expect_100_continue {
+        let response = Response::read_response_head(&mut stream, unit)?;
+        if response.status() != 100 {
+            return Err(Error::Status(response.status(), response));
```
johan-bjareholt, line 292: This code is broken, as it returns a "response" with the stream set to PendingReader::BeforeBodyStart. This means that we will panic if someone tries to read the response body.

johan-bjareholt:
Added some unit tests, implemented support for handling 417 according to spec and added a TODO in the PR description.

src/unit.rs
```diff
@@ -277,14 +284,47 @@ fn connect_inner(
...
 
+    let mut stream = stream::DeadlineStream::new(stream, unit.deadline);
+
+    if expect_100_continue {
+        match Response::read_response_head(&mut stream, unit) {
+            Ok(mut response) => {
+                match response.status() {
+                    100 => debug!("Got 100-continue, proceeding with body"),
+                    200 => {
+                        // TODO: How should we handle this case?
```
johan-bjareholt, line 295: Any idea how to best handle this case?

johan-bjareholt:
@algesten @jsha Feel free to review this again, I think it has all the critical features now.

johan-bjareholt:
@algesten @jsha Ping :slightly_smiling_face:

algesten:
@johan-bjareholt I haven't forgotten. Just been a lot lately.

johan-bjareholt:
Sorry for nagging, just wanted to ping this again.
I understand if you still got a lot to do.

algesten:
I'm not ignoring this. I'm stalling! 🙈

algesten:
To elaborate: ureq is now a quite popular crate, and I'm largely alone in maintaining it. This PR changes some of the inner workings and I've increasingly become more and more hesitant to do such things (don't know if you saw the fallout from trying to fix our test cases with `hootbin`)

Which isn't to say we don't want it, but I have this inertia/emotional block to get over.

johan-bjareholt:
I understand.
Anything I can help to get you to overcome it? Any code that these changes might impact that we need more unittests for?
I've been using this patch continuously for a few months now without any issues at least with over 160 million requests and terrabytes of data, if that's gives any comfort. My use-case might be very specific though.

algesten:
@johan-bjareholt

The ureq 3.x rewrite which now is in main should support expect-100-continue. Sorry for messing you around on this PR.

johan-bjareholt:
It's ok, what's most important is that it's now supported. Thanks!

PR #681, opened 2023-11-20
feat: add functions to return the list of headers and their contents

### LINKED ISSUE #655: Getting the list of all the headers and their content
Hello,

I'm working on an API code generator.
I want all the generated functions to return an access to all the headers and the parsed body.

To return all the headers I currently loop over `response.headers_names()` and then read the `response.header(&name)` :

```rust
pub(crate) fn get_header_map(response: &ureq::Response) -> HashMap<String, String> {
    let mut headers = HashMap::new();

    let names = response.headers_names();
    for name in names {
        if let Some(value) = response.header(&name) {
            headers.insert(name, value.to_string());
        }
    }
    headers
}
```

I find it very inefficient, the most efficient way from my point of view would be to add an access to the `headers` in the `Response` struct.

Do you think this would cause problems ? I can try to open a PR adding this function.

Edit: I found the "http-interop" feature, this could give me the access I want but I don't like the fact that `unwrap()` is called on `into_string()`.

nponsard:
This PR adds a function to get a reference to the list of the headers stored in the `Response`. 

Fixes #655.

Let me know if you have any remarks.

I was also thinking about having a function return a map where the index is the name of the header, but with this PR it can be easily done afterwards (outside of this library).

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -183,6 +183,33 @@ impl Response {
             .collect()
     }
 
+    /// Returns a list of all headers in this response.
+    /// First element of the tuple is the header name, second is the value.
+    pub fn headers(&self) -> Vec<(String, Option<String>)> {
+        self.headers
+            .iter()
+            .map(|h| (h.name().to_owned(), h.value().map(|v| v.to_owned())))
+            .collect()
+    }
+
+    /// Iterate over all headers in this response.
+    /// First element of the tuple is the header name, second is the value.
+    /// This avoids allocating a vector.
+    pub fn headers_iter(&self) -> impl Iterator<Item = (&str, Option<&str>)> + '_ {
+        self.headers.iter().map(|h| (h.name(), h.value()))
+    }
+
+    /// Returns a hashmap of all headers in this response.
+    /// The keys are the header names, and the values are the header values.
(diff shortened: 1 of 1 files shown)


johan-bjareholt:
@nponsard Considering the following change, I assume that this PR is not relevant anymore? https://github.com/algesten/ureq/commit/4ddcc3a27a6c8cb88041dc713010455f776fa753

nponsard:
Ah yes, this commit breaks what I've done.

It would be easy to make the Header struct public again, but that depends on the objectives of this project / what the maintainers want.

algesten:
Headers in ureq public API are `&str` for both key and value. The `Header` struct is internal and I don't think we should make it public here, because if it's public in one part of the API, why isn't it used everywhere?

* https://docs.rs/ureq/latest/ureq/struct.Response.html#method.header
* https://docs.rs/ureq/latest/ureq/struct.Request.html#method.set

The variant of this PR I would consider would be `headers() -> (String, String)` (or maybe `header_iter() -> impl Iterator<Item = (&str, &str)>` if we want to avoid allocation)

algesten:
Closing since we are moving to 3.x. The HTTP crate API will have the functions required.

Issue #683, opened 2023-11-21
Updating `http` was a breaking change


Nemo157:
When using the `http-interop` feature the `http` crate is part of the public API of `ureq`, updating to 2.9.0 with a crate using `http` 0.2 now fails.

```console
> cargo dl -e cargo-dl@=0.1.4 && cd cargo-dl-0.1.4
                         cargo-dl@=0.1.4   extracted cargo-dl 0.1.4 to cargo-dl-0.1.4

> cargo check
...
    Checking cargo-dl v0.1.4 (/tmp/scratch.rust.2023-11-21T09:46.dJ6oRl/cargo-dl-0.1.4)
    Finished dev [unoptimized + debuginfo] target(s) in 13.65s

> cargo update -p ureq
    Updating crates.io index
      Adding http v1.0.0
    Removing rustls-webpki v0.100.1
    Updating ureq v2.7.1 -> v2.9.0
    Updating webpki-roots v0.23.1 -> v0.25.2

> cargo check
    Checking http v1.0.0
    Checking webpki-roots v0.25.2
    Checking ureq v2.9.0
    Checking cargo-dl v0.1.4 (/tmp/scratch.rust.2023-11-21T09:46.dJ6oRl/cargo-dl-0.1.4)
error[E0277]: the trait bound `ureq::Request: From<http::request::Builder>` is not satisfied
   --> src/main.rs:104:27
    |
104 |             let request = ureq::Request::from(request).set("User-Agent", USER_AGENT);
    |                           ^^^^^^^^^^^^^ the trait `From<http::request::Builder>` is not implemented for `ureq::Request`
    |
    = help: the following other types implement trait `From<T>`:
              <ureq::Request as From<http::request::Parts>>
              <ureq::Request as From<http::request::Builder>>

error[E0277]: the trait bound `http::Response<Box<dyn std::io::Read + Send + std::marker::Sync>>: From<ureq::Response>` is not satisfied
   --> src/main.rs:106:51
    |
106 |             let response = read_response(response.into())?;
    |                                                   ^^^^ the trait `From<ureq::Response>` is not implemented for `http::Response<Box<dyn std::io::Read + Send + std::marker::Sync>>`
    |
    = note: required for `ureq::Response` to implement `Into<http::Response<Box<dyn std::io::Read + Send + std::marker::Sync>>>`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `cargo-dl` (bin "cargo-dl") due to 2 previous errors

[...]

PR #684, opened 2023-11-21
Support both http 0.2 and 1.0

Nemo157:
This implements the idea I had in #683, supporting both versions of `http` and with the planned breaking change to supporting `TryFrom<http::request::Builder>` instead of `From<http::request::Builder>` since it's non-breaking as a new feature.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,7 +59,9 @@ jobs:
           - native-certs
           - gzip
           - brotli
+          - http
           - http-interop
+          - http http-interop
     env:
       RUST_BACKTRACE: "1"
       RUSTFLAGS: "-D dead_code -D unused-variables -D unused"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -261,6 +261,17 @@ checksum = "d77f7ec81a6d05a3abb01ab6eb7590f6083d08449fe5a1c8b1e620283546ccb7"
 
 [[package]]
 name = "http"
+version = "0.2.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8947b1a6fad4393052c7ba1f4cd97bed3e953a95c79c92ad9b051a04611d9fbb"
+dependencies = [
+ "bytes",
+ "fnv",
+ "itoa",
+]
+
+[[package]]
+name = "http"
 version = "1.0.0"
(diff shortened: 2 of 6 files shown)


algesten:
Thanks!

The problem with this solve is that the implementations are mutually exclusive. That means if some project get diamond dependencies on `ureq` with different feature flags, the compilation might break completely.

```
cargo test --features http http-interop
```

The immediate problem can of course make one of the flags "win", but that wouldn't help impls that depend on one that doesn't win.

What are the drivers for solving the retaining of the old dependency?

Nemo157:
> The problem with this solve is that the implementations are mutually exclusive.

They aren't, they're entirely independent. `cargo check --all-features` works fine and using both features together from one build will work.

> What are the drivers for solving the retaining of the old dependency?

Avoiding breakage of code relying on the previous API. If it had been documented that the `http-interop` feature wasn't going to be semver compatible then I would have avoided using it in my crate and just performed the conversion manually. I assume my crate is not going to be the only one hit by this incompatibility, I'm likely just the first to notice as I have a CI job that tests against updated dependencies daily.

Nemo157:
Hmm, it's less than I thought, if I did this correctly these should be the only (published) crates that are likely currently broken with ureq 2.9.0:

```console
> fd -0tf -Econfig.json '' crates.io-index | xargs -0 tail -qn1 | jq -r '.name as $name | .deps[] | select(.name == "ureq" and (.features | contains(["http-interop"]))) | $name'
crates-index
cargo-dl
cargo-features-manager
cargo-upgrades
```

algesten:
Alright. Let's try to land this.

Would it be possible to duplicate `src/http_interop.rs` and have one file for http1 and one for http02, and that would give  us a smaller diff?

Nemo157:
Done. It doesn't really reduce the diff but it does probably make it more readable.

lolgesten:
I'll fix the rest here. Thanks!

algesten:
I'll do the rest here. Thanks!

algesten:
In released version 2.9.1, renamed the `http` feature to `http-crate` since in the context of `ureq`, the name "http" is a bit too generic.

MarijnS95:
Did you forget to remove this planned removal from https://github.com/algesten/ureq/blob/main/FUTURE.md, after borrowing the suggested implementation from it?

PR #685, opened 2023-11-21
new feature flag to control whether or not to detect proxy settings from environment by default

toymil:
The default behavior of many libs and tools is to pick up proxy environment variables.  Being able to config `ureq` to behave the same way makes implementing this behavior in other software easier.

If the default is `false` and we have no ways to make it default to `true`, implementing this behavior in other software would mean one of the following:
- no qol methods like `ureq::get(..)`, because we need to construct an `AgentBuilder` and call `try_proxy_from_env`, then `build`, every time.
- construct and config a shared or global `Agent`, use that throughout the code.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -325,8 +325,9 @@ impl AgentBuilder {
 
     /// Attempt to detect proxy settings from the environment, i.e. HTTP_PROXY
     ///
-    /// The default is `false`, i.e. not detecting proxy from env since this is
-    /// a potential security risk.
+    /// The default is `false` without the `proxy-from-env` feature flag, i.e.
+    /// not detecting proxy from env, due to the potential security risk and to
+    /// maintain backward compatibility.
     ///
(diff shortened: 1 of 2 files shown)


src/agent.rs
```diff
@@ -322,8 +325,8 @@ impl AgentBuilder {
 
     /// Attempt to detect proxy settings from the environment, i.e. HTTP_PROXY
     ///
-    /// The default is `false`, i.e. not detecting proxy from env since this is
-    /// a potential security risk.
+    /// The default is `false` without the `proxy-from-env` feature flag, i.e.
+    /// not detecting proxy from env, to maintain backward compatibility.
```
algesten, line 329: Having automatic proxy detection _is_ a security risk. I want you to rephrase this doc to keep the warning.

algesten (changes requested):

algesten (approved):
Thanks!

PR #690, opened 2023-12-01
Update to rustls 0.22

jsha:
DO NOT MERGE until rustls 0.22 is actually released. Right now this uses a Cargo.toml patch to get the latest version before it's released.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = ["tls", "gzip"]
-tls = ["dep:webpki", "dep:webpki-roots", "dep:rustls"]
+tls = ["dep:webpki", "dep:webpki-roots", "dep:rustls", "dep:rustls-pki-types"]
 native-certs = ["dep:rustls-native-certs"]
 json = ["dep:serde", "dep:serde_json"]
 charset = ["dep:encoding_rs"]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -44,8 +44,9 @@ encoding_rs = { version = "0.8", optional = true }
 cookie_store = { version = "0.20", optional = true, default-features = false, features = ["preserve_order"] }
 log = "0.4"
 webpki = { package = "rustls-webpki", version = "0.101", optional = true }
```
messense, line 46: Consider also update `rustls-webpki` to 0.102?

Revertron:
`rustls 0.22` has been released!
Can we update ureq? :)
(I need the MIPS support added by rustls)

algesten:
@jsha once it's failing on httpbin calls, you need to keep rerun only the failed tests many times to "grind" it through.

cpu:
>  I'm really scratching my head on what that method should return.

@amunra If you want your `ServerCertVerifier` to act as it did prior to Rustls 0.22 when there was a default impl of `supported_verify_schemes` you can change the body to defer to the *ring* based provider's supported verify schemes:

```rust
fn supported_verify_schemes(&self) -> Vec<rustls::SignatureScheme> {
    rustls::crypto::ring::default_provider()
        .signature_verification_algorithms
        .supported_schemes()
}
```

emilk:
I would appreciate a patch release with this :pray:

aeyoll:
hello @algesten, can you create a new release with this fix please? Thanks a lot!

algesten:
Alright. Let's do it.

PR #695, opened 2023-12-25
Do not expose crate-private Header struct in public interface

alekitto:
Commit https://github.com/algesten/ureq/commit/4ddcc3a27a6c8cb88041dc713010455f776fa753 made `Header` struct internal (crate-private), but is was used in some non-exported functions maked as public. This raises an error while compiling the library.
Marking the functions as crate-private (as the `Header` struct) fixes the compilation.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -152,7 +152,7 @@ impl Header {
 }
 
 /// For non-utf8 headers this returns [`None`] (use [`get_header_raw()`]).
-pub fn get_header<'h>(headers: &'h [Header], name: &str) -> Option<&'h str> {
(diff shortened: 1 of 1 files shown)


algesten:
Hi @alekitto, welcome to ureq!

I'm curious, how do you compile ureq to trigger this error?

alekitto:
In one of my projects I needed rustls 0.22 support, so I've included the library explicitly targeting the git repository and it simply failed.

nponsard:
I have the same problem when trying to build a simple program using ureq on the commit 26ed596c5a411be0126682c8a508ff349a5d4c75. Doing the same changes fixes it.

algesten:

PR #696, opened 2023-12-26
Fix some compilation problems due to Header

algesten:


johan-bjareholt:
@algesten Could you please integrate this?

algesten:
Right. Let's fix this.

Issue #686, opened 2023-11-23
How to handle "native_tls::HandshakeError::WouldBlock"?


johan-bjareholt:
I am getting the following error on a few requests:
https://github.com/algesten/ureq/blob/89a69a58344cf19e57d618fa4df29062c18e4b7e/src/ntls.rs#L21

It is racy and timing dependent. According to the description of the native_tls documentation, it doesn't sound like this issue should ever occur?

> A stream interrupted midway through the handshake process due to a WouldBlock error.
>
> Note that this is not a fatal error and it should be safe to call handshake at a later time once the stream is ready to perform I/O again.

https://docs.rs/native-tls/latest/native_tls/enum.HandshakeError.html#variant.WouldBlock

How are you supposed to handle this in ureq? To me I would expect this request to rather retry again and possibly timeout, but now it just instantly fails.

jsha:
Huh that's surprising. I think you should never get this so long as the provided `io` performs blocking I/O (like a TcpStream).

Could you share your code?

johan-bjareholt:
I found the issue, it happens when there's a really really slow server and the request times out while trying to do a tls handshake.

If you run the server and then start two clients at the same time, one of them will fail with "TimedOut" and the other with "WouldBlock", as the first one that starts will timeout reading the response while the other one times out during the tls handshake.

So it's not really a "bug" per se, ureq does what it is supposed to, but the "WouldBlock" error type is very confusing.

client

```rust
static ROOT_CERT_PEM : &str = r#"-----BEGIN CERTIFICATE-----
MIIDcTCCAlmgAwIBAgIURAwA7tQhPBDBQ88AeTljaZjwCbUwDQYJKoZIhvcNAQEL
BQAwTzELMAkGA1UEBhMCU0UxDzANBgNVBAgMBlNjYW5pYTENMAsGA1UEBwwETHVu
ZDENMAsGA1UECgwEQXhpczERMA8GA1UEAwwIdGVzdC5jb20wHhcNMjMxMTEyMTUy
MTA5WhcNMjQxMTEyMTUyMTA5WjBPMQswCQYDVQQGEwJTRTEPMA0GA1UECAwGU2Nh
bmlhMQ0wCwYDVQQHDARMdW5kMQ0wCwYDVQQKDARBeGlzMREwDwYDVQQDDAh0ZXN0
[...]

Issue #688, opened 2023-11-24
Replace httpbin.org in all examples and tests


algesten:
It's become almost impossible to get a clean build through the github actions.

The error is a bit strange (which should be looked at in isolation), but it basically boils down to `Connection reset by peer`. I don't know if they are maybe throttling. Re-running failed github actions means it's possible to "grind" through the tests until all are green.

```
stderr:
Error: Transport(Transport { kind: Io, message: None, url: Some(Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("httpbin.org")), port: None, path: "/post", query: None, fragment: None }), source: Some(Custom { kind: ConnectionReset, error: Transport(Transport { kind: Io, message: Some("Error encountered in a header"), url: None, source: Some(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }) }) }) })
```

We need a better solve for this. Something not depending on 3rd parties.

mcr:
I experienced httpbin.org failing on my laptop, while it worked on my desktop
for sometime. Same network. Different NAT44, IPv6.  I never figured it out.

nponsard:
One solution would be to start an httpbin container with the CI. There would be a problem with the HTTPS certificate verification.
This way we are sure that when GitHub CI is available, tests can be run against this local container.

jsha:
My preferred solution is to port more of the tests and doctests to use `test_agent`, which starts a little HTTP server within the test runner and talks to that.

nponsard:
Oh I didn't know there was this in the project.
I've looked at what would be needed to resolve this issue and I have some questions: 

- Do the examples in the `examples` folder have to be changed ? 
- In the doctests, does the initialization of the test server need to be shown ?

algesten:
I've been busy the last couple of weekends exploring a new API for "correct" no_std HTTP library. I call it `hoot` (https://crates.io/crates/hoot). I have addressed this issue using it.

PR #699, opened 2024-01-04
ntls: Convert WouldBlock to ConnectionFailed

johan-bjareholt:
When the tls handshake starts and the server doesn't respond, the std::net::TcpStream will return WouldBlock or TimedOut when a read/write times out to the socket, which in turn native_tls will convert to a HandshakeError::WouldBlock.

Solves #686

### DIFF
--- a/src/ntls.rs
+++ b/src/ntls.rs
@@ -17,9 +17,12 @@ impl TlsConnector for native_tls::TlsConnector {
                 native_tls::HandshakeError::Failure(e) => ErrorKind::ConnectionFailed
                     .msg("native_tls connect failed")
                     .src(e),
-                native_tls::HandshakeError::WouldBlock(_) => {
-                    ErrorKind::Io.msg("Unexpected native_tls::HandshakeError::WouldBlock")
-                }
+                native_tls::HandshakeError::WouldBlock(_) => ErrorKind::Io
+                    .msg("native_tls handshake timed out")
+                    .src(std::io::Error::new(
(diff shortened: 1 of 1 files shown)


src/ntls.rs
```diff
@@ -17,9 +17,9 @@ impl TlsConnector for native_tls::TlsConnector {
                 native_tls::HandshakeError::Failure(e) => ErrorKind::ConnectionFailed
                     .msg("native_tls connect failed")
                     .src(e),
-                native_tls::HandshakeError::WouldBlock(_) => {
-                    ErrorKind::Io.msg("Unexpected native_tls::HandshakeError::WouldBlock")
-                }
+                native_tls::HandshakeError::WouldBlock(_) => ErrorKind::ConnectionFailed
```
johan-bjareholt, line 20: Will change this to ErrorKind::Io with ErrorKind::TimedOut

johan-bjareholt:
The tests were broken on master, fixed it in a seperate commit.

johan-bjareholt:
> The tests were broken on master, fixed it in a seperate commit.

Seems to be the same thing as was fixed in https://github.com/algesten/ureq/pull/696, but not yet merged.

algesten:
It's merged now, but I'm trying to fix this, so accepting PRs are a bit stalled: https://github.com/algesten/ureq/issues/688

algesten (approved):

algesten:
Thanks!

PR #702, opened 2024-01-10
fix: use reader for bytes http response transformation

rogercoll:
The From implementation to transform a ureq::Response into an http::Response<Vec<u8>> is using the into_string() method. This method only allows for UTF-8 characters in the body payload, thus the result is inconsistent for non UTF-8 characters like protobuffers (binary).

The proposed solution uses the reader instead of transforming the input into a UTF-8 character.

### DIFF
--- a/src/http_crate.rs
+++ b/src/http_crate.rs
@@ -122,9 +122,10 @@ impl From<Response> for http::Response<String> {
 /// ```
 impl From<Response> for http::Response<Vec<u8>> {
     fn from(value: Response) -> Self {
-        create_builder(&value)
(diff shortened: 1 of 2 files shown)


algesten:
Thanks!

I will get my head around this soon. Currently I'm blocked on trying to fix: https://github.com/algesten/ureq/issues/688

rogercoll:
Hi @algesten, any update on this small fix? Thank you!

algesten (approved):
Yeah. Going via utf-8 seems just wrong. Thanks!

PR #718, opened 2024-02-03
Implement Happy Eyeballs connection RFC

### LINKED ISSUE #535: Implement Happy Eyeballs to workaround slow/broken IPv6 connections
[Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs) is an algorithm which can help deal with networks with misconfigured or very slow IPv6. It is implemented in most web browsers as well as curl, but not wget. It works by making a delayed IPv4 request in parallel with a IPv6 request after a small delay. This could help smooth over all sorts of potentially issues with IPv6, which are actually reasonably likely because adoption is still very much incremental.

As per https://github.com/JuliaLang/juliaup/issues/340 -- because this is not implemented, some users of this library can experience an issue where files can be downloaded by web browser, but not using this library.

It seems to also be implemented in Python https://bugs.python.org/issue33530

mrkline:
Races interleaved IPv4 and IPv6 connections to provide the fastest one in cases where certain addresses or address families might be blocked, broken, or slow. (See <https://datatracker.ietf.org/doc/html/rfc8305>)

ureq strives for simplicity, and avoids spawning threads where it can, but - like with SOCKS - there's no way around it here.
Some mini internal async executor (discussed in https://github.com/algesten/ureq/issues/535#issuecomment-1229433311) wouldn't help - `connect()` is a blocking syscall with no non-blocking alternative. (Big async runtimes like Tokio "solve" this problem by keeping a pool of OS threads around for just these sorts of blocking calls.) We _could_ have some thread pool (a la rayon) to avoid spawning threads on each connection attempt, but spawning a few threads is a cheap operation compared to everything else going on here. (DNS resolution, handshaking across the Internet...)

Much of this implementation was inspired by attohttpc's:
<https://github.com/sbstp/attohttpc/blob/master/src/happy.rs>

Fixes #535

### DIFF
--- a/src/eyeballs.rs
+++ b/src/eyeballs.rs
@@ -1,0 +1,163 @@
+//! A Happy Eyeballs RFC implementation
+//!
+//! Races interleaved IPv4 and IPv6 connections to provide the fastest connection
+//! in cases where certain addresses or address families might be blocked, broken, or slow.
+//! (See <https://datatracker.ietf.org/doc/html/rfc8305>)
+//!
+//! ureq strives for simplicity, and avoids spawning threads where it can,
+//! but - like with SOCKS - there's no way around it here.
+//! Some mini internal async executor
+//! (discussed in <https://github.com/algesten/ureq/issues/535#issuecomment-1229433311>)
+//! wouldn't help - `connect()` is a blocking syscall with no non-blocking alternative.
+//! (Big async runtimes like Tokio "solve" this problem by keeping a pool of OS threads
+//! around for just these sorts of blocking calls.)
+//! We _could_ have some thread pool (a la rayon) to avoid spawning threads
+//! on each connection attempt, but spawning a few threads is a cheap operation
+//! compared to everything else going on here.
+//! (DNS resolution, handshaking across the Internet...)
+//!
+//! Much of this implementation was inspired by attohttpc's:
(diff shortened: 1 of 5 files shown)


algesten:
Hi @mrkline, welcome to ureq!

This is exciting, and a nice tidy-up. I actually have no idea what ureq users in general think about spawning more threads. It might be we've been unnecessarily conservative. We might want to have another `simple::connect` which behaves like today and make it possible to chose behavior in the `Agent`. However if we do that, I don't mind defaulting to `eyeballs::connect`, since I think it's a better solution.

mrkline:
Thanks for the kind intro! I'll defer to you - if you want the behavior to be selectable in `Agent`, let me know.

rpodgorny:
hello, what is the plan on merging this? thanks.

algesten:
> hello, what is the plan on merging this? thanks.

I don't want to merge this in ureq 2.x because the connection code already there is too complex. 

The rewrite in ureq3 branch is almost ready. I'm planning on releasing a beta very soon. 

The happy eyeball strategy will fit much better into ureq3 where the transport is completely plugin. It can even be released as a separate crate.

algesten:
Closing since we are moving to 3.x.

PR #719, opened 2024-02-04
Add HTTPS proxy support

DanGould:
HTTPS tends to be the default these days, and I would like to proxy via some server that are HTTPS only using ureq.

I ran into a small issue where we're calling [`Response::do_from_stream`](https://github.com/algesten/ureq/blob/2300a5b3787b1ac800314dc47f075f4850026bb3/src/stream.rs#L458) which takes ownership of the `tls_stream` that we ultimately need to return, and there's no easy way to clone a `Box<dyn ReadWrite>` returned from `tls_conf.connect(..)`, but it seems surmountable. The old PR didn't have this problem because I the `Stream` abstraction wasn't being used in `connect_host`. Any suggestion to address this problem?

This reopens #544

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -9,6 +9,7 @@ use crate::{
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum Proto {
     HTTP,
+    HTTPS,
     SOCKS4,
     SOCKS4A,
     SOCKS5,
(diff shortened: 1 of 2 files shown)


DanGould:
force pushed to fix clippy lints

algesten:
Closing since we are moving to 3.x. I believe this is solved in ureq 3.x

PR #725, opened 2024-02-10
Use std::sync::OnceLock instead of once_cell::sync::Lazy

rmg-x:
This PR removes the `once_cell` dependency and uses the std struct instead.
MSRV would go up to 1.70

See:
- https://doc.rust-lang.org/std/sync/struct.OnceLock.html
- https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#oncecell-and-oncelock

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -845,7 +845,6 @@ dependencies = [
  "http 1.0.0",
  "log",
  "native-tls",
- "once_cell",
  "rustls",
  "rustls-native-certs",
  "rustls-pemfile",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,6 @@ proxy-from-env = []
 [dependencies]
 base64 = "0.21"
 cookie = { version = "0.18", default-features = false, optional = true }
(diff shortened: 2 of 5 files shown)


algesten:
Hi @rmg-x, welcome to ureq!

I do agree it would be nice to do this, but at this point we can't bump the MSRV. We decided on a rather conservative version requirement since our users have that need. See this issue from just two weeks ago: https://github.com/algesten/ureq/issues/712 - this was due to trying to use let-else syntax.

DanGould:
Second the msrv concern, other than just being simple and correct, the reason I use ureq is because it builds on my target msrv 1.63.0

rmg-x:
I can understand the concerns (because who wants their project to break?) but personally think it's the wrong approach for this library.

Eventually there will be a point where the MSRV is bumped and users will complain. I think it makes more sense to put the onus on people requiring a lower version (e.g. fork or use an older ureq version)

Regardless, I appreciate the responses and will leave the PR open, feel free to close it.

algesten:

PR #728, opened 2024-02-13
Exclude unnecessary files from being packaged

michel-slm:
These files are unnecessary, and some cause issues (e.g. when packaged in Fedora, having shell scripts creates a package dependency on Bash)


algesten:
This seems like a no brainer. Thanks!!!

PR #733, opened 2024-02-19
Added support for storing default headers in an `Agent`.

### LINKED ISSUE #732: Add support for configuring default headers in AgentBuilder
This would help if an agent builder wants to set a default set of headers (maybe an `Accept` header) which should persist across requests unless overwritten on a per-request basis, in a similar way to how user agents are managed currently. I’m happy to submit a pr.

arihant2math:
Use the `.headers` function in the AgentBuilder to set the default headers. These can be overwritten by per-request headers. Note that there currently isn’t a way to remove them from an already built agent.

Resolves #732

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -593,8 +593,32 @@ impl AgentBuilder {
     /// List of default headers to use when making a request.
     ///
     /// This can be overriden on a per-request basis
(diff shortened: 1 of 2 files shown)


algesten:

PR #735, opened 2024-02-28
Allow setting custom agent when converting from `http` crate types

Jasper-Bekkers:
I encountered a case where we needed to support custom ssl certificates as well as convert between the http crate as ureq. Because the conversion routines call into `crate::agent` directly this isn't possible. This PR adds helpers so I can do that, however, I'm not sure they're up to the standard of how `ureq` would do these things, so I'd love some pointers to help clean up this PR.

### DIFF
--- a/src/http_interop.rs
+++ b/src/http_interop.rs
@@ -188,26 +188,31 @@ impl From<Response> for http::Response<Vec<u8>> {
 /// ```
 impl From<http::request::Builder> for Request {
     fn from(value: http::request::Builder) -> Self {
-        let mut new_request = crate::agent().request(
-            value.method_ref().map_or("GET", |m| m.as_str()),
-            &value
-                .uri_ref()
-                .map_or("https://example.com".to_string(), |u| u.to_string()),
-        );
-
-        if let Some(headers) = value.headers_ref() {
-            for (name, value) in headers {
-                let mut raw_header: Vec<u8> = name.to_string().into_bytes();
-                raw_header.extend(b": ");
-                raw_header.extend(value.as_bytes());
-                let header = HeaderLine::from(raw_header).into_header().unwrap();
-
-                crate::header::add_header(&mut new_request.headers, header)
-            }
-        }
-
-        new_request
+        convert_from_http_builder_with_agent(value, crate::agent())
     }
+}
+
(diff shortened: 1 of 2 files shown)


src/http_interop.rs
```diff
@@ -188,26 +188,31 @@ impl From<Response> for http::Response<Vec<u8>> {
...
+}
 
-        if let Some(headers) = value.headers_ref() {
-            for (name, value) in headers {
-                let mut raw_header: Vec<u8> = name.to_string().into_bytes();
-                raw_header.extend(b": ");
-                raw_header.extend(value.as_bytes());
-                let header = HeaderLine::from(raw_header).into_header().unwrap();
+/// Converts an [`http::request::Builder`] into a [`Request`] with a custom [`Agent`].
+pub fn convert_from_http_builder_with_agent(value: http::request::Builder, agent: crate::Agent) -> Request {
```
tosti007, line 196: `agent` is used only for the `&self` `request` call, so maybe take an `&crate::Agent` here?

src/http_interop.rs
```diff
@@ -227,23 +232,29 @@ impl From<http::request::Builder> for Request {
...
+    }
+}
 
-        for (name, value) in &value.headers {
-            // TODO: Aren't complete header values available as raw byte slices?
-            let mut raw_header: Vec<u8> = name.to_string().into_bytes();
-            raw_header.extend(b": ");
-            raw_header.extend(value.as_bytes());
+/// Converts a [`http::request::Parts`] into a [`Request`] with a custom [`Agent`].
+pub fn convert_from_http_parts_with_agent(value: http::request::Parts, agent: crate::Agent) -> Request {
```
tosti007, line 240: Same here regarding `&crate::Agent`.

src/http_interop.rs
tosti007, line 196: Maybe this (and the from parts variant) should be an associated function on either `Agent` or `Request`? I.e.: `agent.request_builder(builder)`, or `Request::from_builder(builder)` and `Request::from_builder_with_agent(builder, agent)`.

algesten:
I think the cleaner way to do this would be with an extension trait.

I.e. 

```rust
pub trait HttpTraitInterop {
     fn to_ureq_with_agent(agent: Agent) -> ureq::Request;
}
```

which we then implement for `http::Request/Builder` etc

PR #740, opened 2024-03-14
Provide Url getter on Response

phy1729:
It would be nice to have access to the `Url` struct itself, so I don't need to reparse the String and use `expect`.


algesten:
This is deliberately an internal concern for ureq. I do not wish to make `Url` part of ureq's public API.

Issue #741, opened 2024-03-21
Middleware cannot modify query string parameters?


benferse:
I tried to write a simple middleware that adds a query string parameter to an outgoing request, but it didn't appear to get included in what actually hit the wire:

```rust
struct Foo {}

impl Middleware for Foo {
  fn handle(&self, request: Request, next: MiddlewareNext,) -> Result<ureq::Response, ureq::Error> {
    next.handle(request.query("foo", "bar"))
  }
}
```

Looking at the logic in `Request::do_call` this seems intentional - the url that gets captured by the closure that ends up calling `Unit::connect` is parsed _before_ the middleware chain is run. Is this an oversight, or is modifying the URL in middleware not supported for some reason?

benferse:
I made a local modification to the closure to parse the URL just in time, which works as expected. `parse_url` is fallible, but that closure's signature already allows a failure for the same `Error` type, so it seems like a reasonable fix.

algesten:
Can't imagine a reason we shouldn't support that. Can you make a PR for it?

PR #745, opened 2024-03-27
fix: Determine the URL to connect to after middleware finishes

benferse:
Make sure that the URL that's used to establish the transport connection is parsed after middleware runs, in case the middleware made any changes to the URL itself (added query string params, changed the URL itself, etc.)

Fixes: #741


algesten:
Thanks!

Issue #751, opened 2024-04-22
Decide what to do about rustls defaulting to aws-lc-rs instead of ring


Diggsey:
`rustls` now requires `aws-lc` as a default dependency, which doesn't build without a load of extra dependencies. This crate should specify `default-features = false` so that end users can continue to use the `ring` backend of `rustls`

algesten:
Hi @Diggsey, welcome to ureq!

I haven't made up my mind on how to deal with this situation. ureq was built on the idea of having very few dependencies. When `rustls` changed to `aws-lc-rs`, I think we end up with some tough decisions. There might be some way forward I can't see, so I'm happy for input. Observations in no particular order:
 
* Rustls wants the application to set their process default provider, libraries like ureq should not [touch this](https://docs.rs/rustls/0.23.5/rustls/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider)
* Rustls defaults to `aws-lc-rs` and no `ring`, while `rustls-webpki` [defaults to `ring`](https://github.com/rustls/webpki/blob/7fc65f1f1ed61ea5c2faa3a779ceb79252047587/Cargo.toml#L75) and no `aws-lc-rs` (as of 23 Apr 2024).
* Ureq depends on both `rustls` and `rustls-webpki`, which means if we do nothing, we get both `aws-lc-rs` and `ring` (which goes against the idea of having minimal amount of deps)
* aws-lc-rs claims to not require `cmake` for non-fips builds [as of 1.7.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.7.0)
* I still see cmake as a build dep despite aws-lc-rs being 1.7.0, pulled in via `aws-lc-sys`. Unclear what they mean by (aws-lc-rs not requiring cmake)[https://github.com/aws/aws-lc-rs/blob/4216305d6fdff26d95713fad32d45d3f75b44ef5/aws-lc-sys/Cargo.toml#L54] (as of 23 Apr 2024)
* Additional problems around `nasm` being [required on windows](https://github.com/algesten/ureq/issues/751)

```text
$ cargo tree -i cmake
cmake v0.1.50
[build-dependencies]
└── aws-lc-sys v0.15.0
    └── aws-lc-rs v1.7.0
```

I've had no ureq user asking for us to upgrade to rustls 0.23 – obviously just a matter of time though. I hope we can stall and have a clearer way forward once the dust settles.

ctz:
[...]

PR #753, opened 2024-04-23
rustls default to ring backend

algesten:
rustls 0.23 changes the default crypto backend from ring to aws-lc-rs.
aws-lc-rs has some additional build requirements like `cmake` and
`nasm` on windows.

ureq will use feature flags to enable ring until such time rustls
builds more easily out-of-the-box with default features.

Related to #751 

Related recommendation for using Ring instead of rustls default: https://github.com/rustls/rustls/issues/1913#issuecomment-2071760244


mnpw:
Thanks for this! I would like to see this change in the release. When is this planned to be merged?

algesten:
> Thanks for this! I would like to see this change in the release. When is this planned to be merged?

I think I can merge and release this weekend.

PR #755, opened 2024-04-30
Enable `http-crate` feature while building docs

MarijnS95:
`ureq` properly adheres to the additive-only-ness of crate feature flags, allowing interop with the `http 0.2` and `http 1.x` crates concurrently. We should take advantage of this in the docs as well, by generating entries for `http 1.x` `impl`s (with appropriate autogenerated `Available on create feature http-crate only` markers) to make users aware of the bindings they have available.

Also clean up the `http 1` integration a bit following the breaking changes that I proposed (for `http 0.2`) a long time ago, which seem to have been copy-pasted to the new `http 1` interop.


algesten (approved):
Thanks!

PR #757, opened 2024-05-13
Allow skipping the buffer for large reads

davibe:
This mirrors the behaviour of the underlying `std::io::BufReader` allowing it it
to work as intended


algesten:

PR #759, opened 2024-05-22
Rustls supports IP SANs since 0.21.0

aveyrenc:
Rustls is able to validate IP SANs in certificates since version [0.21.0](https://github.com/rustls/rustls/releases/tag/v%2F0.21.0), so no more need to divert users from using it for this usage.


algesten (approved):
Sounds good! Thanks!

PR #760, opened 2024-06-10
Headers from iterator type

SergioRibera:
This added method is intended to make it easier to add multiple headers.

The name of the function is put in dicusion if you see it convenient.

## Motivation
I recently came across some code similar to this online and realized that ureq did not have an easy way to do this and that it resulted in code that was not very performant or readable (even acceptable).

```rs
for header in headers.iter(){
  req = req.set(&header.name, &header.value)
}
```

> [!NOTE]
> It would be ideal to explore a way to implement more From<T> for converting headers from a variety of other types.

I understand that this is probably only applicable to very particular cases, in that case and if you do not see your integration in `ureq` as convenient, feel free to comment and/or reject the PR.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -312,6 +312,43 @@ impl Request {
         self
     }
 
+    /// Set multiple header fields.
+    ///
+    /// ```
+    /// # fn main() -> Result<(), ureq::Error> {
+    /// # ureq::is_test(true);
+    /// let headers = [
+    ///     ("Accept", "text/plain"),
+    ///     ("Range", "bytes=500-999")
+    /// ];
+    /// let resp = ureq::get("http://httpbin.org/bytes/1000")
+    ///     .set_headers(&headers)
+    ///     .call()?;
+    ///
+    /// let headers_vec = vec![
+    ///     ("Accept", "text/plain"),
+    ///     ("Range", "bytes=500-999")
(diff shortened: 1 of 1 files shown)


SergioRibera:
@algesten 

Sorry for the mention, but I wanted to know what you think of this PR

algesten:
@SergioRibera thanks for the PR.

As a side note, the rest of Rust API is deliberately using very simple types, i.e. `&str` instead of `T:  AsRef<str>` – I might change this in a future ureq 3.0, but for now we should keep the API tight.

However I'm in two minds about landing it. It increases the API surface area I must maintain for something that is a convenience function.

SergioRibera:
> As a side note, the rest of Rust API is deliberately using very simple types, i.e. `&str` instead of `T: AsRef<str>` – I might change this in a future ureq 3.0, but for now we should keep the API tight.

Perfect, now I change it, I had it planned to be usable by any type that can be referenced, but I understand the point.

> However I'm in two minds about landing it. It increases the API surface area I must maintain for something that is a convenience function.

Although this solution lends itself to be a solution more aligned to the way it is developed in Rust, I understand your point, so there is no problem if you reject the contribution.

By the way, if you have a milestone for version 3.0 I can check how to help.

algesten:

PR #761, opened 2024-06-28
Fix chunked license

algesten:
The reforks the frewsxcv/rust-chunked-transfer crate under the MIT license to apply our changes. The previous fork was under the Apache 2.0 license, which meant more trouble for our own license situation – see linked issue.

Close #727


decathorpe:
I'm not sure this is an improvement? It changes the "license" we would assign to ureq from `(MIT OR Apache-2.0) AND Apache-2.0` to `(MIT OR Apache-2.0) AND MIT`. The `chunked-transfer` crate is covered by the same `MIT OR Apache-2.0` dual-license as ureq itself - wouldn't the simplest solution be to keep the MIT/Apache dual-license?

algesten:
@decathorpe ah sorry. My experience with this is limited. So you say if I bring in both the parent license that solves it?

decathorpe:
In this case, with both projects now being covered by the same dual license, I think that would be the easiest solution, yes.

algesten:
@decathorpe I believe this is ok now. Do you agree?

decathorpe:
Looks good to me!

PR #762, opened 2024-07-03
ureq 3.x

algesten:
This PR/branch is a complete, from the ground up rewrite of ureq.  For Rust there clearly is a role for a sync, minimal dependency HTTP/1.1 client. The ureq 2.x code has gradually been iterated to its current state, but some early design decisions means we are struggling to make some changes we would like to have (such as being transport agnostic). While ureq's surface API is largely okay (with some warts), internally there is a lack of good separation of concerns and clear contracts between the parts of the library.

This PR starts from a complete clean slate, but the intention is to gradually bring back code from the 2.x branch where it makes sense to do so. Especially I'd like to bring back the majority of the tests to ensure backwards compatibility. The intention is also to preserve as much of the API surface as possible, only making changes in areas where there currently are clear problems (such as making unnecessary allocations). The intention is to also have a 1-1 feature parity with 2.x and the MSRV will not move much.

This PR/effort might not work out. I might discover things that are irreconcilable with the current ureq 2.x, or I might run out of steam to complete the project. There is also no timeline for completion, might take months, might take years. Ideally there are more people wanting to engage in the project – please talk to me here or send an email!

Here are some things I been thinking about fixing:

  * Separation between HTTP/1.1 logic and transport – this effort has started in [hoot](https://github.com/algesten/hoot ) – a project which aims to provide a Sans-IO HTTP/1.1 protocol underpinning.
  * Bring your own transport. The transport should be completely plugin based and possible for the user to write their own.
  * [http crate](https://crates.io/crates/http). While I will try to retain a lot of ureq's API, adhering to a common Rust HTTP API is ultimately a better idea.
  * Less allocations – less Strings, Vec. Try to work more with borrows/lifetimes.
  * Potentially webasm. As a consequence of making the transport possible to swap out, we can maybe make ureq work in a webasm context (needs more thought and input!)
  * Improve middleware support. The current middleware chains are OK, but we want them to be even better.
  * Improve the Error type and error-handling.

Please engage in discussion below if there are things I have omitted or ideas of where ureq could go!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,6 +78,7 @@ jobs:
           - socks-proxy
           - gzip
           - brotli
+          - json
     env:
       RUST_BACKTRACE: "1"
       RUSTFLAGS: "-D dead_code -D unused-variables -D unused"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,7 +93,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --no-default-features --features "_test rustls native-roots ${{ matrix.feature }}"
+          args: --no-default-features --features "_test rustls ${{ matrix.feature }}"
 
 
   cargo-deny:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,10 +286,10 @@ checksum = "d231dfb89cfffdbc30e7fc41579ed6066ad03abda9e567ccafae602b97ec5024"
 [[package]]
 name = "hoot"
 version = "0.2.0"
-source = "git+https://github.com/algesten/hoot?rev=cf0e3a0#cf0e3a04dfba6eefe70279324c82a544f7f3ad42"
 dependencies = [
  "http",
  "httparse",
+ "log",
  "smallvec",
  "thiserror",
  "url",
@@ -419,9 +419,9 @@ checksum = "3fdb12b2476b595f9358c5161aa467c2438859caa136dec86c26fdd2efe17b92"
 
 [[package]]
 name = "openssl"
-version = "0.10.64"
+version = "0.10.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95a0481286a310808298130d22dd1fef0fa571e05a8f44ec801801e84b216b1f"
+checksum = "9529f4786b70a3e8c61e11179af17ab6188ad8d0ded78c5529441ed39d4bd9c1"
 dependencies = [
  "bitflags",
  "cfg-if",
@@ -451,9 +451,9 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.102"
+version = "0.9.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c597637d56fbc83893a35eb0dd04b2b8e7a50c91e64e9493e398b5df4fb45fa2"
+checksum = "7f9e8deee91df40a943c71b917e5874b951d32a802526c85721ce3b776c929d6"
 dependencies = [
  "cc",
  "libc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,6 +286,7 @@ checksum = "d231dfb89cfffdbc30e7fc41579ed6066ad03abda9e567ccafae602b97ec5024"
 [[package]]
 name = "hoot"
 version = "0.2.0"
+source = "git+https://github.com/algesten/hoot?rev=f997661#f997661a043a518326b7553740a74a43295eea8c"
 dependencies = [
  "http",
  "httparse",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
+
+[[package]]
 name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -99,10 +105,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "47de7e88bbbd467951ae7f5a6f34f70d1b4d9cfce53d5fd70f74ebe118b3db56"
 
 [[package]]
+name = "cesu8"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
+
+[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
+name = "combine"
+version = "4.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba5a308b75df32fe02788e748662718f03fde005016435c444eea572398219fd"
+dependencies = [
+ "bytes",
+ "memchr",
+]
 
 [[package]]
 name = "cookie"
@@ -357,6 +379,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "49f1f14873335454500d59611f1cf4a4b0f786f9ac11f4312a78e4cf2566695b"
 
 [[package]]
+name = "jni"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c6df18c2e3db7e453d3c6ac5b3e9d5182664d28788126d39b91f2d1e22b017ec"
+dependencies = [
+ "cesu8",
+ "combine",
+ "jni-sys",
+ "log",
+ "thiserror",
+ "walkdir",
+]
+
+[[package]]
+name = "jni-sys"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8eaf4bc02d17cbdd7ff4c7438cafcdf7fb9a4613313ad11b4f8fefe7d3fa0130"
+
+[[package]]
 name = "libc"
 version = "0.2.155"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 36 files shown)


algesten:
For a while I was considering the main API entry point would be taking the request `http::Request<Body>` by reference, i.e. the most central signature would be something like:

```rust
pub fn run(request: &Request<impl IntoBody>) -> Result<Response<RecvBody>, Error> { ... }
```

However `&Request` precludes sending any body would require `&mut`, such as those implementing `std::io::Read` (`File` comes to mind). Therefore I'm changing to not taking by ref. The internals of `hoot` still runs by ref, which means we should be able to avoid ever needing to clone it.

```rust
pub fn run(request: Request<impl IntoBody>) -> Result<Response<RecvBody>, Error> { ... }
```

amkillam:
Just wanted to comment to say that I am very excited about the potential to use ureq with WASM! I managed to make ureq 2.x compile for WASM with some conditional compilation of rustls things, but am currently facing implementing further conditional compilation of TCPStream and some other things to make requests execute successfully. This rewrite was a great idea and will be really useful for my projects.

algesten:
@amkillam I haven't done any WASM myself, so I don't know what do's and dont's there are. I'm not sure I can go full `no_std`, because living with Box<dyn X> could prove a challenge.

I just tried to compile this branch with `cargo build --target wasm32-unknown-unknown` – and that's fine. Does that mean it's ok?

If I add that to the CI, are there more lints/things to check for?

amkillam:
If compilation works then it should be most of the way there. Could you provide an example of how ureq usage would change in 3.x? I can try to make a working example of ureq usage in WASM based on that, and share my results after.

amkillam:
> If compilation works then it should be most of the way there. Could you provide an example of how ureq usage would change in 3.x? I can try to make a working example of ureq usage in WASM based on that, and share my results after.

Also I'm not totally sure about tests specific to WASM yet, but I think once I've (hopefully) created a working example of usage in WASM, I'll be in a better position to comment or maybe contribute if you'd like!

algesten:
> > If compilation works then it should be most of the way there. Could you provide an example of how ureq usage would change in 3.x? I can try to make a working example of ureq usage in WASM based on that, and share my results after.
> 
> Also I'm not totally sure about tests specific to WASM yet, but I think once I've (hopefully) created a working example of usage in WASM, I'll be in a better position to comment or maybe contribute if you'd like!

I will pick you up on that, but it's a bit too early still. I can't even make ureq 3.x do a regular request just yet. Hopefully in a couple of days.

rmg-x:
> and the MSRV will stay the same.

I really really think the MSRV policy should at least be relaxed and communicated clearly with a major version bump. My personal opinion is that users shouldn't be relying on this in the first place but such is life. 

I'm not saying we should bump MSRV anytime a new Rust version is released but allow for some modernization when it makes sense. 
E.g. #725

algesten:
> > and the MSRV will stay the same.
> 
> I really really think the MSRV policy should at least be relaxed and communicated clearly with a major version bump. My personal opinion is that users shouldn't be relying on this in the first place but such is life.
> 
> I'm not saying we should bump MSRV anytime a new Rust version is released but allow for some modernization when it makes sense. E.g. #725

The problem is developer environments installed via Linux (or similar) package managers instead of rustup. They tend to have a much slower upgrade cadence. I want ureq to be available to as many users as possible, and I personally don't think the code get that much worse by not having "modern" constructs like let-else.

You're right about once cell – there we can avoid a dependency. So far I've managed to avoid using it, we'll see how that goes when I get more into the tests. I've been forced to bump MSRV to 1.63 because rustls bumped, so right now I'm leaning towards 1.63.

But nothing is set in stone.

rmg-x:
> installed via Linux (or similar) package managers instead of rustup. They tend to have a much slower upgrade cadence

That's valid. Just as an example, the creator of once_cell supports at least the previous eight Rust versions which is roughly ~1 year behind current. To me, that seems reasonable.

Just as an aside, compatibility for crates and Rust versions is tracked here if you are interested: https://lib.rs/stats#rustc

algesten:
```text
                             │                                            
                                                                          
                             │                                            
                                               ┌──────────┐               
                             │                 │          │               
                                               │ CONNECT  │               
                             │                 │          │               
                                               └──────────┘               
               ┌──────────┐  │  ┌──────────┐   ┌──────────┐               
               │          │     │          │   │          │               
               │   TLS    │  │  │ CONNECT  │   │   TLS    │               
               │          │     │          │   │          │               
               └──────────┘  │  └──────────┘   └──────────┘               
┌──────────┐   ┌──────────┐     ┌──────────┐   ┌──────────┐   ┌──────────┐
│          │   │          │  │  │          │   │          │   │          │
│TcpStream │   │TcpStream │     │TcpStream │   │TcpStream │   │  SOCKS   │
│          │   │          │  │  │          │   │          │   │          │
└──────────┘   └──────────┘     └──────────┘   └──────────┘   └──────────┘
                             │     Proxy          Proxy          Proxy    
    HTTP          HTTPS           CONNECT        CONNECT         SOCKS    
                             │      HTTP          HTTPS                   
                                                                          
                             │                                            
```

Conceptually we have these building blocks that makes up the default _transports_. I'm toying with the idea of being able to make `Transport` composable so you can wrap one transport in another.

algesten:
```text
                         │                                                  
                                                                            
                         │                                                  
                                        ┌─────────┐                         
                         │              │         │                         
                                        │ CONNECT │                         
                         │              │         │                         
                                        └─────────┘                         
            ┌─────────┐  │  ┌─────────┐ ┌─────────┐             ┌─────────┐ 
            │         │     │         │ │         │             │         │ 
            │   TLS   │  │  │ CONNECT │ │   TLS   │             │   TLS   │ 
            │         │     │         │ │         │             │         │ 
            └─────────┘  │  └─────────┘ └─────────┘             └─────────┘ 
┌─────────┐ ┌─────────┐     ┌─────────┐ ┌─────────┐ ┌─────────┐ ┌─────────┐ 
│         │ │         │  │  │         │ │         │ │         │ │         │ 
│TcpStream│ │TcpStream│     │TcpStream│ │TcpStream│ │  SOCKS  │ │  SOCKS  │ 
│         │ │         │  │  │         │ │         │ │         │ │         │ 
└─────────┘ └─────────┘     └─────────┘ └─────────┘ └─────────┘ └─────────┘ 
                         │     Proxy       Proxy       Proxy       Proxy    
    HTTP       HTTPS          CONNECT     CONNECT      SOCKS       SOCKS    
                         │      HTTP       HTTPS        HTTP       HTTPS    
                                                                            
                         │                                                  
```

Actually. I need TLS also for SOCKS.

algesten:
Considering building out a TLS API in ureq3.x to not be in the situation we have in #765

Instead of exposing/re-exporting rustls config (or native-tls), we would abstract over both. The initial abstraction would need support a reasonable amount of minimal config:

* client certs
* custom roots
* sni enable/disable
*  TLS12/13
* disable certificate verifier (maybe under a `dangerous` feature flag)

Even if we default to rustls, native-tls might be a better starting point for looking at which config options to include, since rustls can do more than native-tls.

https://docs.rs/native-tls/0.2.12/native_tls/struct.TlsConnectorBuilder.html
https://docs.rs/rustls/0.23.11/rustls/struct.ConfigBuilder.html

algesten:
@amkillam i think we're at a place where we can try webasm now. 

What I'm not clear on is whether webasm can give me a TCP socket style interface and if TLS should be in-or-outside the scope ureq. Let me explain what we got and see if this is a good starting point for tinkering.

`Agent::new` constructor takes a `config`, a `connector` and a `resolver`. https://github.com/algesten/ureq/blob/feature/ureq3/src/agent.rs#L148-L161 . The config is for stuff like timeouts and TLS. The `connector` is a composable thing that provides a `transport` per request (which I get back to below), and the `resolver` is something that translates hostnames  to `SocketAddr`.

The `DefaultConnector` has a chain of subconnectors that are run one after another. Each can chose to pass-through or build on top of the previous `Transport`. https://github.com/algesten/ureq/blob/feature/ureq3/src/transport/mod.rs#L72-L93
Simplified we get a chain of connectors: `SOCKS -> TcpStream -> Rustls`. If we don't use SOCKS, we get just `TcpStream -> Rustls` and the scheme, `https` decides whether there `Rustls` connector is used.

`Connector` and `Transport` are public traits that we can implement to provide some completely different transport. The `Transport` trait https://github.com/algesten/ureq/blob/feature/ureq3/src/transport/mod.rs#L55-L63 is quite different to `io::Read`/`io::Write` – rather working on shared buffers. This is because I have io_uring in mind for the future.

`TcpTransport` is maybe the cleanest example: https://github.com/algesten/ureq/blob/feature/ureq3/src/transport/tcp.rs#L78

Let's take a concrete example. Let's say webasm transport does not need an explicit resolving, instead you give it some combo (hostname, port, is_tls) – then we can write a dummy Resolver that does nothing and a corresponding Transport opening the connection.

If I remove all default features, it compiles on  wasm32-unknown-unknown. If we need to bring in some feature like rustls, we can maybe make it compole through feature flags – needs investigating.

cpu:
>  For now I decided to not use the webpki-roots crate instead opting for using rustls-native-certs for both rustls and native-tls. webpki-roots is structured such that it's only compatible with rustls, thus it breaks symmetry to use it.

For unrelated reasons we've been thinking about extending `webpki-roots` with an optional feature that includes the DER of the self-signed root certificates in addition to the more minimal (and [`webpki`](https://github.com/rustls/webpki) crate focused) trust anchor representation. I think if that were to happen you could use `webpki-roots` for both TLS backends similar to `rustls-native-certs`. The main downside (and why it would be opt-in) is the increase in binary size from all the extra (and for rustls, superfluous) data.

If you were open to breaking symmetry between backends the [rustls-platform-verifier](https://github.com/rustls/rustls-native-certs) crate is often the best choice. The README table breaks it down, but it works like a better `rustls-native-certs` on most platforms and degrades to `rustls-native-certs` and `webpki-roots` as appropriate for other platforms (like Linux/BSD/WASM).

> What I'm not clear on is whether webasm can give me a TCP socket style interface and if TLS should be in-or-outside the scope ureq.

I'm very out of the loop with wasm, but have heard folks say they've gotten rustls working there. Certainly you need `webpki-roots` in this context because there is no platform trust store to load certificates from in that environment.

In case it helps colour your decision for support `rustls` is _interested_ in adding first-class wasm support but it hasn't been done yet and isn't something tested in CI. I've heard anecdotally it works, but it could break :-)

algesten:
> For unrelated reasons we've been thinking about extending `webpki-roots` with an optional feature that includes the DER of the self-signed root certificates in addition to the more minimal (and [`webpki`](https://github.com/rustls/webpki) crate focused) trust anchor representation. I think if that were to happen you could use `webpki-roots` for both TLS backends similar to `rustls-native-certs`. The main downside (and why it would be opt-in) is the increase in binary size from all the extra (and for rustls, superfluous) data.

That would be a most welcome feature indeed. In a previous life I hugely preferred `webpki-roots` to the platform, since enterprise sec and others want to MITM by installing their own roots. That's why ureq2.x used the `webpki-roots` as primary. However I'm older and wiser now, and maybe it's more pragmatic to just go with the local root store as the default choice.

> If you were open to breaking symmetry between backends the [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier) crate is often the best choice for `rustls` in particular. The README table breaks it down, but it works like a better `rustls-native-certs` on most platforms and degrades to `rustls-native-certs` and `webpki-roots` as appropriate for other platforms (like Linux/BSD/WASM).

Looks great! Revocations support is nice, and using the OS-default verifiers probably means even better protection in case there are 0-day updates etc. I think it is worth breaking symmetry for, yes.

> I'm very out of the loop with wasm, but have heard folks say they've gotten rustls working there. Certainly you need `webpki-roots` in this context because there is no platform trust store to load certificates from in that environment (e.g. `rustls-native-certs` would never be supported in WASM AIUI).
> 
> In case it helps colour your decision for support `rustls` is _interested_ in adding first-class wasm support but it hasn't been done yet and isn't something tested in CI. I've heard anecdotally it works, but it could break :-)

Interesting. I was hoping that by making the transport truly plug-in in ureq3.x this would open for these possibilities. But I'm not sure I'll tinker with it much myself until I have a real need.

algesten:
With that last push, ureq 3.x is now feature complete in terms of feature flags.

There are a couple of TODOs I want to deal with and I need to glance through the 2.x code to make sure there aren't functionality areas that I've forgotten about.

Next up is porting all the tests over.

cpu:
>> For unrelated reasons we've been thinking about extending webpki-roots with an optional feature that includes the DER of the self-signed root certificates in addition to the more minimal (and [webpki](https://github.com/rustls/webpki) crate focused) trust anchor representation.
>
> That would be a most welcome feature indeed.

In case it ends up being helpful I wanted to mention I finished implementing this. It ended up being a separate crate instead of an opt-in feature: https://crates.io/crates/webpki-root-certs

algesten:
> In case it ends up being helpful I wanted to mention I finished implementing this. It ended up being a separate crate instead of an opt-in feature: https://crates.io/crates/webpki-root-certs

Oh nice! I'll make a feature flag for that.

algesten:
Moving this to the main branch with individual PRs now.

Issue #765, opened 2024-07-08
rustls upgrade in 2.10 is a breaking change?


jerrinot:
Hello,

I believe ureq 2.10 introduces a breaking change that should be reflected in the library's versioning.
Why do I think this is a breaking change?
- ureq exposes Rustls ClientConfig in its API, making Rustls's public API a part of ureq's public API.
- ureq 2.10 updated Rustls from 0.22 to 0.23.
- According to cargo semver, the update from Rustls 0.22 to 0.23 is interpreted as a breaking change.

Example of the breakage:
- We maintain a library that depends on both ureq and rustls.
- Our library uses Rustls both directly and through ureq.
- In our Cargo.toml, we specify ureq = "2.9" and rustls = "0.22".
- Our library is published on crates.io.

After the release of ureq 2.10, our library users automatically receive ureq 2.10 as it is considered compatible with 2.9. As a result, our library users end up with two different versions of rustls:
- Version 0.23 as a transitive dependency of ureq,
- Version 0.22 as declared in our own Cargo.toml.

This leads to a scenario where the library instantiates ClientConfig from rustls 0.22 and attempts to pass it to AgentBuilder.tls_config(), which expects a client config from rustls 0.23, resulting in compatibility issues.
```2298 |                     Some(tls_config) => agent_builder.tls_config(tls_config),
     |                                                       ---------- ^^^^^^^^^^ expected `rustls::client::client_conn::ClientConfig`, found `ClientConfig`
     |                                                       |
     |                                                       arguments to this method are incorrect
     |
     = note: `ClientConfig` and `rustls::client::client_conn::ClientConfig` have similar names, but are actually distinct types
```

I am not very well versed in Rust ecosystem, but I believe that if you expose a 3rd party library in your API and this library declares a breaking change then your library should declare it too.

algesten:
Yeah. I don't know what to do there. Re-exporting rustls like this has proven to be a bad idea. I guess re-exporting anything which is semver 0.x.x is not good, since 0.x means patch versions are allowed to be breaking.

I really don't want to bump major version ureq for this, since I believe many users will not be affected in the same way as yourself.

I can only apologise and promise to do better in ureq 3.x, which I'm working actively on right now (https://github.com/algesten/ureq/pull/762). I will add to the list for 3.x to be super conservative regarding re-exporting dependencies.

I am sorry!

jerrinot:
hello @algesten,

thanks for a super quick reply. No need to apologize - you owe me nothing!
[...]

PR #767, opened 2024-07-16
ureq 2.10 tweaks for Rustls 0.23 update

cpu:
:wave: here are a few changes pulled out of our discussion in https://github.com/algesten/ureq/issues/765

### docs: add more Rustls detail to 2.10.0 CHANGELOG

Two main goals here:
1. Explicitly call out that the Rustls update from 0.22 to 0.23 occurred, and is a breaking change for consumers that were using their own Rustls 0.22 dep to construct configs given to `ureq`. The Rustls 0.23 changelog is linked to discuss the upstream changes in detail. This explicitly addresses [the original issue](https://github.com/algesten/ureq/issues/765#issue-2395808883) flagged in #765.
2. Expand on the implications of `ureq` activating the `ring` feature by default. In particular you must either ensure Rustls features are unambiguous or install a process-wide default provider before building configs, or having `ureq` build you a config. This covers the reports in #765 of [an unexpected panic](https://github.com/algesten/ureq/issues/765#issuecomment-2217950386). This becomes slightly less painful with the subsequent change to update `default_tls_config()`'s builder choice.

### rtls: default_tls_config() explicit `*ring*` choice

This change is [informed by my comment](https://github.com/algesten/ureq/issues/765#issuecomment-2223080932) describing a place for improvement with `ureq`'s preference to use `*ring*` by default. We can avoid a risk of a panic building the `ureq` default TLS config and make the default provider intention clearer by using [`rustls::ClientConfig::builder_with_provider()`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#method.builder_with_provider) instead of [`rustls::ClientConfig::builder()`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#method.builder) (which requires a clear process wide default is available, or panics).

###  Cargo: bytes v1.6.0 -> v1.6.1

<details>
<summary>Small add-on that fixes a `cargo-deny` error:</summary>

```
error[yanked]: detected yanked crate (try `cargo update -p bytes`)
   ┌─ /github/workspace/Cargo.lock:10:1
   │
10 │ bytes 1.6.0 registry+https://github.com/rust-lang/crates.io-index
   │ ----------------------------------------------------------------- yanked version
   │
   = bytes v1.6.0
     └── http v0.2.12
         └── ureq v2.10.0

advisories FAILED, bans ok, licenses ok, sources ok
```
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,22 @@
   * Bump MSRV 1.61 -> 1.63 due to rustls (#764)
   * Update deps (only patch versions in Cargo.lock) (#763)
   * Refork frewsxcv/rust-chunked-transfer to fix MIT/Apache2.0 license (#761)
-  * Fix doc Rustls does now support IP address certificates (#759)
   * Enable http-crate feature for docs (#755)
-  * Rustls dep to default to ring backend (#753)
+  * Update Rustls from 0.22 to 0.23 - this may be a breaking change if your
+    application depends on Rustls 0.22 (e.g. to provide a custom 
+    `rustls::ClientConfig` to `ureq`). See the [Rustls 0.23.0][rustls-0.23.0]
+    changelog for a list of breaking API changes (#753)
+  * Rustls dep to default to ring backend. If your project uses the
+    default `ureq` TLS config, or constructs its own `rustls::ClientConfig`
+    with `rustls::ClientConfig::builder()` you must ensure the Rustls 
+    `aws-lc-rs` feature is not activated, or set the process default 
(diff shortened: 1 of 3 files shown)


algesten (approved):

algesten:
@cpu thank you!

I will merge this. I'm inclined to make another patch version with the change quickly. I think defaulting to ring isn't going to break anything, and it might save some headache for people not having got around to upgrade just yet.

cpu:
Happy to help :-)

PR #773, opened 2024-08-12
[ureq 3.x] Use process-wide default rustls backend if set

laniakea64:
This PR is the idea in https://github.com/algesten/ureq/issues/765#issuecomment-2282921492 : Keep `ring` as default and simplify opting into other `rustls` backend.

Also renames `RustlsWithRing` to `Rustls` since user can easily opt into an alternative backend.


algesten:
That's neat. I like it!

algesten:
Thanks!

PR #789, opened 2024-08-20
Fix panic in state Cleanup

### LINKED ISSUE #787: ureq 3.x Panic Incorrect state: Cleanup
Running the 1000000 top sites test, we sometimes, very rarely get this panic:

```
[2024-08-15T00:32:00Z ERROR ureq::unit] Incorrect state: Cleanup
thread 'main' panicked at src/unit.rs:270:22:
```

The problem is to do with sending `Input::Begin` when unit is in the wrong state. The `extract!` below is what's failing.

```rust
            Input::Begin => {
                let flow = extract!(&mut self.state, State::Begin)
                    .expect("Input::Begin requires State::Begin");

                self.call_timings.time_call_start = Some(now);
                self.set_state(State::Prepare(flow));
            }
```

algesten:
Close #787

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -9,7 +9,7 @@ use http::{HeaderName, HeaderValue, Method, Request, Response, Uri};
 use crate::body::{Body, ResponseInfo};
 use crate::middleware::MiddlewareNext;
 use crate::pool::{Connection, ConnectionPool};
-use crate::resolver::{DefaultResolver, Resolver};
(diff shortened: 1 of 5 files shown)


algesten:
Testing this on  a test host now for 24h.

algesten:
Lol. I've been pulling my hair out trying to understand why the changes I made in this PR didn't fix the panics... only to realise today that I hadn't been testing with the latest code from the PR. 🤦‍♂️

PR #790, opened 2024-08-21
replace once_cell to std impl

stackinspector:
I can't guarantee full equivalence of behavior, but at least it's conceptually equivalent. (The `once_cell` crate should write a migration guide!)
This will bump MSRV to `1.80`. `once_cell` only needs `1.70`, while `lazy_cell` needs `1.80`. I originally thought only the former was needed, but then I saw the latter. The latter is a bit of a span, so it depends on you if you can accept it.
Bumped Rust edition to 2021 & removed redundant `use`s by the way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1004,7 +1004,6 @@ dependencies = [
  "http",
  "log",
  "native-tls",
- "once_cell",
  "rustls",
  "rustls-pemfile",
  "rustls-pki-types",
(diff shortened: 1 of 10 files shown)


algesten:
Hi @stackinspector, thanks for this!

ureq will always be very conservative in MSRV. 

We are not going to bump MSRV at this point. If you look at the numbers at the bottom (Usage of Rust versions based on crates.io requests): https://lib.rs/stats#rustc-usage you see that 1.67 is where we hit 99% of all users. If we were to bump to 1.80 that number drops to 49%.

Also a duplicate of #725

PR #792, opened 2024-08-23
public TransportAdapter::set_timeout()

algesten:


algesten:
Fix problem mentioned in #792

PR #794, opened 2024-08-26
More TCP config using net2

### LINKED ISSUE #547: Interest in a PR that adds TCP keep-alive?
I'm currently working through what to do about a codebase that uses `ureq` and is suffering from what looks like a very occasionally unresponsive server. I'd like to toggle on TCP keepalive to see if that fixes the problem, but currently there is no API for this. `reqwest` has this: https://docs.rs/reqwest/0.11.12/reqwest/blocking/struct.ClientBuilder.html#method.tcp_keepalive and the implementation which is in the middle of being migrated to `hyper-util` looks pretty simple: https://github.com/hyperium/hyper-util/blob/85aade41a080fc4b40a9a53c818680feed14f138/src/client/connect/http.rs#L593-L598

Are you interested in accepting a PR which adds this feature?

algesten:
Close #547

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,11 +74,13 @@ jobs:
         feature:
           - ""
           - charset
+          - native-tls
           - cookies
           - socks-proxy
           - gzip
           - brotli
           - json
+          - net2
     env:
       RUST_BACKTRACE: "1"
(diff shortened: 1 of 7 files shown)


algesten:
Dammit. net2 is very deprecated.

PR #796, opened 2024-08-31
Rewrite the run() loop

algesten:


algesten:
500 rows of code that was a total waste of space 🤪

PR #813, opened 2024-09-17
Mitigate version mismatch of public rustls dependency

kornelski:
Technically [2.10.0, 2.9.2, etc.](https://lib.rs/crates/ureq/versions#version-2.10.0) were breaking releases, because they upgraded rustls, which has its type exposed in the public API. This [has broken `maturin`](https://github.com/PyO3/maturin/issues/1671).

Instead of requiring users to add a matching version of `rustls` dependency themselves, an easier and more robust option is to re-export the version that `ureq` requires.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -354,6 +354,18 @@
 //! [actix-web](https://crates.io/crates/actix-web), and [hyper](https://crates.io/crates/hyper).
 //!
 
+/// Re-exported rustls crate
+///
+/// Use this re-export to always get a compatible version of `ClientConfig`.
+#[cfg(feature = "rustls")]
+pub use rustls;
+
+/// Re-exported native-tls crate
+///
+/// Use this re-export to always get a compatible version of `TlsConnector`.
+#[cfg(feature = "native-tls")]
+pub use native_tls;
+
 mod agent;
 mod body;
 mod chunked;


algesten:
Hi @kornelski, welcome to ureq!

In case you didn't see it, [this is the big thread on 2.10 being a breaking change](https://github.com/algesten/ureq/issues/765). My last comment summarizes [my thoughts](https://github.com/algesten/ureq/issues/765#issuecomment-2284176962).

All this turmoil actually was the final straw to make me rewrite ureq from the ground up to 3.x. Regarding re-exporting and having other crates in ureq's public API. This changes in 3.x. We will not re-export anything which isn't semver 1.x. That means ureq3 creates its own TLS config and cookie config.

The PR you got here is against the main branch, which is now ureq3. These re-exports are not necessary in this branch due to above reason.

You could rebase the PR against [2.x](https://github.com/algesten/ureq/tree/2.x), but I'm not sure it's worth merging it since ureq3.x is around the corner.

What do you think?

kornelski:
I think it's worth merging, because the crate ecosystem can take a long time to upgrade major versions (there's a chicken-egg problem when projects don't want two versions of `ureq`, and probably two or more `rustls`s).

algesten:
Ok. Let's merge

PR #825, opened 2024-09-28
`tls-aws-lc-rs` feature to enable rustls `aws-lc-rs` crypto provider

kolbma:
Provides `tls-aws-lc-rs` feature to be able to use the default __rustls__ `aws-lc-rs` crypto provider.

Also fixes the not available feature `rustls` in `lib.rs` to re-export `rustls` like in _main_ branch.
And removes usage of unavailable `tls-native` feature in https-agent tests.


kolbma:
If you are interested to merge this to __2.x__ I'll also provide the pull request for the __main branch__.

algesten:
I'm sorry, but I don't want further work in the 2.x branch. All my attention is now at 3.x

PR #844, opened 2024-10-08
rustls-tls12 feature

### LINKED ISSUE #842: Provide a way to disable `tls12` feature for rustls
ureq is a lightweight HTTP client often used to communicate with one or a few hardcoded servers (e.g. making API calls). TLS 1.2 support is mostly unnecessary in those situations because we can know for sure if the hardcoded servers support TLS 1.3[^1] or not. 

TLS 1.2 in rustls often requires separate code paths to handle. Disabling the `tls12` feature can reduce binary size by ~100kb. It'd be great if ureq can provide a way to do so.

[^1]: Which is also highly likely, as 70% of sites now support TLS 1.3 according to https://www.ssllabs.com/ssl-pulse/

algesten:
Close #842

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,12 @@ exclude = ["/cargo_deny.sh", "/deny.toml", "/test.sh"]
 rust-version = "1.67"
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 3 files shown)


algesten:
This complicates the feature flag situation for a rather small gain. I don't want this, for now.

PR #850, opened 2024-10-12
Remove duplicated header names

naglis:
`date` and `connection` are currently duplicated.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -190,8 +190,6 @@ const NON_SENSITIVE_HEADERS: &[&str] = &[
     "accept",
     "accept-encoding",
     "accept-charset",
-    "date",
-    "connection",
     "server",
     "user-agent",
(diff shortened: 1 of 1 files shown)


naglis:
On a somewhat related note, I was wondering if using [`HeaderName`](https://docs.rs/http/1.1.0/http/header/struct.HeaderName.html) for `NON_SENSITIVE_HEADERS` would make sense?

algesten:
@naglis hi! welcome to ureq!

You're right. This is duplicated. Thanks for the PR!

> On a somewhat related note, I was wondering if using [`HeaderName`](https://docs.rs/http/1.1.0/http/header/struct.HeaderName.html) for `NON_SENSITIVE_HEADERS` would make sense?

Yeah, the const fn looks good. Let's do that! PR welcome!

PR #866, opened 2024-10-27
Add `vendored` cargo feature

soywod:
This PR forwards the `vendored` cargo features to `native-tls`. Useful when users want to statically link OpenSSL.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -529,6 +529,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
+name = "openssl-src"
+version = "300.4.0+3.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a709e02f2b4aca747929cca5ed248880847c650233cf8b8cdc48f40aaf4898a6"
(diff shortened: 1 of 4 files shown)


algesten:
Hi @soywod welcome to ureq!

This looks good. Notice that the README is not the authorative source. We use `cargo readme` to generate it. So the doc change needs to be made in `lib.rs`.

If you can fix that, we can merge.

Thanks!

soywod:
> Notice that the README is not the authorative source. We use `cargo readme` to generate it. So the doc change needs to be made in `lib.rs`.

Oh I see, my bad. Could be nice to mention it somewhere like in the `README.md` itself or in the `CONTRIBUTING.md`.

I just pushed the fix.

algesten:
> Oh I see, my bad. Could be nice to mention it somewhere like in the `README.md` itself or in the `CONTRIBUTING.md`.

This is the first line of README.md:

```
[comment]: # (README.md is autogenerated from src/lib.rs by `cargo readme > README.md`)
```

:)

Easy to miss though.

algesten:
Thanks!

soywod:
> This is the first line of README.md.
> Easy to miss though.

Ha. Indeed I missed it also. I thought to explicitly write it down in the doc, not as a comment (or in the contributing section). Anyway just a detail.

I will wait for the next release. Thanks!

Issue #869, opened 2024-10-29
Ureq 3.x.x TlsConfigBulder::disable_verification(true) should not set supported signature schemes to an empty array


sosthene-nitrokey:
https://github.com/algesten/ureq/blob/de4f0fb8bf5b666d20e420b7ae35a6f31c800544/src/tls/rustls.rs#L243

Since the disabled verifier accepts any certificate and any signature, it should instead return an array of all defined signature algorithms. Otherwise the other end will reject the handshake.

algesten:
Oops! Yep. Must be done!

PR #871, opened 2024-10-31
disable_verification: Signal that all signature schemes are supported

sosthene-nitrokey:
Without this, the server will not accept the connection because it can't use any signature scheme

This adds a test (against google.com) that fails without this change.

Close https://github.com/algesten/ureq/issues/869

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -634,6 +634,35 @@ pub(crate) mod test {
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


algesten:
I think this needs rebasing off current main?

sosthene-nitrokey:
Done

algesten:
Thanks!

PR #872, opened 2024-11-06
Rename hoot -> ureq-proto

algesten:


algesten:
@sosthene-nitrokey hey, so here's the PR that switches over from `hoot` to `ureq-proto`. I also decided to re-export the `http` crate that is exported from `ureq-proto` to make sure these versions always line up.

Next step would be to start moving the `Transport` traits etc to the proto crate.

Does this look OK to you?

sosthene-nitrokey:
Yes, this looks good to me!

PR #878, opened 2024-11-14
Make MSRV 1.63 work

algesten:
- **Update openssl**
- **Mitigate version mismatch of public rustls dependency**
- **Fix incorrect feature flag**
- **Make MSRV 1.63 work**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,9 +370,9 @@ checksum = "3fdb12b2476b595f9358c5161aa467c2438859caa136dec86c26fdd2efe17b92"
 
 [[package]]
 name = "openssl"
-version = "0.10.64"
+version = "0.10.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95a0481286a310808298130d22dd1fef0fa571e05a8f44ec801801e84b216b1f"
+checksum = "9529f4786b70a3e8c61e11179af17ab6188ad8d0ded78c5529441ed39d4bd9c1"
 dependencies = [
  "bitflags",
  "cfg-if",
@@ -402,9 +402,9 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.102"
+version = "0.9.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c597637d56fbc83893a35eb0dd04b2b8e7a50c91e64e9493e398b5df4fb45fa2"
+checksum = "7f9e8deee91df40a943c71b917e5874b951d32a802526c85721ce3b776c929d6"
 dependencies = [
  "cc",
  "libc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,10 @@
 version = 3
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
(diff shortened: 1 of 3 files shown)


algesten:
I found a combination of versions that work to compile 1.63 locally. However the CI now fails with:

```
error[E0282]: type annotations needed for `Box<_>`
Error:   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.20/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
   = note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`
```

algesten:
I think that means we might be stuck in a situation where ureq 2.x either must bump MSRV to whatever time requires at 0.3.35, or not be able to compile with rustc >= 1.80

epage:
FYI the Cargo team [discourages using non `^` version requirements except in some very specific cases](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-version-requirements).  Enforcing of MSRV is not one of those and is in fact the motivating reason for us writing that documentation.

There are two problems with enforcing MSRV this way
- This makes ureq incompatible with other packages that might require newer versions of these packages.
- Not all users of ureq have the same MSRV and may be able to use newer versions of these packages *and* want to use them

From Cargo's [documentation for `rust-version`](https://doc.rust-lang.org/nightly/cargo/reference/rust-version.html#support-expectations)

> In support of the above, it is expected that each dependency’s version-requirement supports at least one version compatible with your rust-version. However, it is not expected that the dependency specification excludes versions incompatible with your rust-version. In fact, supporting both allows you to balance the needs of users that support older Rust versions with those that don’t.

The way I see people handle this include
- Committing their lockfile and tracking MSRV-compatible versions with it ([yes, its a pain](https://rust-lang.github.io/rfcs/3537-msrv-resolver.html#status-quo))
- Using `-Zminimal-versions` to approximate an MSRV-aware resolver

In 1.84, Cargo will include an MSRV-aware resolver, see https://doc.rust-lang.org/nightly/cargo/reference/config.html#resolverincompatible-rust-versions

Use of the MSRV-aware resolver does not require bumping your MSRV; you just need to change your `Cargo.lock` with 1.84+ and have it enabled.  You can even have it enabled in a `.cargo/config.toml` and Cargo will ignore it in older versions.

kamulos:
There is also another angle to this matter: I think in the case of `rustls`, pinning is especially bad. If a CVE occurs in rustls people will be unable to update without an update to `ureq`. In the worst case this is really time sensitive.

epage:
> The MSRV table you [provided here](https://rust-lang.github.io/rfcs/3537-msrv-resolver.html#status-quo) tells us that if I unlock the rustls dep, a user on anything below <1.71 will fail to compile with less than editing the lock file. That means ~25% of potential users of ureq could fail to use ureq out-of-the box. Did I get that right?

That table was published on 11/10/2023.  Whats more important is to look at the relative value.  1.71 was released on 2023-07-13 which is almost 1.5 years ago.  For a release that was a year old at the time the table was made, 98.766% of requests to crates.io would be satisfied.

> One of ureq's primary goals have always been to work out-of-the-box especially for beginners to Rust.

In considering the out-of-the-box experience for beginniners, also consider the esoteric errors you put your users at risk of seeing, e.g. https://www.reddit.com/r/rust/comments/p8clcx/how_to_fix_cargo_dependency_issue/

algesten:
@epage I have now released 2.12.0. This is MSRV 1.71 with dependencies unpinned.

I am going to maintain two lines:

* 2.11.x -> MSRV 1.67
* 2.12.x -> MSRV 1.71

I will pin dependencies in those lines as needed by releasing patch versions. If I must bump MSRV again, it will be in another minor version.

2.x will not get any more features. So I will only release patch to keep it alive (upon request).

epage:
@algesten is there a reason you are still pinning versions in your `Cargo.toml` file?

Issue #883, opened 2024-11-25
Ureq 3.x: keep the ability to get response uri


Rileran:
In ureq 2.x, the `get_url` method on the `Response` struct allowed to get the final uri used in case of HTTP redirect.

With the use of the `http` crate, this feature is gone because the `Response` object does not keep the uri. I think it may be possible to use the `Extension` feature of the `http` crate to bring back this feature. Would you be interested in a PR?

For reference, this mechanism is currently used by `reqwest` to provide this feature: https://github.com/seanmonstar/reqwest/blob/e2d4b14c695acaebe531af7ce4712e182278f38a/src/response.rs

algesten:
Closed by #884.

Thanks!

PR #884, opened 2024-11-25
Allow accessing last redirected Uri from Response<Body>

Rileran:
Not sure where to put the documentation associated.

Related to https://github.com/algesten/ureq/issues/883

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -1,0 +1,25 @@
+use http::Uri;
+
+use crate::body::Body;
+use crate::http;
+
+#[derive(Debug, Clone)]
+pub(crate) struct ResponseUri(pub http::Uri);
+
+/// Extension trait for `http::Response<Body>` objects
+///
+/// Allows the user to access the `Uri` in http::Response
+pub trait ResponseExt {
+    /// The Uri we ended up at. This can differ from the request uri when we have followed redirects.
+    fn get_uri(&self) -> &Uri;
+}
+
+impl ResponseExt for http::Response<Body> {
+    fn get_uri(&self) -> &Uri {
+        &self
+            .extensions()
+            .get::<ResponseUri>()
+            .expect("uri to have been set")
+            .0
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,6 +431,7 @@ use http::{Request, Response, Uri};
 pub use proxy::Proxy;
 pub use request::RequestBuilder;
 use request::{WithBody, WithoutBody};
+pub use response::ResponseExt;
 pub use send_body::AsSendBody;
 
 mod agent;
@@ -441,6 +442,7 @@ mod pool;
 mod proxy;
 mod query;
 mod request;
+mod response;
 mod run;
 mod send_body;
 mod timings;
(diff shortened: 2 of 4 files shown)


algesten:
Hi @Rileran, and welcome to ureq!

Thank you for this. I wonder whether it's better to expose this as an extension trait for Response, thus getting it documented?

```rs
pub trait ResponseExt {
    fn get_uri(&self) -> Uri;
}

impl ResponseExt for Response<Body> {
...
}
```

src/response.rs
```diff
@@ -0,0 +1,20 @@
+use http::{Response, Uri};
+
+use crate::body::Body;
+use crate::http;
+
+/// Extension trait for http::response::Builder objects
```
algesten, line 6: The trait is for `http::Response<Body>`

src/response.rs
```diff
@@ -0,0 +1,20 @@
...
+///
+/// Allows the user to access the `Uri` in http::Response
+pub trait ResponseExt {
+    /// The Uri we ended up at. This can differ from the request uri when we have followed redirects.
+    fn get_uri(&self) -> &Uri;
+}
+
+impl ResponseExt for Response<Body> {
```
algesten, line 14: Internally we always prefix some `http` crate types to make sure there is no confusion. This line should read:

`impl ResponseExt for http::Response<Body> {`

src/response.rs
```diff
@@ -0,0 +1,20 @@
...
+    fn get_uri(&self) -> &Uri {
+        self.extensions()
+            .get::<Uri>()
```
algesten, line 17: The final question is whether to use a non-public newtype to avoid this becoming a secondary API. You can see another case for that in the same trait, where we pass `Config` in the extensions, but we don't want that object to be a public API via the extensions. So we make this `RequestLevelConfig` wrapper which is `pub(crate)` to lock it down.

https://github.com/algesten/ureq/blob/main/src/config.rs#L563

algesten (changes requested):
Looks good! Some niggly things.

algesten:
Hey @Rileran can you rebase off latest main and I think the build should go through cleanly.

Rileran:
Should be good

algesten:
Now you are only seeing the errors that are to do with this PR.

The main failing problem is that when we run the test suite on CI it runs with `cargo test --all-features` enables a feature `_test` that ensures we are not hitting real servers in the test suite. `cargo test --features _test`

Unless you find another test that is already set up correctly, you'll need to tweak `TestConnector` to handle your case.

src/transport/test.rs
```diff
@@ -290,13 +290,22 @@ fn setup_default_handlers(handlers: &mut Vec<TestHandler>) {
...
+            )?;
+            write!(
+                w,
+                "HTTP/1.1 200 OK\r\n\
+                Content-Type: application/json\r\n\
+                Content-Length: {}\r\n\
+                \r\n",
+                HTTPBIN_GET.as_bytes().len()
+            )?;
+            w.write_all(HTTPBIN_GET.as_bytes())
```
algesten, line 308: I don't think this will really do anything since ureq will make a new request to `/get` as per the `Location` header of the redirect.

algesten: I can't imagine it working. There will be a new mpsc channel for the next request, so writing the second response here should just be discarded.

algesten (approved):

algesten:
Thanks!

Issue #667, opened 2023-09-30
WebAssembly support


brandonros:
I don't know why I expected this to work with WebAssembly (it compiles for the `wasm32-unknown-unknown` target if you disable any TLS/SSL features).

But you will get to JavaScript layer and have one of what I guess is many issues (how would it know to do HTTP over `fetch()`?):

```
panicked at 'time not implemented on this platform', library/std/src/sys/wasm/../unsupported/time.rs:13:9

Stack:

Error
    at http://localhost:8080/pkg/graphql_poc.js:188:21
```

Which comes from

```
 let agent = AgentBuilder::new()
        .timeout_read(Duration::from_secs(30))
        .timeout_write(Duration::from_secs(30))
        .build();
```

I read you can use Chrono::Duration (which does support WASM) but AgentBuilder would need to accept non std::time::Duration?

algesten:
Hi @brandonros! Welcome to ureq!

I have never used WebAssembly myself, so I don't actually know what would be required to make that work. The `Duration` here are ultimately used as socket timeouts. Those sockets are `std::net::TcpStream`, and I'll imagine that even if we get past the `Duration` problems, we will get to a dead end for the sockets. I don't know if it would be possible to solve that – blocking sockets at the bottom of the call stack?

A possible direction for a ureq 3.0, would be to isolate the HTTP request/response function into a Sans IO pattern backing library without allocations, deliberately target no_std and webassembly. Then make nice higher level wrappers for std, no_std and webasm that sits on the same core.

jsha:
I'm curious @brandonros, what's the use case for using ureq in WebAssembly?

If it's useful, we could probably do something where all the meaningful features are stubbed out to `fetch`, but we'd need some motivating use cases.

I do really like the idea of making a no_std version of ureq (with more user-friendly features stacked on top when std is present).

algesten:
@jsha i've been thinking about the no_std thing.

Here's an experiment just fleshing out what a no_std low level API could look like: https://github.com/algesten/h1-call

Design goals:

* `no_std` and no alloc crate. Stack is enough
* Sans-IO. Bring your own transport.
* Allow flushing the buffer to the transport at any point. I.e. do not require all headers to be in the borrowed buffer at the same time.
* Encode protocol as types, i.e. let `State` mean "the state the call is currently in", `Version` means Http/1.0 or 1.1. `Method` means GET/HEAD/etc...
[...]

PR #893, opened 2024-11-26
use web_time

brandonros:
part of https://github.com/algesten/ureq/issues/667

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,6 +142,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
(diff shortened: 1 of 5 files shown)


algesten:
To keep the dependency count low, I'd like to introduce a `std` feature flag that is on by default.

This dependency should be pulled in when `std` is not present.

algesten:
I see you already got that part. I managed to run your wasm test, thanks!

I'll make a branch for the no_std work so you can rebase this PR against that.

brandonros:
moving to https://github.com/algesten/ureq/pull/900

PR #896, opened 2024-11-29
improve ergonomics around AutoHeaderValue

asibahi:
upon upgrading to `rc3` call to set a custom user agent suddently became much longer. 

This is a small suggested code change to improve the ergonomics hopefully without mucking up the design.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -452,8 +452,8 @@ impl<Scope: private::ConfigScope> ConfigBuilder<Scope> {
     /// Setting a value of `""` on the request or agent level will also not send a header.
     ///
     /// Defaults to `Default`, which results in `ureq/<version>`
-    pub fn user_agent(mut self, v: AutoHeaderValue) -> Self {
-        self.config().user_agent = v;
+    pub fn user_agent(mut self, v: impl Into<AutoHeaderValue>) -> Self {
+        self.config().user_agent = v.into();
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -688,6 +688,14 @@ impl AutoHeaderValue {
...
 }
 
+impl From<Arc<String>> for AutoHeaderValue {
+    fn from(value: Arc<String>) -> Self {
+        match value.as_str() {
+            "" => Self::None,
+            _ => Self::Provided(value)
+        }
+    }
+}
```
algesten, line 698: Would this conflict if we wanted to do a `impl<S: AsRef<str>> From<S> for AutoHeaderValue { ... }`?

asibahi: All three impls would, apparently. Would you rather the implementation be that instead of them? I figured the separate impl for `Arc<String>` for people who already have the `Arc` hanging around somewhere.

```suggestion
// impl From<Arc<String>> for AutoHeaderValue {
//     fn from(value: Arc<String>) -> Self {
//         match value.as_str() {
//             "" => Self::None,
//             _ => Self::Provided(value)
//         }
//     }
// }
// impl From<String> for AutoHeaderValue {
//     fn from(value: String) -> Self {
//         match value.as_str() {
//             "" => Self::None,
//             _ => Self::Provided(Arc::new(value))
//         }
//     }
// }
// impl From<&str> for AutoHeaderValue {
//     fn from(value: &str) -> Self {
//         match value {
//             "" => Self::None,
//             _ => Self::Provided(Arc::new(value.to_owned()))
//         }
//     }
// }
impl<S : AsRef<str>> From<S> for AutoHeaderValue {
    fn from(value: S) -> Self {
        match value.as_ref() {
            "" => Self::None,
            _ => Self::Provided(Arc::new(value.as_ref().to_owned()))
        }
    }
}
```

algesten: Yeah. I prefer this one. If you change the PR we merge.

asibahi: Done

algesten:
Thanks!

PR #905, opened 2024-12-03
Prepare 2.12.0 release

algesten:
2.12.0 is unpinning the rustls dep and bumps MSRV to 1.71.

The idea is that 2.12.x will be MSRV 1.71 and 2.11.x will be MSRV 1.67.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ureq"
-version = "2.11.0"
+version = "2.12.0"
 authors = ["Martin Algesten <martin@algesten.se>", "Jacob Hoffman-Andrews <ureq@hoffman-andrews.com>"]
 description = "Simple, safe HTTP client"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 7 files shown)


algesten:
Related to #878

Cargo.toml
```diff
@@ -58,12 +58,7 @@ brotli-decompressor = { version = "4.0.0", optional = true }
 http-02 = { package = "http", version = "0.2", optional = true }
 http = { version = "1.1", optional = true }
 url = "2.5.0"
-
-# These are locked down for MSRV 1.67
-rustls = { version = "=0.23.19", optional = true, default-features = false, features = ["ring", "logging", "std", "tls12"] }
-yoke = "=0.7.4"
-litemap = "=0.7.3"
-zerofrom = "=0.1.4"
+rustls = { version = ">=0.23.19", optional = true, default-features = false, features = ["ring", "logging", "std", "tls12"] }
```
epage, line 61: Is there a reason the default version requirement isn't being used?  rustls could release a `0.24.0` and builds of this version of ureq would break

epage: (ditto for `serde_json`)

algesten: I'll fix it in a patch. Thanks.

PR #907, opened 2024-12-04
Do not use multi-version deps

algesten:


algesten:
Relates to https://github.com/algesten/ureq/pull/905#discussion_r1868218499

PR #910, opened 2024-12-18
Bumps rust-url to v2.5.4

ubiratansoares:
Hi, @algesten !

Raising this PR because I got a couple of Dependabot Security alerts on [idna](https://crates.io/crates/idna), which `ureq` depends on transitively via [rust-url](https://github.com/servo/rust-url).

Some reference links

- https://github.com/advisories/GHSA-h97m-ww89-6jmq
- https://rustsec.org/advisories/RUSTSEC-2024-0421.html

Since scanners will flag this dependency on projects using `ureq`, I'm following the what the advisories recommend and bumping `rust-url` to 2.5.4.

Apologies for raising the PR directly without an issue, hope that works for you 🙏

Thanks again for `ureq`, looking forward to v3.0.0 🚀

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ socks = { version = "0.3.4", optional = true }
 # cookie_store uses Url, while http-crate has its own Uri.
 # Keep url crate in lockstep with cookie_store.
 cookie_store = { version = "0.21.1", optional = true, default-features = false, features = ["preserve_order"] }
-url = { version = "2.3.1", optional = true, default-features = false }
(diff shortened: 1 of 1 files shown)


algesten:
Hi @ubiratansoares, welcome to ureq!

I think your PR needs to read two lines above your change and at least address that comment.

https://github.com/algesten/ureq/blob/main/Cargo.toml#L65

I'm not saying it's a blocker since I believe Cargo will allow the version to be bumped and used for cookie_store too. But it's not in lock-step anymore.

PR #919, opened 2025-01-03
Avoid boxing in Connector chains

### LINKED ISSUE #890: Rethink the Transport boxing
Quoting @sosthene-nitrokey in #870:

> # There would be benefits to not boxing the connectors
> 
> In the connector traits, the transports are boxed, preventing chained connectors to make use of specificities of the previous transport. Ideally for example, it would be possible to use the specific type of the previous transport. For example a TCP keepalive connector could just wrap or come after a TcpTransport, get its TCP stream, and enable Keepalive on it.
> 
> The way I would do it would use associated types for the transports, and make `chained` and not chained connectors part of the trait.
> 
> ```rust
> pub trait Connector: Debug + Send + Sync + 'static {
>     type Transport: Transport;
> 
>     fn boxed(self) -> AnyConnector
>     where
>         Self: Sized,
>     {
>         Box::new(self)
>     }
>     fn connect(&self, details: &ConnectionDetails) -> Result<Self::Transport, Error>;
> }
> 
> pub trait AnyConnectorTrait: Debug + Send + Sync + 'static {
>     fn connect(&self, details: &ConnectionDetails) -> Result<Box<dyn Transport>, Error>;
> }
> 
> impl<C: Connector> AnyConnectorTrait for C {
>     fn connect(&self, details: &ConnectionDetails) -> Result<Box<dyn Transport>, Error> {
>         Ok(Box::new(self.connect(details)?))
>     }
> }
> 
> pub type AnyConnector = Box<dyn AnyConnectorTrait>;
> 
> impl Connector for Box<dyn AnyConnectorTrait> {
>     type Transport = Box<dyn Transport>;
>     fn connect(&self, details: &ConnectionDetails) -> Result<Self::Transport, Error> {
>         (**self).connect(details)
>     }
> }
> 
> pub type AnyTransport = Box<dyn Transport>;
> 
> pub trait MiddleConnector<T: Transport>: Debug + Send + Sync + 'static {
>     type Transport: Transport;
> 
>     fn connect(&self, details: &ConnectionDetails, chained: T) -> Result<Self::Transport, Error>;
> }
> ```
> 
> With an implementation on tuples of many sizes through a macro:
> 
> ```rust
> impl<C, M1, M2> Connector for (C, M1, M2)
> where
>     C: Connector,
>     M1: MiddleConnector<C::Transport>,
>     M2: MiddleConnector<M1::Transport>,
> {
>     type Transport = M2::Transport;
> 
>     fn connect(&self, details: &ConnectionDetails) -> Result<Self::Transport, Error> {
>         self.2
>             .connect(details, self.1.connect(details, self.0.connect(details)?)?)
>     }
> }
> ```
> 
> You could still keep the `ChainedConnector` for one `Connector` + a chain of `MiddleConnector<AnyTransport>` that could also be boxed.
> 
> For connectors that may or may not "chain" like the SocksConnector followed by the TcpConnector, I would instead implement the `SocksConnector` as a:
> 
> ```rust
> struct SocksConnector<C: Connector<Transport = TcpStream>> {
>     inner: C,
> }
> 
> impl<C: Connector<Transport = TcpStream>> Connector for SocksConnector<C> { 
>     type Transport = TcpStream;
>     fn connect(&self, details: &ConnectionDetails) -> Result<Self::Transport, Error> {
>         if !details.use_socks {
>             return self.inner.connect(details);
>         }
>         // Current socks connector logic
>     }
> }
> ```
> 
> This type safety would allow for example a keepalive connector, which is much simpler that re-implementing all the functionality of the TcpConnector just to add that. It makes connectors much more composable.
> 
> ```rust
> struct KeepaliveConnector;
> 
> impl MiddleConnector<TcpTransport> for KeepaliveConnector {
>     type Transport = TcpTransport;
>     fn connect(
>         &self,
>         details: &ConnectionDetails,
>         chained: TcpTransport,
>     ) -> Result<Self::Transport, Error> {
>         let stream: &std::net::TcpStream = chained.tcp_stream();
>         // Configure Keepalive on the TcpStream;
>         Ok(chained)
>     }
> }
> ```

algesten:
Close #890

### DIFF
--- a/src/unversioned/transport/chain.rs
+++ b/src/unversioned/transport/chain.rs
@@ -1,50 +1,133 @@
-use crate::Error;
+use std::fmt;
+use std::marker::PhantomData;
 
-use super::{ConnectionDetails, Connector, Transport};
+use super::{Connector, Transport};
 
-/// Helper for a chain of connectors.
+/// Two chained connectors called one after another.
 ///
-/// Each step of the chain, can decide whether to:
-///
-/// * _Keep_ previous [`Transport`]
-/// * _Wrap_ previous [`Transport`]
-/// * _Ignore_ previous [`Transport`] in favor of some other connection.
-///
-/// For each new connection, the chain will be called one by one and the previously chained
-/// transport will be provided to the next as an argument in [`Connector::connect()`].
-///
-/// The chain is always looped fully. There is no early return.
-#[derive(Debug)]
-pub struct ChainedConnector {
-    chain: Vec<Box<dyn Connector>>,
+/// Created by calling [`Connector::chain`] on the first connector.
+pub struct ChainedConnector<In, First, Second>(First, Second, PhantomData<In>);
+
+impl<In, First, Second> Connector<In> for ChainedConnector<In, First, Second>
+where
+    In: Transport,
+    First: Connector<In>,
+    Second: Connector<First::Out>,
+{
+    type Out = Second::Out;
+
+    fn connect(
+        &self,
+        details: &super::ConnectionDetails,
+        chained: Option<In>,
+    ) -> Result<Option<Self::Out>, crate::Error> {
+        let f_out = self.0.connect(details, chained)?;
+        self.1.connect(details, f_out)
+    }
 }
 
-impl ChainedConnector {
-    /// Creates a new chain of connectors.
-    ///
-    /// For each connection, the chain will be called one by one and the previously chained
-    /// transport will be provided to the next as an argument in [`Connector::connect()`].
-    ///
-    /// The chain is always looped fully. There is no early return.
-    pub fn new(chain: impl IntoIterator<Item = Box<dyn Connector>>) -> Self {
-        Self {
-            chain: chain.into_iter().collect(),
+impl<In, First, Second> ChainedConnector<In, First, Second> {
+    pub(crate) fn new(first: First, second: Second) -> Self {
+        ChainedConnector(first, second, PhantomData)
+    }
+}
+
+impl<In, First, Second> fmt::Debug for ChainedConnector<In, First, Second>
+where
+    In: Transport,
+    First: Connector<In>,
+    Second: Connector<First::Out>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("ChainedConnector")
+            .field(&self.0)
+            .field(&self.1)
+            .finish()
+    }
+}
+
+/// A selection between two transports.
+#[derive(Debug)]
+pub enum Either<A, B> {
+    /// The first transport.
+    A(A),
+    /// The second transport.
+    B(B),
+}
+
+impl<A: Transport, B: Transport> Transport for Either<A, B> {
+    fn buffers(&mut self) -> &mut dyn super::Buffers {
+        match self {
+            Either::A(a) => a.buffers(),
+            Either::B(b) => b.buffers(),
+        }
+    }
+
+    fn transmit_output(
+        &mut self,
+        amount: usize,
+        timeout: super::NextTimeout,
+    ) -> Result<(), crate::Error> {
+        match self {
+            Either::A(a) => a.transmit_output(amount, timeout),
+            Either::B(b) => b.transmit_output(amount, timeout),
+        }
+    }
+
+    fn await_input(&mut self, timeout: super::NextTimeout) -> Result<bool, crate::Error> {
+        match self {
+            Either::A(a) => a.await_input(timeout),
+            Either::B(b) => b.await_input(timeout),
+        }
+    }
+
+    fn is_open(&mut self) -> bool {
+        match self {
+            Either::A(a) => a.is_open(),
+            Either::B(b) => b.is_open(),
+        }
+    }
+
+    fn is_tls(&self) -> bool {
+        match self {
+            Either::A(a) => a.is_tls(),
+            Either::B(b) => b.is_tls(),
         }
     }
 }
 
-impl Connector for ChainedConnector {
+// Connector is implemented for () to start a chain of connectors.
+//
+// The `Out` transport is supposedly `()`, but this is never instantiated.
+impl Connector<()> for () {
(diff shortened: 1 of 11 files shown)


src/unversioned/transport/io.rs
```diff
@@ -8,16 +8,16 @@ use super::{NextTimeout, Transport};
 /// Helper to turn a [`Transport`] into a std::io [`Read`](io::Read) and [`Write`](io::Write).
 ///
 /// This is useful when integrating with components that expect a regular `Read`/`Write`. In
-/// ureq this is used both for the [`RustlsConnector`](crate::unversioned::transport::RustlsConnector) and the
-/// [`NativeTlsConnector`](crate::unversioned::transport::NativeTlsConnector).
-pub struct TransportAdapter {
+/// ureq this is used both for the [`RustlsConnector`](crate::transport::RustlsConnector) and the
+/// [`NativeTlsConnector`](crate::transport::NativeTlsConnector).
+pub struct TransportAdapter<T: Transport = Box<dyn Transport>> {
```
algesten, line 13: This is an incorrect change.

src/unversioned/transport/mod.rs
```diff
@@ -103,13 +121,53 @@ pub trait Connector: Debug + Send + Sync + 'static {
...
+    /// This connector will be called first, and the output goes into the next connector.
+    fn chain<Next: Connector<Self::Out>>(self, next: Next) -> ChainedConnector<In, Self, Next>
+    where
+        Self: Sized,
+    {
+        ChainedConnector::new(self, next)
+    }
+}
+
+/// Box the transport to erase the types.
```
algesten, line 143: Box the _connector_

src/unversioned/transport/chain.rs
```diff
@@ -1,50 +1,133 @@
...
+    First: Connector<In>,
+    Second: Connector<First::Out>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("ChainedConnector")
+            .field(&self.0)
+            .field(&self.1)
+            .finish()
+    }
+}
```
sosthene-nitrokey, line 47: IV depends on the Secure Element and does not run without it being enabled

sosthene-nitrokey, line 47: One way would be to implement this for tuples of sizes 0 to 8 through a macro.

algesten: You mean as an alternative syntax to .chain().chain().chain()?

sosthene-nitrokey: Yes

algesten: Sure. I don't know how much extra it aids clarity though?

sosthene-nitrokey:
Thank you! 

I'll try find the time to see if this fits our use case this month.

src/unversioned/transport/chain.rs
sosthene-nitrokey: This avoids ugly `ChainedConnector<First, ChainedConnnector<Second, ChainedConnector<...,...>>>` types and would instead allow:

`ChainedConnector<(First, Second, Third, ...)>`

algesten: The intention is to never need to interact with `ChainedConnector` directly.  Only via the `Connector.chain()` call. We could `#doc(hidden)` it maybe? 

The ideal would be if `.chain()` could return `-> impl Connector`, but that's not allowed in traits.

sosthene-nitrokey: It's not about the docs as much as the error messages.

> The ideal would be if .chain() could return -> impl Connector, but that's not allowed in traits.

It is since Rust 1.75: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html, though not recommended in this case since it has  the issue of not being able to name the type and will not  solve the error messages.

algesten: Yeah, I want to keep MSRV on 1.71 for now though.

What is the error message you mean?

sosthene-nitrokey: image

Changing a bit the `DefaultConnector` code to create a type error leads to this. The full type is `ChainedConnector<(), ChainedConnector<(), ChainedConnector<(), ChainedConnector<(), ChainedConnector<(), ChainedConnector<(), (), WarnOnNoSocksConnector>, TcpConnector>, RustlsConnector>, WarnOnMissingTlsProvider>, WarnOnMissingTlsProvider>, ConnectProxyConnector>
`

It also means that `DefaultConnector` boxes the `inner` instead of having a more explicit type (and self-documenting)

```rust
#[derive(Debug)]
pub struct DefaultConnector {
    inner: ChainedConnector<(), (
        #[cfg(feature = "_test")]
        test::TestConnector,
        SocksConnector,
        WarnOnNoSocksConnector,
        TcpConnector,
        #[cfg(feature = "rustls")]
        RustlsConnector,
        #[cfg(feature = "native-tls")]
        NativeTlsConnector,
        #[cfg(feature = "_tls")]
        WarnOnMissingTlsProvider,
    )>,
}
```

algesten: Isn't that just the nature of combinators in Rust? Like every single thing on an `Iterator` produces some intermediary type.

sosthene-nitrokey: Iterators are very rarely stored and named. And given the combinators used with iterator using closures, you often wouldn't be able to name them anyways. Here the we combine named types, and there are only two types of combinators: `Chaining` and `Either`. I'm not suggesting doing this for `Either` because it's much more annoying to generate many enum types. `Chaining` it pretty well suited to being made easy to name.

algesten: Just using Iterator as example. Another is something like warp https://crates.io/crates/warp

sosthene-nitrokey: More recently, xilem, axum and leptos do what I'm thinking about, implementing traits on many tuple length through a macro.

Issue #920, opened 2025-01-03
Question : ETA on version 3


rob-p:
Hi,

  Thanks for the awesome crate! I’m planning to transition from minreq to urea in a project we have in my lab, and I noticed that you are on the release candidates for version 3. Do you have any idea what the ETA might be for version 3.0? Also, what is the intended support plan going forward? Will 3.x replace 2.x entirely, or will 2.x continue to be maintained and developed for sometime? Ideally, I’d like to adopt 3.0 rather than 2.0 and then transitioning, but that decision kind of depends on the intended target dates for the new version. Thanks!

algesten:
@rob-p the intention is to release 3.0 "any day now"

We're aiming for a "time of quiet" in bug reports after I publish a Release Candidate. RC3 was published November 27. RC4, yesterday (2nd January).

RC3 had one [really embarrassing bug](https://github.com/algesten/ureq/issues/909). However there was only one, the rest of the points in the [change log](https://github.com/algesten/ureq/blob/main/CHANGELOG.md#300-rc4) are simple improvements. If RC4 has no bugs reported over a couple of weeks, we will release.

As for the future of 2.x. There will be no further development, but we will, upon request, do maintenance patch releases, _within reason_. There's an increasing complexity in upholding the MSRV promises for the 2.11.x (MSRV 1.67) and 2.12.x (MSRV 1.71) lines. At some point the time investment in 2.x will overshadow the returns.

My advice is that any new usage should be on [3.0.0-RC4](https://crates.io/crates/ureq/3.0.0-rc4) version.

rob-p:
Hi @algesten,

  Thanks for the quick answer.  This is really encouraging!  I'll do the minreq <-> ureq replacement with `3.0.0-RC4` for now, and plan to develop using that version going forward.  Thanks for the great crate and good luck with finalizing the 3.0 release.

rob-p:
Hi @algesten, do you think it's safe at this point to rely on 3.0.0-rc5 in an application crate? We are only using basic get functionality (with redirects).

Thanks,
Rob

algesten:
@rob-p I think so, yes.

PR #924, opened 2025-01-04
add `rustls-aws-lc-rs` feature for alternative to `ring`

mrchantey:
I hope you dont mind me opening a pr without prior discussion, feel free to close for whatever reason.

We're working on adding ureq as a bevy dependency, its lightweight nature is perfect for a game engine, but we're unable to use ring due to its license requiring us to make acknowledgements in all advertising material.

This pr adds two features, each of which enables the `rustls` feature:
- `rustls-ring` (default)
- `rustls-aws-lc-rs`

More info for context:

- [bevy ureq pr](https://github.com/bevyengine/bevy/pull/16366)
- [ring licencing issues](https://github.com/briansmith/ring/issues/1827)
- [ureq aws-lc-rs discussion 1](https://github.com/algesten/ureq/issues/751)
- [ureq aws-lc-rs discussion 2](https://github.com/algesten/ureq/issues/765)
  - > You can certainly still use aws_lc_rs with ureq by taking your own rustls dep, activating that feature, and using AgentBuilder::tls_config to provide your own config.
    
    I'm not sure this solution would be accepted in bevy, both as a licensing grey area and the dependency bloat from having two crypto providers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,6 +114,31 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f409eb70b561706bf8abba8ca9c112729c481595893fd06a2dd9af8ed8441148"
+dependencies = [
+ "aws-lc-sys",
+ "paste",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.24.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "923ded50f602b3007e5e63e3f094c479d9c8a9b42d7f4034e4afe456aa48bfd2"
+dependencies = [
+ "bindgen",
+ "cc",
+ "cmake",
+ "dunce",
+ "fs_extra",
+ "paste",
+]
+
+[[package]]
 name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -124,6 +149,29 @@ name = "base64ct"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
+
+[[package]]
+name = "bindgen"
+version = "0.69.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
+dependencies = [
+ "bitflags",
+ "cexpr",
+ "clang-sys",
+ "itertools",
+ "lazy_static",
+ "lazycell",
+ "log",
+ "prettyplease",
+ "proc-macro2 1.0.92",
+ "quote 1.0.38",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn 2.0.94",
+ "which",
+]
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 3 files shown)


algesten:
@mrchantey Hi there! Thanks for considering ureq!

The discussions you linked concerns ureq 2.x, but the changes you provide are against the main branch, which is ureq-3.x. So the first question in my mind is: which version are you looking at? For new usage, I advice using 3.0.0-RC4. I answered a question about [that yesterday](https://github.com/algesten/ureq/issues/920)

For 3.x, the default choice is still ring, but there is an escape hatch by setting the process default: https://docs.rs/ureq/3.0.0-rc4/ureq/#rustls – this could arguably be elaborated better in the docs. Is this a sufficient fix for what you're trying to do?

mrchantey:
Ok great yep happy to wait for `ureq-3.x`, this feature isn't time critical.

If i understand correctly the following configuration will result in both `ring` and `aws-lc-rs` being compiled and included in the final binary, but `ring` would be ignored?

I'm not sure if including code but *absolutely* not using it is a licencing workaround, definitely above my pay grade :sweat_smile:. I'll ask around on the bevy discord.

```toml
[dependencies]
ureq = { version = "3.0.0-rc.4", default-features = false, features = [
  "rustls",
  "gzip",
] }
# ensure the same version as ureq
rustls = { version = "0.23", default-features = false, features = [
  "aws_lc_rs",
] }
```

```rust

fn main(){

   rustls::crypto::aws_lc_rs::default_provider()
        .install_default()
        .expect("failed to install rustls https crypto provider");

   let res = ureq::get(str_path).call();
   // etc

}
```

alice-i-cecile:
I don't think we should ship the code, even unlinked. It would be really nice to have it properly feature flagged out.

algesten:
@mrchantey @alice-i-cecile thanks for the feedback!

To explain where current situation. The choice of TLS backend is often made due to compliance. Typically the story goes "you must use the platform provided TLS library so we can update it centrally" (this is the `NativeTls` config option). Furthermore, when ureq is used in libraries, there's a risk feature flags are enabled that are beyond the user's control. Example:

myapp -> lib1 -> ureq [feature flag A]
myapp -> lib2 -> ureq [feature flag B]

This would end up merging feature flag A+B in the build. This is why we insist the choice of TLS is _explicit config_, not automatically picked up from feature flags.

The rewrite of ureq 3.x happened exactly at the same time as `rustls` switched from `ring` to `aws-lc-rs`. At the time, the chances of compiling correctly, was higher with `ring`. ureq's first and foremost imperative is to "work out of the box", hence defaulting to `ring`. I don't know if the situation has changed in the 6 months since then.

You present a new concern: avoid `ring` from a compliance perspective. I'm scratching my head a bit on how to approach it. Some thought in random order.

* How common is this concern? This is the first time I encountered a situation where `NativeTls` option is not a good route.
* We want to avoid maintaining multiple TLS backends inside ureq 3.x (this went wrong in 2.x)
* ureq 3.x has a new "escape hatch" where `Transport` is possible to change. It's possible to provide other TLS transports in separate crates.
* I don't want new users to ureq having to get their head around feature flags like `rustls-ring` vs `rustls-aws-lc-rs`. ureq is meant to be simple.

I'll fiddle around today and see if I can figure out a good way forward balancing all these thoughts.

Thanks for raising it!

algesten:
#931 is a potential way forward. I like it because it opens up for any `CryptoProvider`, not just `aws-lc-rs`. It does however fall outside of semver major-version guarantees. I don't see a way around that as long as `rustls` is not 1.x.

algesten:
Your specific use case would be:

```text
ureq = { version = "3", default-features = false, features = ["rustls-no-provider"] }
rustls = { version = "0.23", features = ["aws-lc-rs"] }
```

```rs
use std::sync::Arc;
use ureq::{Agent};
use ureq::tls::{TlsConfig, TlsProvider};
use rustls::crypto;
///
let crypto = Arc::new(crypto::aws_lc_rs::default_provider());
///
let agent = Agent::config_builder()
    .tls_config(
        TlsConfig::builder()
            .provider(TlsProvider::Rustls)
            // requires rustls or rustls-no-provider feature
            .unversioned_rustls_crypto_provider(crypto)
            .build()
   )
   .build()
   .new_agent();
```

mrchantey:
Awesome i think we can close this in favor of #931

PR #926, opened 2025-01-04
Consume (some) body on drop

### LINKED ISSUE #94: Attempt to return dropped Body to pool
In one of my apps, I use ureq to read/write from a key/value server. When I perform writes, the status response is sufficient to know failure or success. However, if I do not read the response body, the connection doesn't return to the pool since the data remains on the stream. This seems odd and requires in my app adding an explicit read on the response in order to allow the stream to return to the pool. I was curious why this is the case and whether or not we could just flush the buffer:

https://github.com/algesten/ureq/blob/master/src/pool.rs#L218

algesten:
Close #94

### DIFF
--- a/src/body/build.rs
+++ b/src/body/build.rs
@@ -46,6 +46,7 @@ impl BodyBuilder {
                 mime_type: None,
                 charset: None,
                 body_mode: BodyMode::NoBody,
+                consume_on_drop: None,
             },
             limit: None,
         }
--- a/src/body/mod.rs
+++ b/src/body/mod.rs
@@ -315,6 +315,22 @@ impl Body {
         let handler = self.source.into();
         BodyWithConfig::new(handler, self.info.clone())
     }
+
+    /// Discards the body up to `max` bytes.
+    ///
+    /// This is useful if you want to discard the body without reading it. The connection
+    /// will be returned to the pool (if applicable) and ureq will be able to reuse it.
+    ///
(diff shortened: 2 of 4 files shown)


algesten:
This is probably not possible to implement currently. If we introduce a `Drop` trait on `Body` or `BodyReader`, we end up in a situation that makes this code impossible:

```rs
let reader = ureq::get("http://httpbin.org/bytes/100")
    .call()?
    .body_mut()
    .with_config()
    // Reader will only read 50 bytes
    .limit(50)
    .reader();
```

The error is

```
error[E0716]: temporary value dropped while borrowed
  --> src/body/mod.rs:415:14
   |
6  |   let reader = ureq::get("https://httpbin.org/get")
   |  ______________^
7  | |     .call()?
   | |____________^ creates a temporary value which is freed while still in use
...
12 |       .reader();
   |                - temporary value is freed at the end of this statement
13 |   Ok::<_, ureq::Error>(())
14 |   } _doctest_main_src_body_mod_rs_412_0().unwrap() }
   |   - borrow might be used here, when `reader` is dropped and runs the `Drop` code for type `BodyReader`
```

I believe that introducing a `Drop` trait anywhere in the body-handling code, will trigger this error. The above code is important for ergonomics (using `body_mut().xxx` as opposed to intermediate `let`). Maybe this can be solved in future Rust.

PR #927, opened 2025-01-04
Body::discard and content_length

algesten:
- **Replace some header names with constants**
- **Body::content_length() for known sizes**
- **Body::discard() to discard and re-pool connection**


algesten:
This salvaged what was usable from #926

Issue #731, opened 2024-02-16
Make redirect history available


RossLote:
I've been getting some strange redirect behaviour and I'd like a way to find out whet is happening. For the URL I'm fetching I've been getting a `TooManyRedirects, message: Some("reached max redirects (5)")` error.

I fetched the same URL using the python requests library and I only got 1 redirect so I think there may be a problem. I'm actually in charge of the entire stack and can assure you that there is only 1 redirect in the entire request cycle.

I've managed to get around it by setting the redirect limit to 6 but because I have no visibility of the history I can't find out what is going on.

Is there a reason the history is private?

RossLote:
Did anyone see this?

algesten:
I don't really have an answer to why the history is private more than the smaller the API the less headache to maintain ureq.

algesten:
I welcome help on this for ureq 3.x.

* A new config option that is false by default.
* Collect each redirect location into a Vec
* Make Vec available through `ResponseExt` trait.

thkoch2001:
I'm moving from reqwest to ureq to get rid of async.
Reqwest provides me with an url() method for the final url after redirects:
https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.url

I'd need just the final url, but the full history is of course also a solution.

algesten:
> I'd need just the final url, but the full history is of course also a solution.

This is actually already implemented in the latest RC. https://docs.rs/ureq/3.0.0-rc3/ureq/trait.ResponseExt.html#tymethod.get_uri

algesten:
Closed by #939

PR #939, opened 2025-01-16
Redirect history

harmless-tech:
#731
Adds the option `.save_redirect_history(bool)` to the config.
When true the function `get_redirect_history()` on the `ResponseExt` trait returns a `Some(&[Uri])` which will contain the request, any redirects, and the final uris.
When false `get_redirect_history()` returns `None`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -145,6 +145,7 @@ pub struct Config {
     max_redirects: u32,
     max_redirects_will_error: bool,
     redirect_auth_headers: RedirectAuthHeaders,
+    save_redirect_history: bool,
     user_agent: AutoHeaderValue,
     accept: AutoHeaderValue,
     accept_encoding: AutoHeaderValue,
@@ -275,6 +276,17 @@ impl Config {
     /// Defaults to `None`.
     pub fn redirect_auth_headers(&self) -> RedirectAuthHeaders {
         self.redirect_auth_headers
+    }
+
+    /// If we should record a history of every redirect location,
+    /// including the request and final locations.
+    ///
+    /// Comes at the cost of allocating/retaining the Uri for
+    /// every redirect loop.
+    ///
+    /// Defaults to false
+    pub fn save_redirect_history(&self) -> bool {
+        self.save_redirect_history
     }
 
     /// Value to use for the `User-Agent` header.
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -145,6 +145,7 @@ pub struct Config {
     max_redirects: u32,
     max_redirects_will_error: bool,
     redirect_auth_headers: RedirectAuthHeaders,
+    redirect_history: bool,
```
algesten, line 148: Let's call this `save_redirect_history`, if we follow the logic of "verb + noun" it sounds a bit strange to "redirect history" (redirect it where?).

src/config.rs
```diff
@@ -482,6 +491,15 @@ impl<Scope: private::ConfigScope> ConfigBuilder<Scope> {
         self
     }
 
+    /// If we should record a history of every redirect location,
+    /// including the request and final locations.
+    ///
+    /// Defaults to false
+    pub fn redirect_history(mut self, v: bool) -> Self {
```
algesten, line 498: Explain that enabling this comes at the cost of allocating/retaining the `Uri` for every redirect hop.

src/response.rs
```diff
@@ -6,12 +6,20 @@ use crate::http;
...
 ///
 /// Allows the user to access the `Uri` in http::Response
 pub trait ResponseExt {
     /// The Uri we ended up at. This can differ from the request uri when we have followed redirects.
     fn get_uri(&self) -> &Uri;
+
+    /// The full history of uris, including the request and final uri.
+    ///
+    /// Returns None when `redirect_history` is false.
+    fn get_redirect_history(&self) -> Option<&Vec<Uri>>;
```
algesten, line 22: Let's change this to `-> Option<&[Uri]>`. The user can't really do anything with the fact that it is a Vec, so it's better for us to consider that an internal detail. In the future we might consider using SmallVec/TinyVec and store it on the stack instead.

What do you think about always returning `-> &[Uri]` and always return empty if `save_redirect_history` is off? It would be easier for the user to not have to unwrap the optional. Possibly confusing that it's always empty if you failed to read the docs.

algesten (changes requested):
Hi @harmless-tech, and welcome to ureq!

Thanks for looking at this. Overall looks great. Some small comments about naming etc. One question about signature.

src/response.rs
harmless-tech: > What do you think about always returning -> &[Uri] and always return empty if save_redirect_history is off? It would be easier for the user to not have to unwrap the optional. Possibly confusing that it's always empty if you failed to read the docs.

I did the Option since I figured there could be confusion with `get_redirect_history` returning an empty.
 Vec/Slice. Also, since `extensions().get` returns an Option it is simple to just map it to `Some(Slice)` when `save_redirect_history` is true, or return the `None`.

algesten: Alright. Let's leave it like that.

algesten (approved):

algesten:
Thank you very much for taking this on!

PR #950, opened 2025-01-27
Fix excessive stack sizes

### LINKED ISSUE #949: `Response<Body>` takes up 10920 bytes of memory
Is it reasonable that `Response<Body>` has a size of 10920?

I looked into it and most of it comes from two optional `Flow`s in `BodyHandler`, which take up each 5184.
Further down the reason is `AmendedRequest` , which is mostly made up of 4616 bytes of headers (`ArrayVec<(HeaderName, HeaderValue), MAX_EXTRA_HEADERS>`, `MAX_EXTRA_HEADERS` is 64).

I can see that removing the need for another indirection has it's performance benefits, but the header names and values are `Bytes` structs (see below), which are just fancy (32 byte wide!) pointers anyways.

Are the any benchmarks done or do you need help setting some up?

My intuition tells me, that allocating 11kB on the stack (what usually happens when getting a response) does more harm, than the benefit of reducing 1 indirection, if at all (both `Flow`s in `BodyHandler` are optional).

```rust
// from the bytes crate
pub struct Bytes {
    ptr: *const u8,
    len: usize,
    // inlined "trait object"
    data: AtomicPtr<()>,
    vtable: &'static Vtable,
}
```

algesten:
Close #949

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+  * Fix excessive stack sizes (#950)
+  * Do not enable **json** by default (breaking, but it was a mistake) (#948)
+
 # 3.0.0
 
   * Replace RequestBuilder Deref with explicit wrappers (#944)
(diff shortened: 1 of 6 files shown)


algesten:
Waiting to release new ureq-proto

PR #974, opened 2025-02-01
Make Duration::not_zero public.

sosthene-nitrokey:
This is useful for downstream wanting to re-implement `TcpTransport` This struct is only ever exposed through the `unversionned` module so its fine to expose this method and remove it in the future


src/timings.rs
```diff
@@ -180,7 +180,10 @@ pub struct NextTimeout {
 }
 
 impl NextTimeout {
-    pub(crate) fn not_zero(&self) -> Option<Duration> {
+    /// Returns the duration of the timeout if the timeout must happen, but avoid null timeouts
```
algesten, line 183: It's not exactly `null` timeout, right? It's 0. Since we have an `Optional` in context, it maybe is a bit misleading?

algesten (changes requested):

src/timings.rs
sosthene-nitrokey: Changed it to "but avoid instant timeouts"

algesten: Sounds good

PR #978, opened 2025-02-02
Turn macros into regular code

algesten:
Macros are not cached between compilations. Rewrite all macros to regular code.


algesten:
Relates to #976

PR #981, opened 2025-02-03
Add support for tcp keepalive through socket2

### LINKED ISSUE #547: Interest in a PR that adds TCP keep-alive?

sosthene-nitrokey:
Not fully implemented, probably should be using a custom structure for `TcpKeepalive` to avoid exposing `socket2` publicly.

In that case, should we have the same feature flag as the `socket2` library? Or should we allow each setting and just warn when one is not available for the given platform ?

Fixes https://github.com/algesten/ureq/issues/547

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1241,6 +1241,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 
 [[package]]
+name = "socket2"
+version = "0.5.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c970269d99b64e60ec3bd6ad27270092a5394c4e309314b18ae3fe575695fbe8"
+dependencies = [
+ "libc",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "socks"
 version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


algesten:
Are there other things we might want to tweak using `socket2`? The only thing I'm hesitating about is the name of the `keepalive` since we might want one more generic `tcp-extra` or something.

If this lands in std::net eventually, we can make the feature flags a no-op, but it'd be nice to not have many many flags in that case.

JPDye:
I think `tcp-extra` or `socket-opts` is a better feature flag name as I can see support for setting `TCP_NODELAY` being useful, especially since `ureq` seems to write requests in a way that is known to [occassionally cause issues](https://gocardless.com/blog/in-search-of-performance-how-we-shaved-200ms-off-every-post-request/) when Nagle's algorithm is on the client and delayed acknowledgements are used on the server (writing header and body in two separate calls).

EDIT:
I see `TCP_NODELAY` is already possible to set. Similar logic applies for wanting to set `TCP_QUICKACK` and others.

Issue #989, opened 2025-02-08
No checking if proxy env setting can be converted to URL


jaytaph:
At https://github.com/algesten/ureq/blob/main/src/proxy.rs#L87 ureq tries to convert proxy env settings (like HTTPS_PROXY) to a URL:

```
 let mut uri = proxy.parse::<Uri>().unwrap();
```

This assumes that is always returns an `Ok()` result, but it will fail when either the string is empty or has an invalid format.

```
let s = "";
let uri = s.parse::<Uri>().unwrap();  // fails

let s = "invalid:.////..///";
let uri = s.parse::<Uri>().unwrap(); // fails too
```

algesten:
Fixed by #990

PR #990, opened 2025-02-08
Making sure we error correctly when an invalid uri is passed

jaytaph:
Fixes issue #989 

When the URI is either empty or invalid, the unwrap() will fail so we will catch this and return a correct error result.


algesten (approved):
You're right. This is a glaring error.

Thanks!

algesten:
Just a quick `cargo fmt` to solve that.

PR #996, opened 2025-02-11
Handle request level TlsConfig

### LINKED ISSUE #995: `agent.configure_request().tls_config()` is ineffective if this is not the first request of the agent
If I have an `agent` that have already completed one HTTPS request using normal set of root certificates, and I try something like
```rust
let request = agent.configure_request(request)
    .tls_config(
        ureq::tls::TlsConfig::builder()
            .root_certs(my_specific_root_cert)
            .build() )
    .build();
let response = agent.run(request)?;
```
`my_specific_root_cert` is completely ignored.

I suspect that this is because `rustls::ClientConfig` struct is being cached:
https://github.com/algesten/ureq/blob/ac8065cab33e713319423784e02998404ccf8708/src/tls/rustls.rs#L57

I suppose that a proper solution to this would be somehow detecting a change in the TLS configuration and creating a per-request `rustls::ClientConfig` instead of reusing the cached value.

Alternatively, the documentation of `Agent::configure_request` should mention that changing `tls_config` on the resulting `ConfigBuilder` is unsupported.

algesten:
Close #995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Stop passing internal state in Config (#996)
   * Support request level TlsConfig (#996)
 
 # 3.0.5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Support request level TlsConfig (#996)
+
 # 3.0.5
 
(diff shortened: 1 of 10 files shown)


algesten:
Good point!

I think I can move the flag to be beside the config.

Thanks!

PR #1008, opened 2025-02-19
Verify Transport is TLS

### LINKED ISSUE #1006: ConnectionDetails::needs_tls is a bug hazard
Hi.  For context, I'm currently reviewing [an MR](https://gitlab.torproject.org/tpo/core/arti/-/merge_requests/2724) which adds an `arti_ureq` library to `arti.git`, allowing use of `ureq` over Tor.

I think the `ConnectionDetails::needs_tls` method is an accident waiting to happen.  It would be very easy for someone to `impl Connector` and read only the information they need out of `ConnectionDetails`.  Such a naive implementation would fail to do TLS, and just speak plaintext HTTP.

I think there should be an engineered safeguard that would ensure that this mistake would be discovered.  Ideally, an API safeguard that would detect it at compile time.

(I don't think this is a very serious issue worthy of a RUSTSEC, because in practice failing to do TLS when expected would probably cause the system to fail to work entirely, in normal use, so such a buggy program wouldn't remain deployed for long if at all.)

### Design sketch for remediation

```
-    pub addrs: ResolvedSocketAddrs,
+    pub target: ResolvedConnectionTarget,
...
+enum ResolvedConnectionTarget {
+    Plaintext { addrs: ResolvedSocketAddrs },
+    // connect to `addrs`, perform TLS, and expect a certificate for `tls_accept_hostname`
+    Tls { addrs: ResolvedSocketAddrs, tls_accept_hostname: &'str },
+}

algesten:
Close #1006

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -142,6 +142,12 @@ pub enum Error {
     /// Attempt to connect to a CONNECT proxy failed.
     ConnectProxyFailed(String),
 
+    /// The protocol requires TLS (https), but the connector did not
+    /// create a TLS secured transport.
+    ///
+    /// This typically indicates a fault in bespoke `Connector` chains.
+    TlsRequired,
+
     /// hoot made no progress and there is no more input to read.
     ///
     /// We should never see this value.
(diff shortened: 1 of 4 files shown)


ijackson:
LGTM, thanks!

PR #1011, opened 2025-02-21
Create RequestExtension trait that allows for attaching agents directly to `http::Request`.

### LINKED ISSUE #998: Missing a request builder that accepts request method as argument in ureq 3
In ureq 2, you could define the method by passing it to the `request` function:
```
let resp: ureq::Response = ureq::request("GET", "http://example.com/").call()?;
```
In ureq 3, this `request` function has been removed so there isn't a way to a dynamically define the request method. 

I attempted a workaround by using `ureq::run()` and building a `http::Request` on its own, however that meant that I can no longer access `ureq` methods to attach to the request itself before running it. Any chance that something will be added in version 3 that accepts the request method as an argument to the request builder?

GeeWee:
Creates a RequestExtension trait that will allow attaching ureq-specific methods to an `http::Request`.

Currently the trait method allows attaching default or specified agents, and configuring and running the request inline.

Fixes #998

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -525,6 +525,7 @@ use http::{Request, Response, Uri};
 pub use proxy::Proxy;
 pub use request::RequestBuilder;
 use request::{WithBody, WithoutBody};
+pub use request_ext::RequestExt;
 pub use response::ResponseExt;
 pub use send_body::AsSendBody;
 
@@ -553,6 +554,8 @@ pub mod tls;
 
 #[cfg(feature = "cookies")]
 mod cookies;
+mod request_ext;
+
 #[cfg(feature = "cookies")]
 pub use cookies::{Cookie, CookieJar};
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1144,7 +1144,7 @@ pub(crate) mod test {
             .build()
             .call()
             .unwrap_err();
-        assert!(matches!(err, Error::LargeResponseHeader(65, 5)));
+        assert!(matches!(err, Error::LargeResponseHeader(516, 5)));
     }
 
     // This doesn't need to run, just compile.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1144,7 +1144,7 @@ pub(crate) mod test {
             .build()
             .call()
             .unwrap_err();
-        assert!(matches!(err, Error::LargeResponseHeader(516, 5)));
+        assert!(matches!(err, Error::LargeResponseHeader(65, 5)));
     }
 
     // This doesn't need to run, just compile.
--- a/src/request_ext.rs
+++ b/src/request_ext.rs
@@ -214,7 +214,7 @@ mod tests {
             .unwrap()
             .with_default_agent()
             .configure()
-            .https_only(true)
+            .https_only(false)
             .build();
 
         // Assert that the request-level configuration has been set
@@ -225,7 +225,7 @@ mod tests {
             .cloned()
             .unwrap();
 
-        assert_eq!(request_config.0.https_only(), true);
+        assert_eq!(request_config.0.https_only(), false);
     }
 
     #[test]
--- a/src/request_ext.rs
+++ b/src/request_ext.rs
@@ -1,6 +1,9 @@
-use crate::config::ConfigBuilder;
-use crate::typestate::HttpCrateScope;
-use crate::{http, Agent, AsSendBody};
+use std::ops::{Deref, DerefMut};
+use ureq_proto::http::{Request, Response};
+use crate::config::{Config, ConfigBuilder, RequestLevelConfig};
+use crate::typestate::{HttpCrateScope, RequestScope};
+use crate::{http, Agent, AsSendBody, Body, Error};
+use crate::config::typestate::RequestExtScope;
 
 /// Extension trait for [`http::Request<impl AsSendBody>`].
 ///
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -553,6 +554,8 @@ pub mod tls;
 
 #[cfg(feature = "cookies")]
 mod cookies;
+mod request_ext;
```
GeeWee, line 557: I wasn't sure whether to create a new file or not - I figured the request file maybe was long enough already? Would be happy to move it into that file if you want.

src/request_ext.rs
```diff
@@ -0,0 +1,136 @@
+use crate::config::ConfigBuilder;
+use crate::typestate::HttpCrateScope;
+use crate::{http, Agent, AsSendBody};
+
+/// Extension trait for [`http::Request<impl AsSendBody>`].
```
GeeWee, line 5: I've tried to match the style of documentation and examples to the existing styles, but lmk if you want any changes made.

src/request_ext.rs
```diff
@@ -0,0 +1,136 @@
...
+/// in the plain http API.
+pub trait RequestExt<S>
+where
+    S: AsSendBody,
+{
+    /// Allows ureq-specific configuration for this request.
+    ///
+    /// This function will return a `ConfigBuilder` that can be used to set ureq-specific options.
+    /// The resulting method can be used with `ureq::run` or with an existing agent.
+    /// Configuration values set on the request takes precedence over configuration values set on the agent.
```
GeeWee, line 17: I think this is correct based on a glance of the source code? I wasn't quite sure, because `ureq` somewhere also describes it might be an error to run something created with one agent on another agent, and I'm not quite sure how this plays into this - as technically it's not created with any agent.

src/request_ext.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::RequestLevelConfig;
+
+    #[test]
+    fn set_https_only_to_true_on_get_request() {
```
GeeWee, line 54: I've just tested that the request.extensions() get set properly. To be fair, seeing as this mostly calls into existing ureq-machinery, I consider these tests more "does the API work" tests, than actual test of the functionality :)

GeeWee (commented):
I've tried to do my best to make this easy to merge in and match the existing style. I'd be happy to make changes if needed :)

src/request_ext.rs
lolgesten: Yeah, I was just going to say. This is where it breaks down. When you do `configure_request()`, we clone the agent level config to make the request level config. And this config is further manipulated and later used in `run.rs`.

With this patch, this sequence of events would fail:

1. Configure a bespoke agent with a global timeout of 10 seconds.
2. Use RequestExt::configure() to set something else, like `https_only(true)`.
3. Run the configured Request on agent made in 1.

This would result in a request that does not have the 10 second timeout because we are below using `Agent::new_with_defaults()` as the starting point for the request level config.

I'm toying with the idea whether we can make an API that borrows the agent until the request is executed.

```rs
fn configure<'a>(self, agent: &'a Agent) -> ConfigBuilder<HttpCrateScope<'a, S>> {
...
}
```

and combine that with that instead of `build()` followed by `Agent::run()`, we expose `run()` straight on the builder. The call would be:

```rs
 let request: http::Request<()> = http::Request::builder()
                .method(http::Method::GET)
                .uri("http://foo.bar")
                .body(())
                .unwrap();
  request
        .configure()
        .https_only(false)
        .run()  // <---- run here
```

```

File: src/request_ext.rs (hunk 4 above)
Comment (lolgesten) [reply src/request_ext.rs:54]: Sounds good!

**Review by lolgesten — CHANGES_REQUESTED** [review]
Thanks for looking at this!

Some thoughts.

**Comment by lolgesten** [comment]
Sorry I used my work github user. 🙈

File: src/request_ext.rs (hunk 3 above)
Comment (GeeWee) [reply src/request_ext.rs:17]: Yes I agree with the issue!

I've seen your other suggestion, but I think this is probably the crux of the issue, so will respond here first - I think it makes sense to finish this discussion before any of the other stuff becomes relevant.

Does the same issue not also happen if you use `Agent::configure_request` and then use `ureq::run` on the request afterwards?

I'm wondering if the more intuitive way to do this isn't to have each field on the request-based config be an `Option` and then default to the agent-based configuration if it's not set on the request level?

File: src/request_ext.rs [hunk 5]
```diff
@@ -0,0 +1,254 @@
...
+}
+
+pub struct WithAgent<'a, S: AsSendBody> {
+    pub(crate) agent: AgentRef<'a>,
+    pub(crate) request: Request<S>,
+}
+
+impl<'a, S: AsSendBody> WithAgent<'a, S> {
+    /// TODO
+    pub fn configure(mut self) -> ConfigBuilder<RequestExtScope<'a, S>> {
```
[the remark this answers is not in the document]
algesten: Yes. That is the right way.

src/request_ext.rs
```diff
@@ -0,0 +1,295 @@
...
+        let req_level: &mut RequestLevelConfig = self.request.extensions_mut().get_mut::<RequestLevelConfig>().unwrap();
+
+        &mut req_level.0
+    }
+}
+
+/// Reference type to hold an owned or borrowed [`Agent`].
+pub enum AgentRef<'a> {
+    Owned(Agent),
+    Borrowed(&'a Agent),
```
GeeWee, line 132: This borrowed agent was an `&'a mut Agent` in your PR. I don't think there's any situation where we actually mutate the agent, so I figured it was easier to just hold an agent, as it'll also allow the agent to be used across multiple requests in parallel.

src/request_ext.rs
```diff
@@ -0,0 +1,295 @@
...
+            .extensions()
+            .get::<RequestLevelConfig>()
+            .cloned()
+            .unwrap();
+
+        assert_eq!(request_config.0.http_status_as_error(), true);
+    }
+
+    #[test]
+    fn set_http_status_as_error_to_false_on_post_request_with_default_agent() {
```
GeeWee, line 246: These tests all test the `build` behaviour. Optimally I'd also like to test the `with_agent().configure()...run()` behaviour, but I am unsure how.

src/request_ext.rs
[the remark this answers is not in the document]
GeeWee: Awesome!

src/request_ext.rs
algesten: Sounds good! The pool is behind an `Arc<Mutex<>>`, so it makes sense.

algesten (changes requested):
Thanks @GeeWee. It's looking good and we're very nearly there.

I just got some small notes.

1. Run cargo fmt.
2. Run cargo test (you got two tests failing, one regular, one doc).
3. Shorten the names of the test cases.

GeeWee:
Ah, those are some embarrassing oversights..!
- I've ran `cargo fmt`
- I've fixed the test and the doctest
- I've shortened the names of the test cases, although many of them test the same behaviour, so I haven't really been able to distinguish them meaningfully based on the names. Feel free to suggest concrete names if you have them :)

I'm getting a few errors in tests locally in stuff like `redirect_max_without_error` and `limit_max_response_header_size`, but I think that might be just `httpbin` being a bit wonky.

algesten:
> I'm getting a few errors in tests locally in stuff like `redirect_max_without_error` and `limit_max_response_header_size`, but I think that might be just `httpbin` being a bit wonky.

If you run the tests with `cargo test --features _test` you use a connector that makes no real requests to httpbin. This also happens with `cargo test --all-features`.

algesten (changes requested):
A couple of tiny things left:

1. Run this: `RUSTDOCFLAGS=-Dwarnings cargo doc --all-features --no-deps  --document-private-items`
2. Run this: `cargo clippy`

GeeWee:
Both done! I wasn't quite sure how to link to the `RequestExt` trait in the RustDocs, but it probably doesn't matter if there's a nicer way to do it, as it's internal-only docs :)

GeeWee:
Thanks for your patience!

PR #1014, opened 2025-02-22
RequestExt take 2

algesten:
Alternative idea to #1011 

The idea is to introduce a single `RequestExt::with_agent()` which adds an agent to an http crate `Request`.

That in turn "unlocks" `configure()` and `run()` to configure or run on that agent.


algesten:
Close in favor of #1011

PR #1021, opened 2025-02-25
update MBEDTLS connector for ureq2

mcr:
This updates the ureq-2 example connector for MbedTLS to 0.13,0 so that it works. Now working on ureq-3 connector (provider?) and also psa-crypto connector.

### DIFF
--- a/examples/mbedtls/Cargo.toml
+++ b/examples/mbedtls/Cargo.toml
@@ -5,5 +5,5 @@ edition = "2021"
 
 
 [dependencies]
-mbedtls = { version = "0.11.0" }
+mbedtls = { version = "0.13.0", features=["ssl","x509"] }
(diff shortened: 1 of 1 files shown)


mcr:
I hope this is the right ureq2 branch to send patches against.
I want to be on ureq3, but first, I have to start from working code :-)

algesten:
> I hope this is the right ureq2 branch to send patches against.

It's good. I got two branches for ureq2 depending on MSRV. I will try cherry-pick over. Though I'm not actively updating 2.x unless it's required due to breaking changes.

> I want to be on ureq3, but first, I have to start from working code :-)

Sounds good! My hope is that the new Transport structures will make it possible to maintain a separate ureq-mbedtls crate.

PR #1025, opened 2025-02-27
middleware/doc: use Relaxed atomics for single-variable counter

birkenfeld:
Since there is only one atomic variable in play here, it is not useful to use anything other than Relaxed.


algesten (approved):
Ok! Thanks!

PR #1032, opened 2025-03-08
Doc and Error::Other for bespoke connector chains

### LINKED ISSUE #1029: Document which Error a Transport is supposed to use
We're writing a [binding to try to wire ureq and Arti (Tor Client) together](https://gitlab.torproject.org/tpo/core/arti/-/merge_requests/2724).

Right now, if the connection across the Tor network fails, the ureq API seems to wants us to throw `ureq::Error::ConnectionFailed`.  The result will be that the ultimate user sees almost no information about the error.

It's probably too late to improve `ConnectionFailed`, but maybe a different variant shoudl be provided that would allow us to transport our own error somehow (boxed, presumably).  Or should we be converting everything to `ureq::Error::Io`?  Because that contains an `io::Error`, we could make use of `std::Io::Error`'s escape hatch...

algesten:
Close #1029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Improved errors and doc for bespoke transports (#1032)
+
 # 3.0.8
 
   * Fix incorrect parsing bug "missing http version" (#1026)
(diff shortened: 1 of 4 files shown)


ijackson:
Thanks.  I looked t this and it seems to do a good job of addressing #1029.  So, thanks :-).

Issue #1012, opened 2025-02-21
Add migration guide entries for proxy-from-env and native-certs feature flags


michaeljones:
I'm migrating from 2 to 3 and wondering what to do about the fact that `proxy-from-env` and `native-certs` are no longer feature flags.

Would it be possible to add sections in the migration guide about them if specific steps need to be taken? They are trickier things to test in our application code base so I'm a bit nervous about it. I don't want to burden you though so only makes sense to do it if you think it'll serve the project, of course.

Thanks again for ureq. 

Oh, one of the things I did get caught out by in the migration is around `ureq::OrAnyStatus`. I first thought you couldn't set it on a per-request basis but only through the agent config but I have since found the config method on the `RequestBuilder`. I can try to add a section to the guide about that if that would be useful?

algesten:
Great questions!

> I'm migrating from 2 to 3 and wondering what to do about the fact that `proxy-from-env` and `native-certs` are no longer feature flags.

* `proxy-from-env` is the default now.  CONNECT-proxy needs no extra feature flag, but `socks-proxy` does.
* `native-certs` is built-in. In the `TlsConfig`, which you can set on agent or request level, you have three choices via the [`RootCerts`](https://docs.rs/ureq/3.0.6/ureq/tls/enum.RootCerts.html#variant.PlatformVerifier) enum. `Specific` when you want to set the root certs yourself, `PlatformVerifier`, which for rustls delegates to the system, and for native-tls means using the root certs native-tls is picking up (this is what you want), and finally `WebPki`, which uses the root certs bundled with ureq.


> Would it be possible to add sections in the migration guide about them if specific steps need to be taken? 

Sure, let's document it!

> Oh, one of the things I did get caught out by in the migration is around `ureq::OrAnyStatus`. I first thought you couldn't set it on a per-request basis but only through the agent config but I have since found the config method on the `RequestBuilder`. I can try to add a section to the guide about that if that would be useful?

Yes please!

A PR documenting all of this would be most welcome!

PR #1041, opened 2025-03-25
Update to migrate doc

devmatteini:
Hello!

Update to migrate doc based on questions from #1012

I think it would also be useful to add a link to the migrate doc in the README.md since many people will search for it before upgrading to 3.x.


algesten:
Thanks!

PR #1050, opened 2025-04-10
implement facet-json - very naive implementation

asibahi:
Hello.

`fasterthanlime` created a new rust reflection library: [`facet`](https://www.reddit.com/r/rust/comments/1jvh6vo/facet_rust_reflection_serialization/).

I do not expect this to be merged as-is, or for now. `facet-json` is still extremely barebones. Simply a stub to expand things later on, I suppose. Also it requires a higher Rust version than `ureq` (1.86 vs 1.71)

I'd appreciate a review for how to iterate, improve on this, or if it is interesting at all.


algesten:
Hi @asibahi!

Looks fun! Facet looks very cool, and I hope it takes off. Overall my thoughts are:

* A conservative MSRV is very important to this project.
* Will facet take off and become as ubiquitous as serde?
* If it does become popular, people will probably want one or the other, meaning we should have separate feature flags for it.

So for now, my overall feedback is that we need to wait and see where facet is going. It could be that by the time facet has gained enough traction, the MSRV issue is less of a problem.

PR #1068, opened 2025-05-26
Fix cookie-value compliance check

### LINKED ISSUE #1067: cookies compliance checking does not follow RFC
The cookie value checking is done here:
https://github.com/algesten/ureq/blob/8508e5ce73339051fe76ad117e1441aa5b6eb871/src/cookies.rs#L274

The checking ignores one case in the RFC:
```
 cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
```
where the value could be wrapped in a pair of double quotes.

Proposed change:
```rust
 let valid_value = value.strip_prefix(br#"""#).and_then(|value| value.strip_suffix(br#"""#)).unwrap_or(value).iter().all(is_valid_value); 
```

johnmave126:
Fix #1067

Check the whole cookie value only if it is not enclosed by a pair of double-quotes, otherwise strip exactly one pair of double-quotes and check the rest of the cookie value.

### DIFF
--- a/src/cookies.rs
+++ b/src/cookies.rs
@@ -271,7 +271,12 @@ fn is_cookie_rfc_compliant(cookie: &cookie_store::Cookie) -> bool {
 
     let value = cookie.value().as_bytes();
 
-    let valid_value = value.iter().all(is_valid_value);
+    let valid_value = value
+        .strip_prefix(br#"""#)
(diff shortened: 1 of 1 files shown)


algesten:
Thanks!

PR #1071, opened 2025-05-27
Fix form urlencoding using `form_urlencoded` crate from servo

### LINKED ISSUE #1070: send_form() encoding does not follow spec
Per https://url.spec.whatwg.org/#application/x-www-form-urlencoded (emphasis mine)
> Let value be the result of running [percent-encode after encoding](https://url.spec.whatwg.org/#string-percent-encode-after-encoding) with encoding, tuple’s value, the [application/x-www-form-urlencoded percent-encode set](https://url.spec.whatwg.org/#application-x-www-form-urlencoded-percent-encode-set), and **true**.

Per https://url.spec.whatwg.org/#string-percent-encode-after-encoding (emphasis mine)
> To percent-encode after encoding, given an [encoding](https://encoding.spec.whatwg.org/#encoding) encoding, [scalar value string](https://infra.spec.whatwg.org/#scalar-value-string) input, a percentEncodeSet, and **an optional boolean spaceAsPlus (default false)**:

This means when performing `x-www-form-urlencoded` encoding, the space should be encoded as `+` instead of `%20`. 

However, in the current form
https://github.com/algesten/ureq/blob/9cd5cea5e7663a8b7555bc48e445b4ea5394da01/src/request.rs#L584-L586
https://github.com/algesten/ureq/blob/9cd5cea5e7663a8b7555bc48e445b4ea5394da01/src/query.rs#L50-L53

The space is encoded as `%20` since https://docs.rs/percent-encoding/latest/percent_encoding/fn.utf8_percent_encode.html does not provide the `spaceAsPlus` flag.

It might be worth it to simply replace the form encoding part by https://crates.io/crates/form_urlencoded, also from the servo team.

johnmave126:
Should fix #1070 under my limited testing.
It is debatable whether this is the best way since it introduces a new dependency. It is possible to hack it in `query.rs` and introduces a `form_url_enc()` to handle this. If you want to minimize dependencies and elect for that instead, let me know and I can get that done relatively quickly as well.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -58,10 +58,14 @@ pub fn form_url_enc(i: &str) -> Cow<str> {
         "%20" => "+",
         _ => part,
     });
+
+    // We try to avoid allocating if we can (returning a Cow).
     match iter.next() {
         None => "".into(),
         Some(first) => match iter.next() {
+            // Case avoids allocation
             None => first.into(),
+            // Following allocates
             Some(second) => {
                 let mut string = first.to_owned();
                 string.push_str(second);
--- a/src/query.rs
+++ b/src/query.rs
@@ -52,6 +52,26 @@ pub fn url_enc(i: &str) -> Cow<str> {
     utf8_percent_encode(i, ENCODED_IN_QUERY).into()
 }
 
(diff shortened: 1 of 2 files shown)


algesten:
Thanks!

PR #1074, opened 2025-05-29
Chunked transfer handle abrupt abort after 0\r\n

schneems:
## Problem

I was under the mistaken impression that the reader returned by the `into_reader` didn't perform checks for Content-Length. It does, but even after consulting the documentation in several places, I moved forward with adding my own check in https://github.com/heroku/libcnb.rs/pull/947. A co-worker corrected my implementation, which led me to discover it wasn't needed. Neither of us came to the conclusion that we were already protected via documentation alone.

One of the doc locations that I checked was `into_reader`.

## Fix

I added two error cases to the docs where the reader can (and should) make errors: when a connection is closed before Content-Length bytes are received ~and if an ungraceful close of the TCP connection occurs~.

At the same time, the subject of the bullet point on the reader not being "limited" wasn't clear to me. To help, I added some more details.

I understand the documents weren't written for me personally, and we cannot add every possible caveat to every possible source of documentation. Feel free to take this feedback and run with it if you like. Thanks for this library!

----

Addition. ureq can now handled a server sending `0\r\n` to end a transfer and then abruptly aborting the connection.

### DIFF
--- a/src/body/mod.rs
+++ b/src/body/mod.rs
@@ -212,8 +212,6 @@ impl Body {
     ///   To set a limit use [`Body::into_with_config()`].
     /// * Reader will error if `Content-Length` is set, but the connection is closed
     ///   before all bytes are received.
-    /// * Reader will error if `Transfer-Encoding: chunked` is set, but the connection
-    ///   is closed without a FIN message (graceful connection shutdown) from the server.
     ///
     /// # Example
     ///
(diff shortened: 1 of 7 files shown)


src/body/mod.rs
```diff
@@ -207,7 +207,13 @@ impl Body {
...
     /// get a non-shared, owned reader via [`Body::into_reader()`].
     ///
-    /// * Reader is not limited. To set a limit use [`Body::with_config()`].
+    /// * Reader is not limited by default. That means a malicious server could
+    ///   exhaust all avaliable memory on your client machine.
+    ///   To set a limit use [`Body::into_with_config()`].
+    /// * Reader will error if `Content-Length` is set, but the connection is closed
+    ///   before all bytes are received.
+    /// * Reader will error if `Transfer-Encoding: chunked` is set, but the connection
+    ///   is closed without a FIN message (graceful connection shutdown) from the server.
```
algesten, line 216: This last point sounds like we potentially could fix. If we have received the chunk ending, we could maybe treat further `io::Error` as unimportant (not pool the connection, but also not surface any error).

algesten (commented):
Looks good! Thanks! One question.

PR #1085, opened 2025-06-17
Example of bespoke rustls connector

algesten:


algesten:
This is not a working example. We should maybe make a completely functioning variant? 🤔

Issue #1098, opened 2025-08-05
SendBody to automatically content-length and content-type


algesten:
> Change SendBody to track the size of the body in more cases (at least supporting File and &File).
> 
> This is an interesting idea, that I'd like to discuss in a separate PR. The other day I was wondering whether `SendBody` should track both `content-length` and `content-type` where appropriate. 

 _Originally posted by @algesten in [#698](https://github.com/algesten/ureq/issues/698#issuecomment-3153894367)_

# Idea

The `SendBody` currently does not carry `content-length` or `content-type`, even in cases where those values are known. The trait `AsSendBody` is implemented for stuff like `&[u8]` and it clearly could set at least `content-length`, but maybe also `content-type: application/binary` (to be discussed).

ureq would still of course defer to whatever headers the user sets.

# Breaking?

The BIG question is whether this could be a breaking change for some users. For example: the default behavior when not setting a `content-length` is to use `transfer-encoding: chunked`. Are there ureq users that _must_ use chunked transfer and are currently relying on the default behavior?

algesten:
@gmacon let's discuss this part here.

gmacon:
The `SendBody` does currently track content length for `&str`, `String`, `&[u8]`, `Vec<u8>`, and `impl Serialize` (and you could argue for the degenerate case of `()`). I was proposing adding it for `File` by capturing the length (if it's available) before type-erasing it down to `dyn Read`.

I think it's less clear that `SendBody` should track the content type; you could use `text/plain` for `&str` and `String`, `application/json` for `impl Serialize`, and `application/octet-stream` for everything else, but that's starting to feel more opinionated that I really want an HTTP client to be. The existing multpart form code infers the content type from the file extension when adding a file from a `&Path`, but requires the user to specify the type for the most general `add_stream` API.

I don't think this would be a breaking change; I would argue that if switching from CTE:chunked to Content-Length causes a problem, it's the *server* that's broken. (The other way, i.e. switching from explicit length to CTE:chunked, I think *would* be breaking.)

algesten:
Sounds good. In summary: We will track content-length, but not content-type.

PR #1099, opened 2025-08-07
Track size for files

gmacon:
To allow producing Content-Length headers when using a File as the body, try to get the size of the file before type-erasing it. If the File points to something other than a regular file (e.g. a named pipe), or if we fail to get the metadata for some reason, use the current behavior of sending it with Content-Transfer-Encoding: chunked.

This is an implementation of the first half of what's proposed in #1098, but I haven't done all of the polishing yet. For instance, there are no tests or documentation updates.

### DIFF
--- a/src/send_body.rs
+++ b/src/send_body.rs
@@ -64,6 +64,7 @@ impl<'a> SendBody<'a> {
             BodyInner::Reader(v) => v.read(buf),
             BodyInner::OwnedReader(v) => v.read(buf),
             BodyInner::Body(v) => v.read(buf),
+            BodyInner::SizedReader(v, _) => v.read(buf),
         }?;
 
         if n == 0 {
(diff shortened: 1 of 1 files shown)


src/send_body.rs
```diff
@@ -64,6 +64,7 @@ impl<'a> SendBody<'a> {
             BodyInner::Reader(v) => v.read(buf),
             BodyInner::OwnedReader(v) => v.read(buf),
             BodyInner::Body(v) => v.read(buf),
+            BodyInner::SizedReader(v, _) => v.read(buf),
```
algesten, line 67: I think we should make both `Reader` and `OwnedReader` have an extra `Option<ContentLength>` instead of introducing a new enum type here.

algesten:
I've implemented this ins #1100

Issue #668, opened 2023-10-04
Respect `NO_PROXY` environment variable.


zicklag:
With the `try_proxy_from_env(true)` option, `ureq` correctly reads the `HTTP[S]_PROXY` environment variables, but it doesn't honor the `NO_PROXY` environment variables. Also, it's not possible to efficiently implement from outside of `ureq`, because you would have to create a new `Agent` for each request to a different domain, instead of being able to share the same agent state among all domains.

mcr:
Zicklag ***@***.***> wrote:
    > `NO_PROXY` environment variables. Also, it's not possible to
    > efficiently implement from outside of `ureq`, because you would have to
    > create a new `Agent` for each request to a different domain, instead of
    > being able to share the same agent state among all domains.

Also redirects that had different proxy needs wouldn't work.

algesten:
@zicklag 

> but it doesn't honor the NO_PROXY environment variables.

I know very little about proxies. PR welcome.

> Also, it's not possible to efficiently implement from outside of `ureq`, because you would have to create a new `Agent` for each request to a different domain, instead of being able to share the same agent state among all domains.

Agent is reusable for multiple requests.

@mcr 

> Also redirects that had different proxy needs wouldn't work.

Not sure I follow. Do you mean there should be different proxy settings per request host?

mcr:
Martin Algesten ***@***.***> wrote:
    >> Also redirects that had different proxy needs wouldn't work.

    > Not sure I follow. Do you mean there should be different proxy settings per request host?

The point of NO_PROXY is that it does not use the proxy for certain hosts.
So, in effect, yes, there are different proxy settings (namely: none) for
some hosts.

algesten:
ureq 3.x needs support for this. We should do all the options curl does:

https://github.com/curl/curl/blob/master/docs/libcurl/opts/CURLOPT_NOPROXY.md

swsnr:
In general, proxy lookup should use the interface `lookup(url: Url) -> Vec<Url>` provided by the user of the API.  ureq would call this function for every request, pass the target, and get a list of proxy URLs in return, in order of priority.  

[...]

PR #1111, opened 2025-09-04
Implement support for NO_PROXY

arielf212:
This should close (the very long standing) issue #668.

I am willing to do anything needed to get this merged, as this will allow me to close typst/typst#6366.
I know that there are currently no tests, but I want to see that I got at least the idea correctly.

### DIFF
--- a/src/hostname_matcher.rs
+++ b/src/hostname_matcher.rs
@@ -5,6 +5,8 @@ use ureq_proto::http::Uri;
 pub enum HostnameMatcher {
     /// Matches the pattern literally - by string equality
     Literal(String),
+    /// Subdomain match - equivalent to checking if ther pattern is the suffix to the string
+    Suffix(String),
     /// Matches any string
     MatchAll,
 }
@@ -14,7 +16,13 @@ impl HostnameMatcher {
         if pattern == "*" {
             return Self::MatchAll;
         }
-        Self::Literal(pattern.to_owned())
+
+        let pattern = pattern.to_owned();
+
+        if pattern.starts_with('.') {
+            return Self::Suffix(pattern);
+        }
+        Self::Literal(pattern)
     }
 
     pub fn matches(&self, uri: &Uri) -> bool {
@@ -24,6 +32,7 @@ impl HostnameMatcher {
 
         match self {
             Self::Literal(lit) => lit == hostname,
+            Self::Suffix(suffix) => hostname.ends_with(suffix),
             Self::MatchAll => true,
         }
     }


algesten:
Hi! Thank you for looking at this!

I was starting to give PR feedback, when I realized I don't really know what values can go into NO_PROXY. Asking Kagi AI, I got:

```text
Valid values for the NO_PROXY environment variable can include:

    * Hostnames: Such as google.com or localhost.
    * Fully qualified domain names (FQDNs): For example, internal.example.com.
    * IP Addresses: Like 192.168.1.1.
    * IP address ranges or subnets: Using CIDR notation, such as 10.96.0.0/12 or 192.168.1.0/24.
    * Domains with wildcards: To include all subdomains, you can use patterns like *.example.com.
    * Ports: You can specify a port along with a hostname or IP address, e.g., google.com:443.
```

Being AI, I'd take those bullets with a grain of salt.

If we look at reqwest though, we see that we need to support a bit more than just host names: https://docs.rs/reqwest/latest/reqwest/struct.NoProxy.html#method.from_string

It seems we need at least 4 classes of matching here:

* Host/sub-domain (`google.com` meaning both the direct host as well as `some.foo.google.com`)
* Wildcard (`*.google.com`, which I guess matches `foo.google.com`, but not `some.foo.google.come`, but we need to verify this!)
* IP address (`192.168.0.1`, both ipv4 and ipv6)
* CIDR (`192.168.0.1/24`, both ipv4 and ipv6)

(not sure about port, i.e. `google.com:1234`, is that required?)

I don't see any of this matching being particularly hard to plug in to ureq, but there are some semantics to iron out. Like, if we set an IP or CIDR, does that mean we're matching against the original URL, like `https://192.168.0.42` or do we match against the DNS resolved address? Or both?

To proceed I think we need to dig into the exact expected semantics here. It could be enough to read the source code for reqwest (and/or curl). This probably means we end up with some enum in this impl like `enum NoProxy` that has a variant for each of the cases.

This is in turn would affect the public API in terms of what `Config` exposes.

arielf212:
Currently as it stands, there is no standard which defines what NO_PROXY means.
Every program and library has their own implementation of it which mostly-kinda-coincide with one another.

IMO its kinda of your library your rules, so I will support whatever combination you'll request as none of them are particularly hard to implement.

I'd advise against trying to match ips after DNS resolving as that can open up a giant can of worms. :)

algesten:
> I'd advise against trying to match ips after DNS resolving as that can open up a giant can of worms. :)

Yeah. This seems to be the way. Only match the CIDR against potential IP in URLs.

arielf212:
This is a great article by Gitlab on the topic: https://about.gitlab.com/blog/we-need-to-talk-no-proxy/.
Anyways, I upstreamed a basic design for the matcher. I think I'll add the rest of the matcher types tommorow. :)

arielf212:
We can go in one of three ways, it seems:
1. We are a minimal library: only implement literal match, maybe subnet matching and also support hardcoded `*` as a "disable proxy at all" variable.
2. Majority is correct - implement and respect whatever the majority of the libraries implement.
3. Curl is correct - Curl is obviously the reference implementation for http clients for most people, might as well align ourselves to them.

What are your thoughts on this?

arielf212:
I am honestly just kind of scared that we will introduce a lot of parsing complexity for a relatively niche feature if we'll try to support the whole spec.

algesten:
Closing in favor of #1118 

On balance, I think we can keep all NO_PROXY business inside `proxy.rs` and when I was looking more closely, I don't think there needs to be additional API `Config`. #1118 also needs a bit of work, but it's closer to what I think we should land.

PR #1118, opened 2025-09-18
feat: respect `NO_PROXY`

sirimhrzn:
A bit of backstory, we ran into a issue in production where [nushell](https://github.com/nushell/nushell) was ignoring the `NO_PROXY` environment variable. Then after diving a bit deep it turns out that it depends on `ureq` which seems to not parse the the environment value.

This PR implements minimal pattern matching for `NO_PROXY` environment variable values to ensure proper proxy bypass behavior.

## Supported Patterns

| Pattern       | example.com | api.example.com | sub.domain.example.com |
|---------------|-------------|-----------------|------------------------|
| *.example.com | ❌ No match | ✅ Match        | ✅ Match               |
| .example.com  | ❌ No match | ✅ Match        | ✅ Match               |
| example.com   | ✅ Match    | ❌ No match     | ❌ No match            |
| *             | ✅ Match    | ✅ Match        | ✅ Match               |

> [!NOTE]
> IP address & CIDR (e.g., `192.168.1.1`, `192.168.1.1/24`) matching is not supported.

Multiple hosts can be specified in the `NO_PROXY` variable separated by commas.

## Related Issues
- [http get not using $env.no_proxy](https://github.com/nushell/nushell/issues/15244)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Support for env NO_PROXY and proxy config #1118
+
 # 3.1.2
 
   * Fix bug when query is after host "example.com?query" #1115
--- a/README.md
+++ b/README.md
@@ -143,13 +143,13 @@ The default enabled features are: **rustls** and **gzip**.
 * **platform-verifier** enables verifying the server certificates using a method native to the
   platform ureq is executing on. See [rustls-platform-verifier] crate
 * **socks-proxy** enables proxy config using the `socks4://`, `socks4a://`, `socks5://`
-   and `socks://` (equal to `socks5://`) prefix
+  and `socks://` (equal to `socks5://`) prefix
 * **cookies** enables cookies
 * **gzip** enables requests of gzip-compressed responses and decompresses them
 * **brotli** enables requests brotli-compressed responses and decompresses them
 * **charset** enables interpreting the charset part of the Content-Type header
-   (e.g.  `Content-Type: text/plain; charset=iso-8859-1`). Without this, the
-   library defaults to Rust's built in `utf-8`
+  (e.g.  `Content-Type: text/plain; charset=iso-8859-1`). Without this, the
+  library defaults to Rust's built in `utf-8`
 * **json** enables JSON sending and receiving via serde_json
 
 #### Unstable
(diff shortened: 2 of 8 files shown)


m4siri:
The current implementation has a `try_from_env()` which reads the `NO_PROXY` environment variable. In nushell, it builds the **Proxy** using the `new()` method. I think it would be nice to add something such as `respect_no_proxy()` to read from env or a `with_no_proxy(list: &[&str])`.
https://github.com/nushell/nushell/issues/15244#issuecomment-3310421213

algesten:
Hey @sirimhrzn! Welcome to ureq!

I got a lot on lately, and hope to get to this PR this weekend. Turns out we have two PRs for NO_PROXY. This one and #1111 – would you mind making a quick comparison and we'll decide which PR to focus our efforts on.

m4siri:
Sure, https://github.com/algesten/ureq/pull/1111 introduces a separate module for the matcher which I do not believe is necessary. It being in the same file as proxy should be enough.
https://github.com/algesten/ureq/blob/fb80defbc8c487a046b30523b2605f4f1a1b9120/src/config.rs#L10
It also carries the list for no_proxy within the **Config** struct. My implementation provides a method on **Config** which verifies if the host shall be proxied or not. It is a wrapper around the actual implementation that lives in the **ProxyInner** struct. I'm not sure if https://github.com/algesten/ureq/pull/1111 has been tested properly as it has not checked the *no proxy list* on `connect(5)`. I had to check there as well for mine to work.
```diff
diff --git a/src/run.rs b/src/run.rs
index 508ebc88..244975b7 100644
--- a/src/run.rs
+++ b/src/run.rs
@@ -356,8 +356,11 @@ fn connect(
     // For most proxy configs, the proxy itself should resolve the host name we are connecting to.
     // However for SOCKS4, we must do it and pass the resolved IP to the proxy.
     let is_proxy_local_resolve = config.proxy().map(|p| p.resolve_target()).unwrap_or(false);
+    // This determines if we should proxy the request.
+    // This respects the environment variable `NO_PROXY`.
+    let should_proxy = config.should_proxy(uri);
 
-    let addrs = if !is_proxy || is_proxy_local_resolve {
+    let addrs = if !should_proxy || !is_proxy || is_proxy_local_resolve {
         agent
             .resolver
             .resolve(uri, config, timings.next_timeout(Timeout::Resolve))?
```

Issue #1120, opened 2025-10-04
rustls-pemfile is archived


skull-squadron:
Found this while running `cargo unmaintained`: https://github.com/rustls/pemfile/issues/61

algesten:
@skull-squadron thanks!

This happened quite recently and as of [rustls 0.23.32](https://crates.io/crates/rustls/0.23.32/dependencies) we're still pinned on rustls-pki-types ^1, meaning there isn't a way to move on to 2.2.0.

We can only monitor the situation for dependencies to update first.

cpu:
> we're still pinned on rustls-pki-types ^1, meaning there isn't a way to move on to 2.2.0.

I think it's workable to drop `rustls-pemfile` with the current `rustls-pki-types` dep. PTAL: https://github.com/algesten/ureq/pull/1122

algesten:
@cpu The comment in in `rustls-pemfile`, what does it mean?

https://github.com/rustls/pemfile/issues/61#issue-3334846983

> The main function of this crate has been incorporated into [rustls-pki-types](https://crates.io/crates/rustls-pki-types). 2.2.0 maintains the existing public API for this crate, on top of this new implementation.

Why is mentioning 2.2.0 if the function was present already in 1.9.0? If it was in 1.9.0, but "2.2.0 maintains the existing public API", does that mean there's another public API change coming (affecting ureq's `Error::Pem`)?

cpu:
> Why is mentioning 2.2.0 if the function was present already in 1.9.0? 

2.2.0 is a reference to the final version of **`rustls-pemfile`**. The 2.2.0 release replaces the guts of that crate with calls to the `rustls-pki-types` PEM parsing that has been present since 1.9.0 of that crate. Your current version spec for `rustls-pemfile` is 2.1.0 and so I think `cargo update` would take 2.2.0 and there would be no changes required (but, it won't address the issue the OP raised about the repo being archived, and no future updates will occur).

> does that mean there's another public API change coming (affecting ureq's Error::Pem)?

No. There's no plan to change the API in `rustls-pki-types`. We're even more conservative about breaking changes there than in other crates.

Hope that helps!

algesten:
Got it! Thanks!

PR #1122, opened 2025-10-05
tls: replace rustls-pemfile usage with rustls-pki-types

cpu:
[rusts-pemfile](https://github.com/rustls/pemfile) has been archived and won't see future updates. The [rustls-pki-types](https://github.com/rustls/pki-types) crate has supported PEM parsing since [v1.9.0](https://github.com/rustls/pki-types/releases/tag/v%2F1.9.0).

This commit converts the existing parsing while maintaining semver compatibility, with the exception of the `Error::Pem()` enum variant. This variant previously exposed a `rustls-pemfile` error type, and now switches to a similar `rustls-pki-types` error type. This breaking change is **not** considered in-scope for `ureq`'s semver policy based on the pre-existing rustdoc comment:

> *Note:* The wrapped error struct is not considered part of ureq API. Breaking changes in that struct will not be reflected in ureq major versions.

Resolves https://github.com/algesten/ureq/issues/1120

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1096,15 +1096,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "rustls-pemfile"
-version = "2.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


algesten:
@cpu Excellent! Thank you for straightening this out.

PR #1124, opened 2025-10-13
error:Translate Error::Timeout into io::ErrorKind::Timedout

rawler:
This enables better categorization of errors.

In particular, TcpTransport translates `ErrorKind` to `ureq::Error::Timeout` errors on timeout conditions. These should then be re-raised as `ErrorKind::Timedout` errors behind the `io::Error` abstraction.

Omitting this error-structure forces downstream error-handlers to resort to string-matching to distinguish timeout-errors from other errors.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -185,11 +185,14 @@ impl std::error::Error for Error {}
 impl Error {
     /// Convert the error into a [`std::io::Error`].
     ///
-    /// If the error is [`Error::Io`], we unpack the error. In othe cases we make
-    /// an `std::io::ErrorKind::Other`.
+    /// If the error is [`Error::Io`], we unpack the error. If the error is
+    /// [`Error::Timeout`] we translate to `io::ErrorKind::TimedOut`.
+    /// In other cases we make it an `std::io::ErrorKind::Other`.
     pub fn into_io(self) -> io::Error {
         if let Self::Io(e) = self {
             e
+        } else if let Self::Timeout(_) = self {
+            io::Error::new(io::ErrorKind::TimedOut, self)
         } else {
             io::Error::new(io::ErrorKind::Other, self)
         }
--- a/src/unversioned/transport/io.rs
+++ b/src/unversioned/transport/io.rs
@@ -85,3 +85,45 @@ impl<T: Transport> io::Write for TransportAdapter<T> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::io::{self, Read};
+
+    use super::super::*;
+    use super::*;
+
+    #[test]
+    fn timeout_errors_translate_transparently() {
+        #[derive(Debug)]
+        struct FailingTransport(LazyBuffers);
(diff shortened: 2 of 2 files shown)


rawler:
Not sure this is necessarily the _best_ way to solve it. But in particular, I've discovered _some_ (but not all) errors disguising themselves as `Fetch error: IO(Other, Some("timeout: receive response"))` in production. In my application, these are kindof expected occasionally, and deserve slightly less aggressive logging than others.

algesten:
@rawler thanks for submitting this.

I've been thinking back/forth on whether we can land it. In some respects it can be considered a breaking change.

rawler:
Good reflection. I briefly touched on that thought too, but my thinking is that anyone relying on the old behavior would (hopefully) not be matching on `ErrorKind=Other`, but on string-matching the message, which I believe to be unchanged.

rawler:
Are we reaching any verdict? Is there some _other_ approach that would allow us to classify timeout errors (ideally without string-matching)?

algesten:
I haven't forgotten, just not made up my mind.

rawler:
I poked around at this again and (after messing with getting a `rustls` test-server to shoot at) finally realized the repro-case is smaller and more reliable than I thought. In short, to read a `ureq::Body` in a streaming fashion, one has to convert it to an `std::io::Read` using `.into_reader()`. When doing this, ureq-errors are coerced into `io::Error`s, causing timeout categorization to be lost.

So, if I haven't missed something, there's currently no way short of string matching, to read a response as a stream with timeout-handling, short of string-matching on errors.


```
    fn make_timeout_request() -> (std::sync::mpsc::Sender<()>, ureq::Body) {
        let listener = std::net::TcpListener::bind((std::net::Ipv6Addr::LOCALHOST, 0)).unwrap();
        let port = listener.local_addr().unwrap().port();
        let (sentinel, receiver) = std::sync::mpsc::channel();

        std::thread::spawn(move || {
            let (mut stream, _) = listener.accept().unwrap();

            stream
                .read(&mut [0u8; 1024])
                .expect("Failed to read client request");

            // Write some of the response, but not all of it, to trigger a timeout while client is reading the body.
            stream
                .write_all(b"HTTP/1.1 200 OK\r\nContent-Length: 1024\r\n\r\nAAAAA...")
                .expect("Write failed");

            // Do not close the connection, to trigger timeout on client side.
            // This is expected to block until the test concludes
            let _locked = receiver.recv().ok();
        });

        let request = ureq::get(&format!("http://[::1]:{port}"))
            .config()
            .timeout_recv_response(Some(Duration::from_millis(40)))
            .build();

        let response = request.call().expect("Request failed");

        (sentinel, response.into_body())
    }

    #[test]
    fn test_std_timeout_visible() {
        let (_sentinel, body) = make_timeout_request();

        let mut buf = Vec::new();
        let error = body.into_reader().read_to_end(&mut buf).unwrap_err();

        assert_eq!(error.kind(), io::ErrorKind::TimedOut);
    }

```

Issue #1013, opened 2025-02-22
Ring is not doing so well


algesten:
https://github.com/briansmith/ring/discussions/2414

Currently ring is the default backend for ureq via `rustls`. This is something we might need to revise. This is not a breaking change because we got the `TlsConfig` abstraction between us and the crypto provider.

We also clearly point out that the [default crypto provider might change](https://docs.rs/ureq/3.0.6/ureq/#rustls):

> ureq does not guarantee to default to ring indefinitely. `rustls` as a feature flag will always work, but the specific crypto backend might change in a minor version.

I'm hopeful there one day will be a TLS library that is 100% Rust, but the natural choice at this point would be to switch rustls to `aws-lc-rs`.

I'd love to hear if any users of ureq have thoughts on this.

parasyte:
Yesterday I went through the cumbersome process of learning how to replace

```rust
let agent = ureq::agent();
```

with

```rust
let agent = Agent::from(
    Agent::config_builder()
        .tls_config(
            TlsConfig::builder()
                .provider(TlsProvider::NativeTls)
                .build(),
        )
        .build(),
);
```

For all the downsides, this is preferable to `cargo deny` loudly pointing out that I'm using an unmaintained library.

----

`aws-lc-rs` might even be ok in the short term. But I am personally highly skeptical of the team developing it. They have [signaled in past conversations](https://github.com/rustls/rustls/issues/1913) a general disinterest of working with the community unless it suits their own needs.

I agree that the ideal situation would be a pure Rust TLS library with great performance, but I'm willing to sacrifice performance for the promise of "if it builds, it works." The alternative is not building (whatever the reason) and that's unacceptable.

This might be a bit on the contentious side, but `ureq` users are unlikely to have HTTPS workloads that benefit from shaving off [less than one millisecond](https://github.com/aochagavia/rustls-bench-results) [^1] from the request time. We don't need to jump through hoops to milk the CPU. Simple is more than adequate.

[^1]: Comparing `handshake-full_1.3_ECDHE_RSA_AES256-GCM_SHA384_client` (client handshakes per second) between `ring` and `aws-lc-rs`, the average _difference_ is just 21 µs per handshake. With an average total handshake time of 237 µs. Round up to 1 ms to conservatively account for cryptography during transfer. (Though it has significantly less overhead.)

cpu:
[...]

PR #1129, opened 2025-11-05
Remove default dependency to ring

maxammann:
A lot of library crates removed the explicit default dependency to "ring". This was done such that binary authors can choose which TLS implementation to choose from.

By having `_ring` in the default options, binary authors no longer have the choice to choose the TLS library based on the dependency setup.

Then it is needed to opt into a provider at runtime:

```
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.
See the documentation of the CryptoProvider type for more information.
```

If a binary author depends on ureq like this then ring will be pulled in:

https://github.com/mdrokz/auto_generate_cdp/blob/master/Cargo.toml#L20

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ default = ["rustls", "gzip"]
 
 ######## SUPPORTED FEATURES
 
-rustls = ["rustls-no-provider", "_ring"]
+rustls = ["rustls-no-provider"]
 native-tls = ["dep:native-tls", "dep:der", "_tls", "dep:webpki-root-certs"]
 platform-verifier = ["dep:rustls-platform-verifier"]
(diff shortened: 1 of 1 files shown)


algesten:
Hi @maxammann and welcome to ureq!

Having ring as a dependency out-of-the box is a [deliberate choice](https://github.com/algesten/ureq?tab=readme-ov-file#rustls). While I don't believe ureq should ship without a TLS provider with the default flags (it is supposed to "just work"), here's the place to discuss the current situation: https://github.com/algesten/ureq/issues/1013

PR #1136, opened 2025-11-14
Allow `post_file_sends_file_length` test with CRLF line endings

cptpiepmatz:
Hey 👋
I mainly develop on Windows, so my local git uses CRLF line endings by default. This causes the `cargo test` run to fail on `post_file_sends_file_length`. This PR updates the test to allow both LF and CRLF line endings.


algesten (approved):
let's do it

Issue #639, opened 2023-07-06
Support for socks5h


ghost:
It would be nice if there was support for `socks5h` proxying. The `h` in the scheme ensures that DNS resolution is done by the remote server that's being proxied into.

https://github.com/algesten/ureq/blob/main/src/proxy.rs#L98 - passing `socks5h` currently does not appear to be supported.

algesten:
Sounds good. I know very little about proxies. PR welcome!

algesten:
This is still todo for ureq 3.x . Help wanted.

algesten:
This depends on https://github.com/sfackler/rust-socks/issues/21

harmless-tech:
[sfackler/rust-socks](https://github.com/sfackler/rust-socks) does seem to support `socks5h` proxying already. If you pass a domain and port for `target_addr` for `Socks5Stream::connect` or `Socks4Stream::connect` at [src/unversioned/transport/socks.rs#L156](https://github.com/algesten/ureq/blob/main/src/unversioned/transport/socks.rs#L156) it will be passed to the proxy server to do the [DNS resolution](https://github.com/sfackler/rust-socks/blob/master/src/lib.rs#L30).

The implementation for `socks4a` is probably wrong too since we resolve the domain name on the host instead of sending it to the proxy server to do the DNS lookup.

algesten:
We could switch backing lib no problem.

Issue #1134, opened 2025-11-12
`Proxy::try_from_env` should prefer more specific environment variables


niklasmohrin:
I have a concern that came up in tealdeer here: https://github.com/tealdeer-rs/tealdeer/issues/450#issuecomment-3517767544

In `Proxy::try_from_env`, the environment variables `all_proxy`, `https_proxy`, and `http_proxy` are checked, in this order: https://github.com/algesten/ureq/blob/428bba83ae6a48be65bdf133137af60a28fbc2e2/src/proxy.rs#L211-L218

However, doesn't it make more sense to prefer the more specific `https_proxy` variable over the general `all_proxy` one? Additionally, should `https_proxy` and `http_proxy` even be used for for HTTP and HTTPS requests respectively?

For reference, the described behavior is what happens in hyper-util, and thus, reqwest: https://github.com/hyperium/hyper-util/blob/d91ea8efe6f3b09cc2fd6cc9e303566bd887a3ea/src/client/proxy/matcher.rs#L315-L319

Thanks for the library!

algesten:
> However, doesn't it make more sense to prefer the more specific `https_proxy` variable over the general `all_proxy` one? Additionally, should `https_proxy` and `http_proxy` even be used for for HTTP and HTTPS requests respectively?

Proxy environment variables doesn't have a spec, and it's a bit wild west out there.

ureq is a bit weird here because it treats all these variables as equals. And they (probably) are not. This is just a carry-over from the first implementation, and since I personally never been in a situation where I need to rely on these, I simply can't guide how it should work.

Order changes aside, the question is whether there are other oddities that should be corrected.

niklasmohrin:
Fair enough. Changing the order is a breaking change anyways, so maybe it's just something to keep in mind. I personally don't use a proxy, so I don't have much input on it, but the way hyper does it seems more right to me. From what was said in the linked issue, users that regularly use proxies also found the behavior unexpected.

From a security perspective, the use of http_proxy at all when performing https requests is a bit scary, but again, I am not sure how relevant it actually is.

@Jisu-Woniu can you provide some input?

Jisu-Woniu:
> the use of http_proxy at all when performing https requests is a bit scary

Actually, no. HTTPS is encrypted stream instead of plain text, this means the client and the server are safe to exchange data stream, and the proxy cannot get much information from the stream.

[...]

PR #1138, opened 2025-11-21
Introduce SOCKS5h

JPDye:
resolves algesten/ureq#639

Commonly, when using SOCKS proxies, the scheme "socks5h" is used to mean domain resolution should happen on the proxy server and "socks5" means resolution should happen locally. CURL is one of many tools to do things this way.

I've introduced `ProxyProtocol::SOCKS5h` which defaults to resolving on the proxy server and reverted `ProxyProtocol::SOCKS5` to resolve locally. I've added tests to ensure this is set properly when constructing SOCKS proxies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+  * Make socks5:// locally resolve before calling proxy #1138
+  * Add socks5h:// which DOESN'T locally resolve before calling proxy #1138
+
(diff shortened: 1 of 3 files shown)


algesten (approved):
Looks good!

Somewhat breaking, but I think it's okay because it brings ureq into line.

Do you have any thoughts on #1134

algesten:
Thanks!

Issue #1143, opened 2025-12-07
`Proxy::try_from_env` and `NoProxy::try_from_env` does not respect Windows system settings


mokurin000:
The system proxy settings on windows could be read at `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Internet Settings`

mokurin000:
This was completed in #1144

PR #1144, opened 2025-12-07
feat: read windows system proxy

mokurin000:
#1143 

This PR does respect Windows proxy settings, and implemented `HostPrefix` to match for:

```text
localhost;127.*;192.168.*;10.*;172.16.*;172.17.*;172.18.*;172.19.*;172.20.*;172.21.*;172.22.*;172.23.*;172.24.*;172.25.*;172.26.*;172.27.*;172.28.*;172.29.*;172.30.*;172.31.*;<local>
```

## New Features

- `NoProxy` and `Proxy` from windows proxy settings

## New dependencies

- [optional, default] `winreg`, which depends `windows-sys 0.59`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ rust-version = "1.71.1"
 features = ["rustls", "platform-verifier", "native-tls", "socks-proxy", "cookies", "gzip", "brotli", "charset", "json", "multipart", "_test", "_doc"]
 
 [features]
-default = ["rustls", "gzip", "win-system-proxy"]
+default = ["rustls", "gzip"]
 
 ######## SUPPORTED FEATURES
 


algesten:
Hi @mokurin000! Thanks for this!

It looks interesting. Out of curiosity is this how other libs do it as well? Like curl and reqwest? Like, pick up windows registry values in addition to env vars I mean.

mokurin000:
@algesten thanks for asking that, I had dive into curl and reqwest to check their implementation.

> It looks interesting. Out of curiosity is this how other libs do it as well? Like curl and reqwest? Like,

`reqwest` used hyper_util::client::proxy::matcher::[`Matcher`](https://rustdocs-v1.ferron.sh/hyper_util/client/proxy/matcher/struct.Matcher.html#method.from_system),
which also tries reading environment variables before try parsing from registry.

Meanwhile, the curl [reads specific environment variables over all_proxy](https://github.com/curl/curl/blob/a78a07d3a9dc808a51f464d0dd297939e544e2a4/lib/url.c#L2100-L2117), and does not respect windows/macos system proxy.

> pick up windows registry values in addition to env vars I mean.

I preserved the old behavior for people configuring `HTTPS_PROXY`, `HTTP_PROXY` and alike on windows for CLI tools that don't care windows system proxy settings.

BTW, many linux world CLI tools don't respect MacOS and Windows world system standard.

algesten:
@mokurin000 thanks for investigating.

If `curl` and `reqwest` don't read this registry value by default, I don't think ureq should either. The idea is to be as similar as possible to other libraries the user might encounter.

We should keep it under the feature flag you defined, but don't put it in the default flags.

mokurin000:
> @mokurin000 thanks for investigating.
> 
> If `curl` and `reqwest` don't read this registry value by default, I don't think ureq should either. The idea is to be as similar as possible to other libraries the user might encounter.

No, reqwest actually did, I misstated that, its parsing registry if not found from env.

https://docs.rs/reqwest/0.12.25/src/reqwest/async_impl/client.rs.html#413-415

> We should keep it under the feature flag you defined, but don't put it in the default flags.

Okay, however its a bloated feature, will include `windows-sys`

algesten:
Thanks!

PR #1152, opened 2026-01-01
Timeout per resolved ip for try_connect

### LINKED ISSUE #1150: Connect timeout vs multiple IP addresses
A user of an application I wrote that uses ureq seems to have a broken ipv6 setup, where the router doesn't let through ipv6. This causes ureq to timeout on connect, without falling back to ipv4.

Things we've tried that don't work:
 - basic request with `ureq::get("https://example.com")`
 - set global timeout `config.timeout_global(Some(Duration::from_secs(5))))`
 - set connect timeout `config.timeout_connect(Some(Duration::from_secs(5)))`
 - set global and connect timeout to 100s and 50s respectively

The only way to get ureq to work was disabling ipv6 altogether using `config.ip_family(IpFamily::Ipv4Only)`.

I suspect ureq doesn't fall back to ipv4 if it gets a ipv6 address for a domain and the subsequent ipv6 connect fails.

Solutions would include a [Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs) implementation, or at least a proper ipv6 connect timout -> ipv4 fallback.

algesten:
Close #1150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Timeout per resolved ip for try_connect #1152
   * Fix body header bug on redirect #1140
   * ureq-proto 0.5.3 to fix unsolicited 100-continue #1139
   * Make socks5:// locally resolve before calling proxy #1138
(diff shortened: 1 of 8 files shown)


algesten:
@laundmo it would be very interesting to know if the main branch solves your problem.

PR #1156, opened 2026-02-01
Strip Content-Encoding and Content-Length headers after decompression

pythcoiner:
When ureq transparently decompresses a gzip/brotli response, Content-Encoding and Content-Length headers are left in the response, leaving a stale Content-Length (compressed size) and a misleading Content-Encoding on what is now plaintext. This makes it impossible for callers to distinguish whether the body is still compressed, and breaks any code that trusts Content-Length for buffer sizing. 

This PR strips both headers when decompression is active (matching reqwest) and adds a regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Strip Content-Encoding/Content-Length headers after decompression #1156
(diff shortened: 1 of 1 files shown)


algesten:
Hi @pythcoiner this looks good! Should this PR move out of draft?

PR #1179, opened 2026-06-27
Serde interface for query params for consideration

DoumanAsh:
I would like to propose serde based interface for query parameters similar to that of [reqwest](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.query)

While I prefer simplicity of ureq, this piece of interface is pretty nice feature in reqwest as it allows you to have statically typed query parameters.

Of course there is problem with serde interface is that you cannot validate it fully at compile time, but I think accidental panic should be minimal impact as otherwise user is most likely to use well defined structs.

This PR introduces interface with struct based implementation for review and consideration
@algesten, if you think this complexity is ok, I will add map/sequence/tuple serializers which has more complexity due to dynamic nature of `key` in these objects to finalise PR

The entry point to this interface is hidden behind `serde-query` which requires to include serde, itoa and ryu dependencies.
All serde implementation is defined in `src/serde/query.rs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1120,6 +1120,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ryu"
+version = "1.0.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9774ba4a74de5f7b1c1451ed6cd5285a32eddb5cccb8cc655a4e50009e06477f"
+
+[[package]]
 name = "same-file"
 version = "1.0.6"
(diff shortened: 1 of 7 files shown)


algesten:
Hi @DoumanAsh 

I don't have anything against this in principle more than API bloat. However if there's enough interest, we could add it. Let's leave this PR open and I encourage people wanting this to speak up.

DoumanAsh:
The implementation is complete and mostly in line with serde_urlencoded for these who will be interested

PR #1180, opened 2026-06-28
Support testing in release mode

musicinmybrain:
Disable the default features for `assert_no_alloc` (specifically, `disable_release`). This should be safe for library users because all relevant allocator overrides are gated by `cfg(test)`.

This is helpful in Fedora downstream packaging, since we package Rust crates individually, run as many tests as we reasonably can, and do so in release mode ([for technical/practical reasons](https://codeberg.org/rust2rpm/rust2rpm/issues/230)).


algesten:
Sounds good! Thanks!

PR #1187, opened 2026-08-06
Remove redundant clone

FranciscoTGouveia:
If I am not overlooking anything, I believe this clone is redundant and can thus be safely removed.

For reference, this change was found by an optimizer developed as a research project.

CC @nunoplopes


FranciscoTGouveia:
I believe the CI failure was spurious... :/

algesten:
I'll see if cargo-deny fails in main.

algesten:
Thanks!

PR #1191, opened 2026-08-18
Buffer read_json into memory when Content-Length is below the limit

### LINKED ISSUE #1151: read_json is bottlenecked by serde_json's reader API
You get around a 5x performance improvement by reading the entire JSON document into memory and passing serde_json a slice rather than passing it a reader.

I confirmed using serde_json directly where even wrapping an in-memory buffer in a `std::io::Cursor` and using their reader API resulted in a big performance hit.

In ureq,`response.into_body().read_json()` took 13ms while `response.into_body().read_to_vec()` plus a seperate call to `serde_json::from_slice()` dropped that to 2ms for a 256KB document which I confirmed over multiple runs.

serde_json has this documented here: https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html

> Note that counter to intuition, this function is usually slower than reading a file completely into memory and then applying [from_str](https://docs.rs/serde_json/latest/serde_json/fn.from_str.html) or [from_slice](https://docs.rs/serde_json/latest/serde_json/fn.from_slice.html) on it. See [issue #160](https://github.com/serde-rs/json/issues/160).

Plus an open GitHub issue: https://github.com/serde-rs/json/issues/160

Suggest changing ureq's `read_json` function to read the body into a `Vec` first or at least document the problem in your docs similar to what serde_json has done.

ChrisJr404:
Fixes #1151.

`read_json()` currently always parses through `serde_json::from_reader`, which reads the response one byte at a time and is measurably slower than parsing an in-memory slice. Following the approach you sketched in the issue, this buffers the body into a `Vec` and parses with `from_slice` on the fast path, falling back to the streaming reader otherwise.

The fast path only triggers when the `Content-Length` is known and below `limit / 3`. That cutoff is what you suggested: it keeps the buffered bytes plus the deserialized value within the configured memory budget (so a 10MB limit stays roughly under 10MB total rather than doubling). Chunked responses, close-delimited responses, transparently-decompressed bodies, and anything at least a third of the limit keep the previous streaming behaviour unchanged. The limit itself is still enforced by `read_to_vec`, so a server that lies about `Content-Length` and sends more still errors out at the limit as before.

Both `Body::read_json` and `BodyWithConfig::read_json` get the optimization. I pulled the `Content-Length` logic out of `Body::content_length()` into a small private `ResponseInfo::content_length()` helper so both call sites share it.

## Changes
- Fast-path buffering in both `read_json` methods, gated on a known `Content-Length` below the 1/3 cutoff.
- `ResponseInfo::content_length()` helper (public `Body::content_length()` now delegates to it; behaviour unchanged).
- A `# Performance` note on `read_json` documenting when buffering happens.
- A unit test exercising both the buffered and streamed paths and asserting identical results.

## Testing
- `cargo test --no-default-features --features "_test rustls json"` — new `read_json_buffered_and_streamed` test and existing body tests pass.
- `cargo clippy --all-features` clean.
- `cargo fmt --check` clean.
- Builds cleanly with `json` disabled under `-D dead_code -D unused`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Speed up read_json for responses with a known, small body size #1191
+
 # 3.4.0
 
   * Seal RequestExt (technically breaking) #1189
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Unreleased
 
-  * Speed up read_json for responses with a known, small body size #1191
-
 # 3.4.0
 
   * Seal RequestExt (technically breaking) #1189
--- a/src/body/mod.rs
+++ b/src/body/mod.rs
@@ -31,14 +31,10 @@ const MAX_BODY_SIZE: u64 = 10 * 1024 * 1024;
 
 /// Fraction of the read limit under which `read_json()` buffers the body into memory
 /// and parses it from a slice (faster) rather than streaming it through serde_json's
-/// reader. The cutoff leaves some headroom for the deserialized value.
+/// reader. Buffering only when the Content-Length is below `limit / JSON_BUFFER_DIVISOR`
+/// keeps the buffered bytes plus the parsed value within the configured limit.
 #[cfg(feature = "json")]
 const JSON_BUFFER_DIVISOR: u64 = 3;
-
-#[cfg(feature = "json")]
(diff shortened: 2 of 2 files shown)


algesten (approved):
Looks good!

algesten (approved):

algesten:
Thanks!

PR #1193, opened 2026-08-21
Complete TLS handshakes while connecting

tamird:
A short send-request timeout can abort TLS establishment despite a longer connection timeout. The connection timeout is documented to cover TLS [1], but both rustls and native-tls defer handshaking until the first request write. The request loop has already marked Connect complete by then, so handshake I/O receives the timeout selected for SendRequest.

Connect-only configurations retain some handshake protection because the scheduler incorrectly checks completed-phase timeouts during later phases [2]. Correcting that accounting without moving the handshake would remove this accidental protection.

Complete each handshake inside its TLS connector, using the connection-phase timeout, before returning the transport. The Connect completion timestamp then follows TLS establishment, and a short send-request timeout no longer interrupts handshake I/O. The connector tests check that handshaking happens before return and that underlying timeout errors keep their original reason.

[1]: https://github.com/algesten/ureq/blob/5e803fc4ce031a0d667dc6cdb97d1413ac8cdab9/src/config.rs#L697-L711
[2]: https://github.com/algesten/ureq/blob/5e803fc4ce031a0d667dc6cdb97d1413ac8cdab9/src/timings.rs#L45-L175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Complete TLS handshake during connect so timeout_connect covers it #1193
   * Speed up read_json for responses with a known, small body size #1191
 
 # 3.4.0
--- a/src/config.rs
+++ b/src/config.rs
@@ -696,17 +696,26 @@ impl<Scope: private::ConfigScope> ConfigBuilder<Scope> {
 
     /// Max duration for establishing the connection
     ///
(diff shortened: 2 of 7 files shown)


tamird:
@algesten this and #1194 are needed to fix timeout bugs in aya (worked around in https://github.com/aya-rs/aya/pull/1689). Would be great to get your attention on these PRs.

algesten:
Looks good! Thanks!

PR #1194, opened 2026-08-21
Anchor timeout budgets to phase boundaries

tamird:
Aya's kernel-package downloads can fail while copying the response body with "timeout: receive response" [1]. timeout_recv_response is documented to exclude the body [2], but its deadline remains a candidate during body reads.

Phase timestamps record completion, yet the scheduler adds each completed phase's own timeout to that timestamp and checks the result in the next phase (650e9c2cc358). The active phase's deadline also restarts at the current time on every lookup (78555734b084), allowing repeated reads to renew the body budget.

Anchor the active phase's budget to its latest completed predecessor and compare only that deadline with the recorded global and per-call deadlines. Completed-phase limits then stop affecting later phases, while repeated lookups preserve the active deadline. Clock-based regression tests cover expiry, optional phase transitions, and global/per-call accounting across redirects.

[1]: https://github.com/aya-rs/aya/actions/runs/32503366233/job/96838036031
[2]: https://github.com/algesten/ureq/blob/5e803fc4ce031a0d667dc6cdb97d1413ac8cdab9/src/config.rs#L734-L751

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix timeout budgets restarting and applying to later phases #1194
   * Complete TLS handshake during connect so timeout_connect covers it #1193
   * Speed up read_json for responses with a known, small body size #1191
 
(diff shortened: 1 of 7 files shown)


tamird:
@algesten this is ready for review!

algesten (approved):

algesten:
Thanks!

PR #1200, opened 2026-09-13
Add optional reset for send and receive body timeouts

asv7c2:
Add optional reset behavior for SendBody and RecvBody timeouts.

By default, the existing behavior is preserved: the timeout uses the total budget for the corresponding body phase.

When enabled, the timeout is reset for each send or receive operation.


asv7c2:
@algesten Could you take a look at this PR when you have a chance? I have added support for reseting recv and send body timeouts between each operation.

algesten:
@asv7c2 what I don't like about this PR is that it alters the semantics of an already existing timeout. I took a stab in #1202 make something that maybe solves your problem in a similar way. See if you think it would work for you.

asv7c2:
@algesten 
Looks like #1202 works for me.

PR #1201, opened 2026-09-13
Respect request connection settings when pooling

algesten:
## Summary

Bypass Agent pool lookup and return when a request's connection settings are incompatible with the Agent configuration. This prevents request-level TLS identities and other connection overrides from borrowing or populating the Agent's pool, while preserving reuse for compatible requests.

Use exhaustive destructuring of Config and TlsConfig so every added field requires an explicit pooling decision. Compare TLS credentials, explicit root sets, and custom crypto providers by shared identity; cloned settings remain compatible. Proxy settings remain separated by the existing pool key, and request-only settings do not disable pooling.


aoto-tech:
@algesten san
Thank you for the quick response. I think it makes sense to cover not only TLS but all settings that affect connection establishment. The design looks very clean to me.

algesten:
Thanks! Let's release it.

algesten:
Independent adversarial review round 1 completed against the complete PR at c1ef805463d55f5ae40407afb486773d0d7a41e4. No actionable findings.

Reviewed exhaustive Config/TlsConfig classification, request configuration extraction and redirects, both pool lookup and return, built-in transport/resolver configuration usage, TLS connector caching, and recursive CONNECT proxy establishment. The conservative identity comparisons and disconnected Weak pool reference implement the intended boundary; no universal equivalence guarantee for custom connectors/providers is assumed.

Validation on this head (CARGO_TARGET_DIR=/Users/martin/dev/ureq/target):
- cargo test --lib pooling: 4 passed.
- cargo test --lib tls::test::: 4 passed (including root/provider identity tests and the rustls handshake timeout test).
- cargo test --lib --no-default-features pooling: 2 passed.
- cargo test --lib --no-default-features --features native-tls pooling: 4 passed.
- git diff --check origin/main...HEAD: passed.

No regression-test commit was needed; working tree remains clean.

algesten:
@aoto-tech It's released in 3.4.2

Thanks!

PR #1202, opened 2026-09-16
Add timeout_per_read and timeout_per_write

algesten:
Add `timeout_per_read` and `timeout_per_write` so a transfer can have short I/O limits while retaining separate total body, phase, per-call and global budgets. For example, an upload can allow ten minutes overall but only five seconds per transport write. Both settings default to `None` and work at Agent and request level, including on reused connections.

The limits cover request/response headers and bodies, TLS handshakes, and HTTP CONNECT negotiation. Timeouts report `Timeout::PerRead` or `Timeout::PerWrite`; an earlier existing budget retains its own reason. While awaiting 100-continue, only expiry of the 100-continue budget permits sending the body; other timeout errors propagate.

The public `Timeouts` struct is unchanged. The unversioned transport API's `NextTimeout` carries the directional allowances and adds `for_read`, `for_write`, and `Default` (existing struct literals need the additional fields or `..NextTimeout::default()`). Limits are passed with each operation instead of retained on pooled connections. Forwarding layers preserve the full budget; only the transport performing I/O selects the read or write limit, so TLS protocol traffic uses the correct direction.

SOCKS negotiation itself is excluded because the dependency opens and negotiates its socket internally without exposing timeout configuration. HTTP/TLS traffic through an established SOCKS tunnel is covered. DNS and TCP connection establishment retain their existing budgets. Documentation also explains buffering and partial-write limitations: these are transport allowances, not strict wall-clock deadlines around application reads/writes.

Validation:
- `cargo test --all-features -- --skip test::https_connect_proxy_to_https_target`: 160 unit tests and 85 doctests pass. The excluded test fails with macOS native-TLS `bad certificate format` on both this branch and unchanged main.
- `cargo test --no-default-features --features _test`: 124 unit tests and 75 doctests pass.
- `cargo fmt --check`, `cargo clippy --all-features -- -D warnings`, and documentation with `RUSTDOCFLAGS=-Dwarnings` pass.
- Regression coverage includes allowance renewal versus total budgets, headers/body traffic, pooled overrides and disabling limits, both TLS backends, CONNECT, zero limits, 100-continue errors, a real TCP read stall, and a layered read that emits a protocol write.

Related to #1200, with independent I/O settings instead of changing the meaning of body timeouts.


src/pool.rs
```diff
@@ -122,6 +123,7 @@ impl Connection {
     }
 
     pub fn maybe_await_input(&mut self, timeout: NextTimeout) -> Result<bool, Error> {
+        let timeout = timeout.for_read();
```
algesten: [P2] Preserve the original budget through layered transports. Capping here irreversibly replaces `after`/`reason` with the read allowance before TLS receives it. A TLS read can emit protocol writes (rustls `Stream::prepare_read` calls `complete_prior_io`, which writes pending TLS output); its TransportAdapter then applies `for_write` to an already read-capped value. With read=5s and write=7s, that write receives 5s/PerRead rather than 7s/PerWrite. The same premature capping also exists in the adapter and CONNECT response path, which can themselves wrap TLS. The smallest fix should preserve raw budgets through forwarding layers and select the direction where actual I/O occurs, rather than adding more timeout state. Regression commit 8137cc4 exercises the real Agent/Connection/TransportAdapter path with a transport that emits protocol output during a read: `cargo test --no-default-features --features _test layered_read_uses_write_allowance_for_protocol_output` fails with actual PerRead vs expected PerWrite.

src/io_timeout_tests.rs
```diff
@@ -0,0 +1,309 @@
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use crate::config::Config;
+use crate::transport::{
```
algesten: [P2] Split the multiline import to satisfy repository CI. The snowflake check in run 35155538510 fails specifically on src/io_timeout_tests.rs:6-8 (`Rust import multi-line failed with 1 violations`). Use separate single-line imports as required by that existing check. No additional regression test is useful: the existing CI job already demonstrates the failure.

src/pool.rs
algesten: Fixed in 1500b06. Removed directional capping from Connection, TransportAdapter and the CONNECT response loop. Forwarding transports now preserve the complete budget; the underlying transport selects the actual I/O direction. This avoids adding recovery state and simplifies the production change. Preserved regression 8137cc4: it failed with PerRead before the fix and now passes. Validation: 160 all-feature unit tests + 85 doctests (excluding the documented baseline macOS certificate failure), 124 no-default-feature unit tests + 75 doctests, Clippy, rustfmt and documentation checks pass.

src/io_timeout_tests.rs
algesten: Fixed in 1500b06 by splitting the import into two single-line use declarations. cargo fmt --check passes; the corrected imports follow the formatting required by the failing snowflake job.

michaeljones:
Can I check my understanding of this? I am hoping for something that helps in this situation: https://github.com/algesten/ureq/discussions/1125

But this sounds like it might be the time of the read itself and not the space between reads?

algesten:
@michaeljones Previously we could only set a timeout for the entire read of the body – which would translate into all the kernel calls for the entire body. These new timeouts are more narrowly over each kernel read call themselves (there are some buffers etc involved, so it's not super precisely the socket read, but that's the intention). The idea is that if you are making progress, albeit slowly, you want to continue. That couldn't be expressed before.

